Issue #3, opened 2019-08-09
(RPC) getConstants


jevonearth:
Add Interface and an RPC implementation for getting constants.

RPC API docs;
https://tezos.gitlab.io/tezos/api/rpc.html#get-block-id-context-constants

Example curl command;
`curl http://10.60.58.16:8732/chains/main/blocks/5/context/constants`
`curl http://10.60.58.16:8732/chains/main/blocks/head/context/constants`

- [ ] Implement interface
- [ ] Implement default rpc implementation for interface
- [ ] Implement type to represent constants 
- [ ] Add call to `examples/`

## Tentative description for API doc;

getConstants returns the economic constants for the given chain. Constants are constant in the context of a block. Constants were initialized in the genesis block and can only be changed by a Tezos protocol amendment. Constants are or can be different on mainnet, alphanet or other nets.

progital:
@carte7000 clarified:

With regards to issue #3 I think it does not involve exposing it in the tezos-ts package yet. issue #3 would be to implement the getConstants function in the @tezos-ts/rpc package
```
import RpcClient from '@tezos-ts/rpc';
const client = new RpcClient();
client.getConstants();
```

carte7000:
@jevonearth, @progital 

As per this document: https://gitlab.com/cryptiumlabs/tezos/blob/master/specs/migration_004_to_005.md

There is minor changes to this endpoint coming with the next Babylon_005

> There are two new constants at `context/constants` endpoint - `quorum_min` and `quorum_max`, both in centile of percentage (a fraction out of 10,000).

I suggest we create another issue so we handle that case for the new update that is coming

PR #19, opened 2019-08-21
#3 getConstants implementation

progital:


packages/tezos-ts-rpc/src/types.ts
```diff
@@ -16,6 +16,33 @@ export interface ContractResponse {
...
+  proof_of_work_nonce_size: number;
+  nonce_length: number;
+  max_revelations_per_block: number;
+  max_operation_data_length: number;
+  preserved_cycles: number;
+  blocks_per_cycle: number;
+  blocks_per_commitment: number;
+  blocks_per_roll_snapshot: number;
+  blocks_per_voting_period: number;
+  time_between_blocks: BigNumber[];
```
carte7000: It's good that we put the `BigNumber` type there however since Typescript do not do implicit type conversion it does not guaranty that the RPC response will be of that type, in fact it is probably a `string`.

That means that we need to convert the json returned by the RPC to match this type see: https://github.com/ecadlabs/tezos-ts/blob/master/packages/tezos-ts-rpc/src/tezos-ts-rpc.ts#L111

In that example the property balance is a BigNumber so we convert the balance property from a json string to a BigNumber doing:
```
return {
   ...contractResponse,
   balance: new BigNumber(contractResponse.balance)
};
```

I suggest that we make a unit test with a real json response and verify that our getConstant actually convert string to big number like it is done here: https://github.com/ecadlabs/tezos-ts/blob/master/packages/tezos-ts-rpc/test/tezos-ts-rpc.spec.ts#L33

Example getConstants response that we can use for the test
```json
{"proof_of_work_nonce_size":8,"nonce_length":32,"max_revelations_per_block":32,"max_operation_data_length":16384,"max_proposals_per_delegate":20,"preserved_cycles":3,"blocks_per_cycle":2048,"blocks_per_commitment":32,"blocks_per_roll_snapshot":256,"blocks_per_voting_period":8192,"time_between_blocks":["30","40"],"endorsers_per_block":32,"hard_gas_limit_per_operation":"400000","hard_gas_limit_per_block":"4000000","proof_of_work_threshold":"70368744177663","tokens_per_roll":"10000000000","michelson_maximum_type_size":1000,"seed_nonce_revelation_tip":"125000","origination_size":257,"block_security_deposit":"128000000","endorsement_security_deposit":"16000000","block_reward":"0","endorsement_reward":"0","cost_per_byte":"1000","hard_storage_limit_per_operation":"60000"}
```

progital:
@carte7000 What do you think about changing snake case to camel case while we are at it?
e.g.
blocks_per_roll_snapshot -> blocksPerRollSnapshot
blocks_per_voting_period -> blocksPerVotingPeriod
etc

carte7000:
I think it's a good idea to change it to camel case while we are just starting.

progital:
so I converted keys to camelCase
added casting to BigNumber
and tests
@carte7000 please reviews it again

packages/tezos-ts-rpc/src/tezos-ts-rpc.ts
```diff
@@ -219,4 +221,38 @@ export class RpcClient {
...
+   * @see http://tezos.gitlab.io/master/api/rpc.html#get-block-id-context-constants
+   */
+  async getConstants({ block }: RPCOptions = defaultRPCOptions): Promise<ConstantsResponse> {
+    const response = await this.httpBackend.createRequest<ConstantsResponse>({
+      url: `${this.url}/chains/${this.chain}/blocks/${block}/context/constants`,
+      method: 'GET',
+    });
+
+    const convResponse: any = camelCaseProps(response);
+    const castedResponse: any = castToBigNumber(convResponse, [
```
carte7000: I like what you did here we should apply this to all the other endpoint method.

PR #22, opened 2019-08-22
feat(rpc): Add getDelegates (baker) endpoint support to RPC

### LINKED ISSUE #21: (RPC) getBaker info
As a developer

I want to get baking information of a delegate

The RPC endpoint to implement support for is http://tezos.gitlab.io/mainnet/api/rpc.html#get-block-id-context-delegates-pkh

Properties needs to be camel case and we need to ensure that all balances are bigNumber

carte7000:
closes #21

### DIFF
--- a/packages/tezos-ts-rpc/src/tezos-ts-rpc.ts
+++ b/packages/tezos-ts-rpc/src/tezos-ts-rpc.ts
@@ -8,6 +8,8 @@ import {
   BigMapGetResponse,
   ManagerResponse,
   DelegateResponse,
+  DelegatesResponse,
+  RawDelegatesResponse,
 } from './types';
 import BigNumber from 'bignumber.js';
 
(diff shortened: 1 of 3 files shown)


packages/tezos-ts-rpc/src/tezos-ts-rpc.ts
```diff
@@ -166,4 +168,39 @@ export class RpcClient {
       key
     );
   }
+
+  /**
+   *
```
jevonearth: Add description? At a minimum we can copy the description from the RPC docs and in some cases improve.

"Everything about a delegate."

"Fetches information about a delegate from RPC"

packages/tezos-ts-rpc/test/tezos-ts-rpc.spec.ts
```diff
@@ -106,6 +106,71 @@ describe('RpcClient test', () => {
     });
   });
 
+  describe('getDelegates', () => {
+    const sampleResponse = {
```
jevonearth: Not important now, but it might be a good idea to name `sampleResponse` with proto number, or use an array (table based test styles).

Data shape might change in future protocol upgrades. 
This does not need to be addressed now, but worth thinking about for the future. May also be YAGNI :stuck_out_tongue_winking_eye:

carte7000: That's a good point maybe we should capture that into an issue?

jevonearth: Good idea, created issue here: #27

PR #48, opened 2019-09-06
feat(utils): Introduce @tezos-ts/utils which contain utility function  for encoding and decoding

carte7000:


packages/tezos-ts-utils/src/tezos-ts-utils.ts
```diff
@@ -0,0 +1,258 @@
...
+    };
+
+    return b58cencode(value.substring(4), prefix[value.substring(0, 4)]);
+  }
+
+  return b58cencode(value.substring(2, 42), new Uint8Array([2, 90, 121]));
+}
+
+export const hex2buf = (hex: string): Uint8Array =>
+  // @ts-ignore
```
jevonearth: Can you give a reason for the ignore. Is it a good reason, or a "fix it later" thing?

packages/tezos-ts-utils/src/tezos-ts-utils.ts
```diff
@@ -0,0 +1,258 @@
...
+    for (let i = 0; i < byteArray.length; i++) {
+      let hex = byteArray[i].toString(16);
+      let paddedHex = ('00' + hex).slice(-2);
+      hexParts.push(paddedHex);
+    }
+    return hexParts.join('');
+  };
+
+  const prefix = {
+    [new Uint8Array([6, 161, 159]).toString()]: '0000',
```
jevonearth: Instead of using literals here, using constants with a descriptive name will make the code a lot more readable and easier to grok.

packages/tezos-ts-utils/src/tezos-ts-utils.ts
```diff
@@ -0,0 +1,258 @@
...
+  n.set(prefix);
+  n.set(payloadAr, prefix.length);
+
+  return bs58check.encode(Buffer.from(n.buffer));
+}
+
+export const b58cdecode = (enc: string, prefixArg: Uint8Array): Uint8Array =>
+  bs58check.decode(enc).slice(prefixArg.length);
+
+export function b58decode(payload: string) {
```
jevonearth: Let's put some time into adding comments to these methods so they don't look so scary to new comers.

PR #49, opened 2019-09-06
feat(signer): Introduce signer package to provide signing capabilities during operation injections

carte7000:


packages/tezos-ts-signer/src/tezos-ts-signer.ts
```diff
@@ -0,0 +1,76 @@
+import sodium from 'libsodium-wrappers';
+import { hex2buf, mergebuf } from '@tezos-ts/utils';
+import toBuffer from 'typedarray-to-buffer';
+import { Tz1 } from './ed-key';
+import { Tz2, ECKey, Tz3 } from './ec-key';
+import pbkdf2 from 'pbkdf2';
+
+/**
+ * @description A local implementation of the signer. Will represent a Tezos account and be able to produce signature in it's behalf
```
jevonearth: ...signature *on its* behalf..

Can you add "InMemorySigner is to assist in development and not intended for signing in a production context"

PR #50, opened 2019-09-06
feat(tezos-ts): Provide high level function to inject operation into node

carte7000:
Also refactor the way we propagate rpc and signer through the library

### DIFF
--- a/packages/tezos-ts/src/context.ts
+++ b/packages/tezos-ts/src/context.ts
@@ -2,6 +2,9 @@ import { RpcClient } from '@tezos-ts/rpc';
 import { Signer } from './signer/interface';
 import { NoopSigner } from './signer/noop';
 
+/**
+ * @description Encapsulate common service used throughout different part of the library
+ */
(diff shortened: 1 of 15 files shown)


packages/tezos-ts/src/context.ts
```diff
@@ -0,0 +1,30 @@
+import { RpcClient } from '@tezos-ts/rpc';
+import { Signer } from './signer/interface';
+import { NoopSigner } from './signer/noop';
+
+export class Context {
```
jevonearth: add docs

packages/tezos-ts/src/contract/interface.ts
```diff
@@ -1,4 +1,93 @@
 import { Schema } from '@tezos-ts/michelson-encoder';
+import { Operation } from './operations';
+import { OperationObject } from '@tezos-ts/rpc';
+
+export interface OriginateParams {
```
jevonearth: add docs

PR #59, opened 2019-09-11
feat(tezos-ts): Add new contract methods abstraction

carte7000:


packages/tezos-ts/src/contract/contract.ts
```diff
@@ -0,0 +1,131 @@
...
+import { ContractProvider } from './interface';
+
+interface SendParams {
+  fee?: number;
+  storageLimit?: number;
+  gasLimit?: number;
+  amount?: number;
+}
+
+class ContractMethod {
```
jevonearth: Needs docs

packages/tezos-ts/src/contract/contract.ts
```diff
@@ -0,0 +1,131 @@
...
+  } else {
+    return 1;
+  }
+}
+/**
+ * @description Smart contract abstraction
+ */
+export class Contract {
+  /**
+   * @description Contains smart contract method and allow to call them
```
jevonearth: Better description.
```
@description Contains methods that are implemented by the target Tezos Smart Contract and offers the user to call the Smart Contract methods as if they were native TS/JS methods. 
```

PR #126, opened 2019-09-24
Patching angular webpack config at post-install

mbykovskyy:


carte7000:
For this to work properly I think we would need to move this patch to the taquito sub-package https://github.com/ecadlabs/taquito/tree/angular-webpack-config-patching/packages/taquito

Also we would need to add the patch.js file to the `files` property of the package.json

"files": [
  "signature.json",
  "dist",
  "patch.js"
],

mbykovskyy:
Good catch, Simon!

PR #136, opened 2019-10-03
fix(taquito): Include new baby2 protocol hash

carte7000:


packages/taquito/src/constants.ts
```diff
@@ -18,6 +18,9 @@ export enum DEFAULT_STORAGE_LIMIT {
 }
 
 export const protocols = {
-  '004': 'Pt24m4xiPbLDhVgVfABUjirbmda3yohdN82Sp9FeuAXJ4eV9otd',
-  '005': 'PsBABY5HQTSkA4297zNHfsZNKtxULfL18y95qb3m53QJiXGmrbU',
+  '004': ['Pt24m4xiPbLDhVgVfABUjirbmda3yohdN82Sp9FeuAXJ4eV9otd'],
+  '005': [
```
jevonearth: I like this. 
How do you feel about adding `name` or `vanity_name` as a property?

`Babylon 2.0.1` is `PsBabyM1eUXZseaJdmXFApDSBqj8YBfwELoxZHHW77EMcAbbwAS`
`Babylon 2.0` is `PsBABY5HQTSkA4297zNHfsZNKtxULfL18y95qb3m53QJiXGmrbU`

and so forth.

This would be both useful for the developer and also give the dev a trivial way to surface the name of a protocol should they wish to do so.

PR #141, opened 2019-10-08
Refactor context to contain protocol

carte7000:


packages/taquito/src/constants.ts
```diff
@@ -17,10 +17,13 @@ export enum DEFAULT_STORAGE_LIMIT {
   REVEAL = 300,
 }
 
+export enum Protocols {
+  Pt24m4xi = 'Pt24m4xiPbLDhVgVfABUjirbmda3yohdN82Sp9FeuAXJ4eV9otd',
```
carte7000: @jevonearth This is related to what suggested on a previous PR

PR #188, opened 2019-12-02
feat(subscription): Add operation subscription with filter on source/destination/hash/kind

### LINKED ISSUE #186: Streaming implementation that polls RPC
Initial implementation will poll the RPCs, and as such not require middle-ware. This implementation offers a low cost/barrier to entry, but the network overhead will be higher and requires more round-trips to the RPC node.

Users can start out with this implementation to get up and running, and if the network overhead becomes a burden, they can choose to switch to using the streaming middleware implementation when it is ready.

Filter options;

- src
- destination
- kind
- hash

carte7000:
RPC based stream provider now supports operation stream with filter on hash/source/destination/kind

- SubscribeProvider interfaces are now exported from taquito package
- RPC based stream provider have been refactored to use rxjs
- Mark WebSocket based streamer package as temporary unmaintained

### BREAKING CHANGES
- WS streamer package is now in an unmaintained state
- Subscription type from SubscribeProvider is now generic might break some implementation of SubscribeProvider implementation

closes #186


carte7000:
An example on how to use the API need to be added before merging the PR

Issue #160, opened 2019-10-25
SPIKE: Design Better error pattern


carte7000:
As of v5.1.0 Taquito's errors are not well defined and are inconsistent.

- [ ] We must define a strategy for our error schema and make them part of our API.
For instance, some error from the RPC are just plain object and other packages returns an Error instance. By applying a consistent approach to errors, the developer experience will improve.
- [ ] Double check if errors are informative and actionable
- [ ] Create follow-up issues based on the design.

jevonearth:
Some progress made in 5.2.0-beta.1 but still more to do.

carte7000:
#225 is related

Innkst:
#363 is related

Innkst:
related #1091

dsawali:
Some custom error classes are not documented properly in Typedoc, we should make an improvement to better explain what the error classes are.

dsawali:
created #1472 to do preliminary evaluation on our error handling

dsawali:
created #1506 as a continuation to #1472

roxaneletourneau:
Error improvement has been done in version 12.1. 
Keeping the issue open to validate if all the comments have been addressed.

roxaneletourneau:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Taquito-Development-Team-60ca841c87ca580010057707/issues/ecadlabs/taquito/160?planning-poker) with Zenhub @dsawali @hui-an-yang @zainen

roxaneletourneau:
Here is a plan to improve the errors: https://hackmd.io/oxQqthGlRnWEk8hUADoZgg?both

roxaneletourneau:
Follow up issues created

PR #211, opened 2019-12-17
Add validation error in michelson encoder

carte7000:
Add validation error during storage/parameter encoding

Check if the value passed to encoder is of the proper type, before we were relying on the RPC to surface these kind of error.

Example of error that might be thrown:

A string `test` instead of a number
```
IntValidationError: [balance] Value is not a number: test
```

A string value instead of an object for a map token
```
MapValidationError: [allowances] Value must be an object
```
A string `test` instead of a number (annotation cannot be surfaced)
```
IntValidationError: [0] Value is not a number: test
```

re #160


packages/taquito-michelson-encoder/test/tokens/bytes.spec.ts
```diff
@@ -0,0 +1,40 @@
+import { BytesToken, BytesValidationError } from '../../src/tokens/comparable/bytes';
+
+describe('Byt5es token', () => {
```
jevonearth: typo?

Issue #168, opened 2019-11-01
Remove support for Athens 004


jevonearth:
Now that Athens is no longer on `mainnet`, and our test nets now run Babylon, we can remove support for injections of Athens operations.

The ability to read blocks, ops etc. that were produced on the Athens protocol will continue to work.

carte7000:
We also need to remove documentation that are Athens specific

PR #219, opened 2019-12-31
feat(taquito): Remove support for athens and add support for carthage

### LINKED ISSUE #207: Add support for proto 006 Carthagenet
Here is the list of changes that might affect Taquito based on: https://tezos.gitlab.io/protocols/006_carthage.html#changes-to-rpcs

- [x] Changes to baking_rights RPC
- [x] The gas limit per block and per operation was increased by 30%

carte7000:
BREAKING CHANGE: Support for Athens has been dropped

closes #207 and #168


packages/taquito-rpc/src/types.ts
```diff
@@ -58,9 +50,9 @@ interface Frozenbalancebycycle {
 
 export type BigMapKey = { key: { [key: string]: string }; type: { prim: string } };
 
-export type ContractResponse = ContractResponse004 | ContractResponse005;
+export type ContractResponse = ContractResponse005;
```
jevonearth: Is `ContractResponse005` copatible with Carthagnet `006`? 
Would it be reasonable/useful (to the reader of the code) to alias `ContractResponse005` as `ContractResponse006`? It would have no functional value for the code but might make the code more literate. I don't feel strongly about this, it's for discussion.

carte7000: I agree, I remove all suffix for RPC types as there should be no difference between 005 and 006 response now

PR #226, opened 2020-01-10
108 local forging of operation

carte7000:
Local forging package provide encoding/decoding of Tezos operation from binary to JSON format and from JSON to binary.

This adds local forging as an additional config for our integration tests
As of we test against our 2 supported node + (remote and local forging for each) so 4 tests run

Local forging can be used as follow:

```ts
import { TezosToolkit } from '@taquito/taquito'
import { localForger } from '@taquito/local-forging'
Tezos.setProvider({ forger: localForger })
```


packages/taquito-local-forging/test/michelson/int.spec.ts
```diff
@@ -0,0 +1,43 @@
+import { intEncoder, intDecoder } from '../../src/michelson/codec';
+
+import { Uint8ArrayConsummer } from '../../src/uint8array-consummer';
+
+import { fromHexString } from '../utils';
+
+describe('Int', () => {
```
jevonearth: Should we add tests for lower/upper bounds of int?

PR #234, opened 2020-01-17
Add reasonML bindings

abate:
This is still work in progress, we need to add more tests and integrate these tests with the github CI


jevonearth:
Happy to reopen this if the bindings get updated.

Issue #108, opened 2019-09-20
Local Forging of Operations


carte7000:
As of v5.1.0, Taquito supports the forging of operations by calling the RPC forge method. We wish to support local forging, so developers do not need to rely on a remote node, due to the associated security and performance implications.

Adding an opt-in package `@taquito/forge` could provide flexibility to our users.

PR #238, opened 2020-01-21
feat(taquito): Add the composite forger

carte7000:
Composite forger allow to use multiple forger implementation in parallel and assert they all return
the same result

re #108


integration-tests/config.ts
```diff
@@ -43,6 +45,9 @@ export const CONFIGS: ConfigWithSetup[] =
       const Tezos = new TezosToolkit();
       if (forger === ForgerType.LOCAL) {
         Tezos.setProvider({ rpc, forger: localForger })
+      } else if (forger === ForgerType.COMPOSITE) {
+        const composite = new CompositeForger([new RpcForger(Tezos['_context']), localForger]);
```
carte7000: We should provide a better of creating instance of the RpcForger as it depends on the `context` which is not exposed to the end user

PR #263, opened 2020-02-22
docs(indexer): Mark indexerclient as deprecated, and remove example f…

jevonearth:
…rom quickstart docs

### DIFF
--- a/packages/taquito-indexer/src/taquito-indexer.ts
+++ b/packages/taquito-indexer/src/taquito-indexer.ts
@@ -10,14 +10,15 @@ export interface BalanceHistoryOptions {
 }
 
 /***
- * @description RpcClient allows interaction with Tezos network through an rpc node
+ * @description IndexerClient is a client to the server counterpart https://github.com/ecadlabs/tezos-indexer-api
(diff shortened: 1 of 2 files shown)


packages/taquito-indexer/src/taquito-indexer.ts
```diff
@@ -10,14 +10,15 @@ export interface BalanceHistoryOptions {
 }
 
 /***
- * @description RpcClient allows interaction with Tezos network through an rpc node
+ * @description IndexerClient is a client to the server counterpart https://github.com/ecadlabs/tezos-indexer-api
+ * @deprecated The IndexerClient is currently unmaintained and should be avoided until https://github.com/ecadlabs/taquito/issues/185 is addressed
```
carte7000: Perhaps we can just remove it along with the query API in Taquito? It seems like a low risk thing to do as I believe that https://api.tez.ie/indexer/mainnet is not functional anymore?

jevonearth: Agree

jevonearth:
Closing this PR, submitting a new one with full removal of the indexer package.

PR #271, opened 2020-02-25
fix(local-forging): Validate entrypoint name while encoding/decoding

### LINKED ISSUE #265: Entrypoint name encoding can violate Tezos specifications
Hi! Doyensec has been engaged to perform a security assessment of this library. I will be opening issues to document our findings.

## Description
Tezos transactions can define custom entrypoints starting from protocol version 5. According to [the official documentation](https://tezos.gitlab.io/protocols/005_babylon.html#transactions-now-have-an-entrypoint), custom entrypoints maximum length is 31 characters, but taquito does not enforce this limitation when encoding or decoding transactions.

This is an extract from the documentation:
```
The entrypoint format is as follows:
one byte 0 for entrypoint %default
one byte 1 for entrypoint %root
[...]
one byte 255 for a named entrypoint,  then one byte of entrypoint name size (limited to 31), and the name itself
```

This is the relevant code from codec.ts:
```
export const entrypointDecoder = (value: Uint8ArrayConsumer) => {
  const preamble = pad(value.consume(1)[0], 2);

  if (preamble in entrypointMapping) {
    return entrypointMapping[preamble];
  } else {
    const entry = extractRequiredLen(value, 1);
    return Buffer.from(entry).toString('utf8');
  }
};

export const entrypointEncoder = (entrypoint: string) => {
  if (entrypoint in entrypointMappingReverse) {
    return `${entrypointMappingReverse[entrypoint]}`;
  } else {
    const value = { string: entrypoint };
    return `ff${valueEncoder(value).slice(8)}`;
  }
};
```

Observe that no checks are performed on the length of the entrypoint name being encoded or decoded. You might want to consider enforcing the size limit required by the specification.

&lt;sales pitch&gt;If you’re looking for an independent vendor to perform security testing or to develop security automation solutions, let us know! [https://doyensec.com](https://doyensec.com)&lt;/sales pitch&gt;

carte7000:
fix #265


packages/taquito-local-forging/src/constants.ts
```diff
@@ -5,6 +5,8 @@
 
 import { pad } from './utils';
 
+export const ENTRYPOINT_MAX_LENGTH = 31;
```
jevonearth: Add a comment with a link to https://tezos.gitlab.io/protocols/005_babylon.html#transactions-now-have-an-entrypoint

useful to the reader to easily jump to the definition.

PR #301, opened 2020-03-07
Make sure undefined is thrown when michelson map key is not found

### LINKED ISSUE #277: BigMap.get with passing a defaultValue as param
Can we do something similar to `myMap.get(key, defaultValue = None)` of smartPy with Taquito?
An instruction like: 

```
const myContract = await Tezos.contract.at(contractAddress);
const contractStorage = await myContract.storage();
const value = await contractStorage.myMap.get(key, defaultValue = 0)
```

Thnak you.

Sm00g15:
fix #277

### DIFF
--- a/packages/taquito-michelson-encoder/src/michelson-map.ts
+++ b/packages/taquito-michelson-encoder/src/michelson-map.ts
@@ -116,7 +116,7 @@ export class MichelsonMap<K extends MichelsonMapKey, T extends any> {
     this.assertTypecheckKey(key);
 
     const strKey = this.serializeDeterministically(key);
-    return this.valueMap.get(strKey);
+    return this.valueMap.get(strKey) ? this.valueMap.get(strKey) : undefined;
   }
 
   /**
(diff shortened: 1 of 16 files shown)


packages/taquito-michelson-encoder/src/michelson-map.ts
```diff
@@ -116,7 +116,7 @@ export class MichelsonMap<K extends MichelsonMapKey, T extends any> {
     this.assertTypecheckKey(key);
 
     const strKey = this.serializeDeterministically(key);
-    return this.valueMap.get(strKey);
+    return this.valueMap.get(strKey) ? this.valueMap.get(strKey) : undefined;
```
AndrewKishino: This can be problematic if we assume the test case:
```
it('returns the value for an existing key', () => {
  const map = new MichelsonMap();
  map.set('test', 0);
  expect(map.get('test')).toEqual(0);
});
```
Since all falsely values will default to `undefined` in this case which would be incorrect.

Sm00g15: ah yes, good catch @AndrewKishino !  In that case I think the code that's already implemented should do the trick. No PR needed

PR #308, opened 2020-03-10
WIP: Tzip 10 wallet provider api

carte7000:


packages/taquito-beacon-wallet/src/taquito-beacon-wallet.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  private removeFeeAndLimit<T extends { gas_limit: any, storage_limit: any, fee: any }>(op: T) {
+    const { fee, gas_limit, storage_limit, ...rest } = op;
+    return rest;
+  }
+
+  async originate(contract: WalletOriginateParams): Promise<OriginationWalletOperation> {
+    const op = await createOriginationOperation(contract as any)
+    const network = this.permissions!.permissions.networks;
```
jevonearth: We will need to handle the permissions here.

packages/taquito-beacon-wallet/src/taquito-beacon-wallet.ts
```diff
@@ -0,0 +1,74 @@
...
+  message = 'You need to initialize BeaconWallet by calling beaconWallet.requestPermissions first';
+}
+
+export class BeaconWallet implements WalletProvider {
+  public client: DAppClient;
+
+  private permissions?: PermissionResponse;
+
+  constructor(name: string) {
+    this.client = new DAppClient(name);
```
carte7000: We need to provide more option to configure the dApp client. Probably allowing the user to pass in the DAppClient instance would be better.

AndreasGassmann: Yes, there will probably be more options to configure the dApp client soon (eg. disabling standard behaviour like showing the QR code alert).

What if we introduced a `DAppClientOptions` object that can be passed in the constructor, so you can use the same type in your constructor?

carte7000: Good idea, let's do it that way :+1:

Issue #315, opened 2020-03-12
Remove Tezos.importKey()


jevonearth:
As part of #212 `Tezos.importKey()` has been moved out of the `@taquito/taquito` package. In v7, remove `Tezos.importKey()` method completely.

carte7000:
We need to make sure we add the type back here https://github.com/ecadlabs/taquito/pull/316/files#diff-1d42e4b6cd2de739e9eabea6224c8b99R14

And change the dependency direction between @taquito/taquito and @taquito/signer

PR #316, opened 2020-03-13
fix(signer): Fix merge issues from previous commit

AndrewKishino:
Resolve cyclical dependency in @taquito/signer and fix dependencies from previous rebase in order to fix build issues.

The previous commit introduced a cyclical dependency. This commit should resolve that and reset the dependencies that were outdated from a rebase of master. The only issue is that in `import-key.ts` toolkit does not explicitly set a type.
```
export async function importKey(
  toolkit: any,
  privateKeyOrEmail: string,
  passphrase?: string,
  mnemonic?: string,
  secret?: string
) {
```

Any ideas on how we can add a type to toolkit instead of `any`?

### DIFF
--- a/packages/taquito-signer/src/import-key.ts
+++ b/packages/taquito-signer/src/import-key.ts
@@ -1,5 +1,4 @@
 import { InMemorySigner } from './taquito-signer';
-import { TezosToolkit } from '@taquito/taquito';
 
 /**
  *
@@ -12,7 +11,7 @@ import { TezosToolkit } from '@taquito/taquito';
  * @param secret Faucet secret
  */
 export async function importKey(
-  toolkit: TezosToolkit,
+  toolkit: any,
   privateKeyOrEmail: string,
   passphrase?: string,
   mnemonic?: string,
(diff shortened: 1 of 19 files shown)


packages/taquito-signer/src/import-key.ts
```diff
@@ -12,15 +11,15 @@ import { TezosToolkit } from '@taquito/taquito';
  * @param secret Faucet secret
  */
 export async function importKey(
-  toolkit: TezosToolkit,
+  toolkit: any,
```
AndrewKishino: Not ideal to set this to any

carte7000: I agree with you, however, it seems like would just be necessary while we wait for removing the signer package from the main taquito package. 

Maybe an acceptable solution is to put a TODO over this and add a new task in this issue https://github.com/ecadlabs/taquito/issues/315, That way we can remember to remove it and put back the type.

What do you think?

jevonearth: I think this is acceptable for the short term.

We can also consider removing the method (#315) in 6.2 instead of waiting for v7.

Issue #275, opened 2020-02-27
Use default instead of main for smart contract default entrypoint


carte7000:
Currently when a smart contract only have 1 entrypoint (default) Taquito will create a synthetic method called main. This is confusing because it really refers to the default entrypoint in Michelson

http://tezos.gitlab.io/whitedoc/michelson.html#the-default-entrypoint

This issue is about change this synthetic entrypoint name from `main` to `default`

This will cause breaking change with implementation that use the current `main` methods

jevonearth:
If or when we make this change, we must update this SE question: https://tezos.stackexchange.com/questions/2234/taquito-default-entrypoint-not-found

jevonearth:
This will be a breaking change to users of Taquito, but one worth making. Let's make this change on either 6.2 or v7.

PR #317, opened 2020-03-14
refactor(contract.ts): change default entrypoint to 'default' from 'm…

sean-magin:
BREAKING CHANGE: changes default entrypoint from default to main

#275


jevonearth:
Nice work. We will merge this for V7 as it's a breaking change.

jevonearth:
@roxaneletourneau 
Can you draft instructions for the migration notes that will go in the v7 release notes? Make sure to include the error message people will get if they update to v7 without fixing their code `main` -> `default` so that when searching they have a better chance of finding the migration/updating docs.

roxaneletourneau:
> @roxaneletourneau
> Can you draft instructions for the migration notes that will go in the v7 release notes? Make sure to include the error message people will get if they update to v7 without fixing their code `main` -> `default` so that when searching they have a better chance of finding the migration/updating docs.

**Version 7 of Taquito introduces a breaking change for implementations that use the `main` method for calling the `default` entrypoint of a contract.**

Before version 7, there was a synthetic method called `main` in Taquito to interact with the default entrypoint of a contract. This was confusing because it was referring to the [default](http://tezos.gitlab.io/whitedoc/michelson.html#the-default-entrypoint) entrypoint in Michelson.
In version 7, this synthetic entrypoint name has been changed from `main` to `default`. 

**Example:**
The contract `KT1FUF3yapxt5pKGsYgnekrJR4e25oJjYvCp` has a single entrypoint (default). 
**Before version 7**, this is how the default entrypoint was called :
```
const contract = await tezos.contract.at('KT1FUF3yapxt5pKGsYgnekrJR4e25oJjYvCp')
const opMethod = await contract.methods.main("2").send();
```
**In version 7**, the code above needs to be replaced with: 
```
const contract = await tezos.contract.at('KT1FUF3yapxt5pKGsYgnekrJR4e25oJjYvCp')
const opMethod = await contract.methods.default("2").send();
```
If we update to version 7 without replacing `main` with `default` in the code, we will get the following **error at runtime**:
`contract.methods.main is not a function`.

PR #328, opened 2020-03-18
277/bigmap-default-undefined

### LINKED ISSUE #277: BigMap.get with passing a defaultValue as param

Sm00g15:
Closes #277


packages/taquito/test/contract/data.ts
```diff
@@ -111,7 +122,15 @@ export const miSample = [
   { prim: 'parameter', args: [{ prim: 'int', args: [] }] },
   {
     prim: 'storage',
-    args: [{ prim: 'pair', args: [{ prim: 'int', args: [] }, { prim: 'address', args: [] }] }],
```
Sm00g15: This change was enforced by prettier in pre-commit

packages/taquito/test/contract/data.ts
```diff
@@ -385,7 +404,10 @@ export const tokenCode = [
               { prim: 'address', annots: [':from'] },
               {
                 prim: 'pair',
-                args: [{ prim: 'address', annots: [':to'] }, { prim: 'nat', annots: [':value'] }],
```
Sm00g15: This change was enforced by prettier in pre-commit

integration-tests/contract-api-scenario.spec.ts
```diff
@@ -74,6 +74,18 @@ CONFIGS.forEach(({ lib, rpc, setup, knownBaker, createAddress, protocol }) => {
       done();
     })
 
+    it('Return undefined when BigMap key is not found', async () => {
+      const myContract = await Tezos.contract.at(rpc === "https://api.tez.ie/rpc/carthagenet" ? "KT1MLLuyxgxWzW8iRMrNVeamerasADFqiKPW" : "KT1T2KjQdqeNzeaSGm9MfzfgMN8rWC94BrTP");
+      const contractStorage: any = await myContract.storage();
+      let value;
+      if (rpc === "https://api.tez.ie/rpc/carthagenet") {
+        value = await contractStorage[0].get("tz1bwsEWCwSEXdRvnJxvegQZKeX5dj6oKEyz")
```
carte7000: It would probably be easier to use a contract with a similar implementation to avoid different test logic per network.

I deployed a contract with similar storage as the one that is used for babylonnet on carthagenet: https://better-call.dev/carthage/KT1HqWsXrGbHWc9muqkApqWu64WsxCU3FoRf

Here is the link to ligo ide: https://ide.ligolang.org/p/BNv2SLMXnRvKGrB2Q2gjIg this is what I used to deploy this new contract on carthagenet

packages/taquito/src/contract/big-map.ts
```diff
@@ -6,7 +6,12 @@ export class BigMapAbstraction {
   constructor(private id: BigNumber, private schema: Schema, private provider: ContractProvider) {}
 
   async get(keyToEncode: string) {
-    return this.provider.getBigMapKeyByID(this.id.toString(), keyToEncode, this.schema);
+    try {
+      const id = await this.provider.getBigMapKeyByID(this.id.toString(), keyToEncode, this.schema);
+      return id;
+    } catch (e) {
```
carte7000: This seems to return undefined for every all errors that can be thrown by the `this.provider.getBigMapKeyByID` call.

I am not sure if this the behavior described in the ticket. It seems like the correct behavior would be to only return undefined when the key does not exist which would mean that the specific error would be of type `HTTPErrorResponse` see: https://github.com/ecadlabs/taquito/blob/master/packages/taquito-http-utils/src/taquito-http-utils.ts#L21 with a status code of 404

If we do not get this specific error I believe we should still throw the error to the caller and avoid handling it a this level

packages/taquito/test/contract/rpc-contract-provider.spec.ts
```diff
@@ -161,6 +163,15 @@ describe('RpcContractProvider test', () => {
     });
   });
 
+  describe('BigMapAbstraction', () => {
+    it('returns undefined on bad key in BigMap', async done => {
```
carte7000: Do you think it would be interesting to be more explicit with regards to the error that is thrown by rpcContractProvider.getBigMapKeyByID?

For instance, we could have a test for HTTPResponseError where we mock the response using:
`rpcContractProvider.getBigMapKeyByID.mockRejectedValue(new HTTPResponseError(Status.NotFound))`
In this case we would expect undefined

Another test could be mocking the value using:
`rpcContractProvider.getBigMapKeyByID.mockRejectedValue(new Error())`
In this case we would expect the call to throw this error again

integration-tests/contract-api-scenario.spec.ts
```diff
@@ -74,6 +74,18 @@ CONFIGS.forEach(({ lib, rpc, setup, knownBaker, createAddress, protocol }) => {
       done();
     })
 
+    it('Return undefined when BigMap key is not found', async () => {
+      const myContract = await Tezos.contract.at(knownBigMapContract);
+      const contractStorage: any = await myContract.storage();
+      let value;
+      if (rpc === "https://api.tez.ie/rpc/carthagenet") {
```
carte7000: Could this check unnecessary now? I believe we could use the same PKH for both network and expect the `value` to be undefined?

packages/taquito/src/contract/big-map.ts
```diff
@@ -1,12 +1,22 @@
...
+      if (e instanceof HttpResponseError && e.status === STATUS_CODE.NOT_FOUND) {
+        return undefined;
+      } else {
+        throw new Error(e);
```
carte7000: What do you think about just doing `throw e` instead of `throw new Error(e)` so that do not change the Error type?

Sm00g15: I think that's a great idea, I've changed the code and pushing up now.

integration-tests/contract-api-scenario.spec.ts
Sm00g15: @carte7000 do you think we should make a note of this process of a) how to find the necessary smart contract code (in Michelson/Ligo or w/e was used to launch it) and b) quickest way to launch the same contract on a different network?

DalyIng:
Hi @Sm00g15 , Thanks Team Taquito , this feature will be released in the next version?

jevonearth:
@DalyIng Yes :)

PR #333, opened 2020-03-19
fix(taquito): Split reveal operation status from other operation status

carte7000:


packages/taquito/src/operations/operation-emitter.ts
```diff
@@ -258,6 +258,8 @@ export abstract class OperationEmitter {
     const opResponse: OperationContentsAndResult[] = [];
     const results = await this.rpc.preapplyOperations([forgedBytes.opOb]);
 
+    console.log(JSON.stringify(forgedBytes.opOb), JSON.stringify(results));
```
jevonearth: Boop! 😁

PR #337, opened 2020-03-20
fix(signer): Some signatures were invalid (s value too short) for ec key

carte7000:
Implemented a padding solution to ensure that signatures length is 64 bytes as suggested in this
issue https://github.com/indutny/elliptic/issues/195

### DIFF
--- a/packages/taquito-signer/src/ec-key.ts
+++ b/packages/taquito-signer/src/ec-key.ts
@@ -72,10 +72,9 @@ export class ECKey {
   async sign(bytes: string, bytesHash: Uint8Array) {
     const key = new elliptic.ec(this.curve).keyFromPrivate(this._key);
     const sig = key.sign(bytesHash, { canonical: true });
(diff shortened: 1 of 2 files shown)


packages/taquito-signer/src/ec-key.ts
```diff
@@ -72,10 +72,9 @@ export class ECKey {
   async sign(bytes: string, bytesHash: Uint8Array) {
     const key = new elliptic.ec(this.curve).keyFromPrivate(this._key);
     const sig = key.sign(bytesHash, { canonical: true });
-    const signature = new Uint8Array(sig.r.toArray().concat(sig.s.toArray()));
-    const signatureBuffer = toBuffer(signature);
-    const sbytes = bytes + buf2hex(signatureBuffer);
+    const signature = sig.r.toString('hex', 64) + sig.s.toString('hex', 64);
```
carte7000: This should have a comment

PR #341, opened 2020-03-24
Add lambdaView to ContractProvider

lambdahands:
## Description

This PR adds a `ContractProvider.lambdaView()` function that creates an instance of the `LambdaView` class by providing a "lambda" contract whose main method _always_ fails, a target contract and an associated view method (for example `getTotalSupply`).

When executed, the `LambdaView` instance sends an operation that returns a failure containing the result of calling the view method in the `failWith` property.

For example:

```typescript
import { Tezos } from '@taquito/taquito';

async function getTotalSupply() {
  const lambdaContract = "KT1E1trWsE1A9yrbgNeRJC54VCYgEtrbYLSE";
  const viewContract = "KT1RUhPAABRhZBctcsWFtymyjpuBQdLTqaAQ";
  const lambdaView = await Tezos.contract.lambdaView(
    lambdaContract, viewContract, "getTotalSupply"
  );

  return await lambdaView.execute(); // { 'int': 12 }
}
```

## Changes

The biggest changes are to:
- The `ContractProvider` interface, which adds the `lambdaView` function.
- The `Contract` class, which exposes its previously `private entrypoints` as `public readonly entrypoints`. 

## Todo
- [x] Provide default `lambdaContract`
- [x] Write tests


lambdahands:
Running
```
$ npm install
$ lerna bootstrap
$ npm run build
```
resulted in all of the monorepo's `package-lock.json` files to be updated. This has always been an issue I've had trouble with when working with NPM - the new lockfiles _may_ be correct, but if they're not I can revert the commit that changes them and stick with the old ones.

packages/taquito/src/contract/rpc-contract-provider.ts
```diff
@@ -207,4 +208,32 @@ export class RpcContractProvider extends OperationEmitter implements ContractPro
...
+  async lambdaView(
+    lambdaContractOrAddress: Contract | string,
+    viewContractOrAddress: Contract | string,
+    viewMethod: string,
+    entrypointName?: string,
+    contractParameter?: MichelsonV1Expression
+  ): Promise<LambdaView> {
+    let lambdaContract;
+    let viewContract;
+    if (lambdaContractOrAddress instanceof Contract) {
```
carte7000: I believe the use of instanceof here can introduce a problem similar to what we faced in this issue: https://github.com/ecadlabs/taquito/pull/325

For example when a Contract is created by one taquito dependency and consumed by another instanceof will return false. This was the case with Truffle for example.

lambdahands:
Hey, @carte7000 - sorry for the late reply/update from my end. I've changed the implementation a bit thanks to your feedback and added a couple of simple integration tests as well. As for default contracts, I added a "carthagenet" contract address, but I'm unsure about who would be responsible for generating/maintaining a contract on mainnet.

jevonearth:
I think we can originate the contract on mainnet, we will be happy to do that.

Can you;

- add a section to the docs for this feature? Explaining its purpose and any tradeoffs you can think of. Website is typical docusarus.
- Add an origination script for the contract itself. I'll use this to originate it to mainnet, but the real value I see is for sandbox users to be able to originate the contract into their environment so they can make use of this in their local dev workflow.

lambdahands:
Hi, @jevonearth - I've added the lambda contract code to the `michelson-encoder` package (wasn't sure where to put it, let me know if there's a better place for it live). I also added a new docs section in `docs/`; would love any feedback in that area as well. Thanks!

michaeljklein:
@lambdahands @jevonearth We now have a mainnet deploy! https://better-call.dev/mainnet/KT1CPuTzwC7h7uLXd5WQmpMFso1HxrLBUtpE/code

jevonearth:
Great! 

@lambdahands what do you think of the recent suggestions from @roxaneletourneau and myself?

lambdahands:
@jevonearth Hey, they look great - I can hopefully have something ready by the end of this week!

jevonearth:
Closing as superseded by #471

PR #346, opened 2020-04-08
fix(origination): Ensure correct order of code properties

jevonearth:
Some HLL's (SmartPy) produce JSON represented Michelson in an order that is not accepted by the
Tezos Node RPC.

### DIFF
--- a/packages/taquito/src/contract/prepare.ts
+++ b/packages/taquito/src/contract/prepare.ts
@@ -30,7 +30,10 @@ export const createOriginationOperation = async ({
     );
   }
 
-  const contractCode = Array.isArray(code) ? code : ml2mic(code);
+  const order = ['parameter', 'storage', 'code'];
+  const contractCode = Array.isArray(code)
+    ? code.sort((a: any, b: any) => order.indexOf(a.prim) - order.indexOf(b.prim)) // Ensure correct ordering for RPC
+    : ml2mic(code);
 
   let contractStorage: object;
   if (storage !== undefined) {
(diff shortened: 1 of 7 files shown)


packages/taquito/src/contract/prepare.ts
```diff
@@ -30,7 +30,10 @@ export const createOriginationOperation = async ({
     );
   }
 
-  const contractCode = Array.isArray(code) ? code : ml2mic(code);
+  const order = ['parameter', 'storage', 'code'];
+  const contractCode = Array.isArray(code)
+    ? code.sort((a: any, b: any) => order.indexOf(a.prim) - order.indexOf(b.prim)) // Ensure correct ordering for RPC
+    : ml2mic(code);
```
AndrewKishino, line 36: I guess my one concern here is that ml2mic produces the same order that is being defined and if not, `const schema = new Schema(contractCode[1].args[0])` could have an issue as well.

AndrewKishino: This needs to be validated, perhaps something like this would cover it:
```
const contractCode = Array.isArray(code)
    ? code.sort((a: any, b: any) => order.indexOf(a.prim) - order.indexOf(b.prim))
    : ml2mic(code).sort((a: any, b: any) => order.indexOf(a.prim) - order.indexOf(b.prim));
```

jevonearth: Hmm. From testing with `https://smartpy.io/demo/` they output the correct order when producing normal/naked Michelson, but when producing JSON, the order is incorrect.

I kinda feel like the `ml2mic` case is low risk. This issue probably originates from son encoders that don't preserve order.

PR #368, opened 2020-05-20
Update all deps including Typescript to 3.9

jevonearth:
## Release Note Draft Snippet

Taquito is now built using Typescript 3.9!


packages/taquito/src/contract/index.ts
```diff
@@ -1,4 +1,3 @@
 export * from './errors';
 export * from './interface';
 export * from './manager-lambda';
-export { UnitValue } from '@taquito/michelson-encoder';
```
jevonearth, line 4: `UnitValue` is already exported from `src/taquito.ts`

jevonearth:
The docusarus website build fails after these updates. I'm working on fixing this. Not sure what the root issue is. I have tested with newer versions of docusarus and all lead to the same error;

`TypeError: w is not a function`

After some webpack build activity.

PR #373, opened 2020-06-01
refactor(remote-signer): verify signature from remote-signer

AndrewKishino:
On receipt of signature from remote signer, validate the signature against the public key. Also
validate that the public key returned from the remote signer corresponds to the public key hash used to initialize the remote signer.

- Added a new `verify` method which takes the bytes and signature from the remote signer and validates it against the public key to ensure that the signature is indeed valid. This method also verifies that the public key being returned from the server matches the public key hash used to initialize the remote signer.

### DIFF
--- a/packages/taquito-remote-signer/package-lock.json
+++ b/packages/taquito-remote-signer/package-lock.json
@@ -894,11 +894,29 @@
 				"@babel/types": "^7.3.0"
 			}
 		},
+		"@types/bn.js": {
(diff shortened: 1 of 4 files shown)


jevonearth:
Looks great.
Can we get some unit test for these?
We could also create integration tests against a signatory docker instance.

AndrewKishino:
Sure, I will add some unit tests to this PR

PR #382, opened 2020-06-15
Documentation for validation functions of Taquito-utils

roxaneletourneau:
Added a demo page for the validation functions to Taquito website and added examples/descriptions to TypeDoc

### DIFF
--- a/docs/validators.md
+++ b/docs/validators.md
@@ -3,7 +3,7 @@ title: Validation functions
 author: Roxane Letourneau
 ---
 
-Taquito provides functions that allow us to see if an address, a chain, a key hash, a contract address, a public key or a signature is valid. The `ValidationResult` returned by these functions is an enum which can take the following values:
+Taquito provides functions that allow us to see if an address, a chain, a key hash, a contract address, a public key or a signature is valid. Note that these validations do not rely on a node but are done based on checksums. Thus, they allow us to check if a value is valid and not if it exists on a chain. The `ValidationResult` returned by these functions is an enum which can take the following values:
   ```
   0 = NO_PREFIX_MATCHED,
   1 = INVALID_CHECKSUM,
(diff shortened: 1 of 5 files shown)


docs/validators.md
```diff
@@ -0,0 +1,146 @@
+---
+title: Validation functions
+author: Roxane Letourneau
+---
+
+Taquito provides functions that allow us to see if an address, a chain, a key hash, a contract address, a public key or a signature is valid. The `ValidationResult` returned by these functions is an enum which can take the following values:
```
jevonearth, line 6: It would be worth explaining that the validation is done based on checksums to see if it is valid, and that it does not rely on a node, or tell the user if the address/key hash/block hash actually exists on a chain.

docs/validators.md
```diff
@@ -0,0 +1,146 @@
...
+  ```
+  0 = NO_PREFIX_MATCHED,
+  1 = INVALID_CHECKSUM,
+  2 = INVALID_LENGTH,
+  3 = VALID
+  ```
+
+### Validate an address
+
+This function can be used to validate addresses as well as contract addresses. Calling the `validateAddress` function with the public key hash (pkh) from a faucet (https://faucet.tzalpha.net/) will return valid :
```
jevonearth, line 16: I don't think mentioning the faucet here adds value.

docs/validators.md
```diff
@@ -0,0 +1,146 @@
...
+Taquito provides functions that allow us to see if an address, a chain, a key hash, a contract address, a public key or a signature is valid. The `ValidationResult` returned by these functions is an enum which can take the following values:
+  ```
+  0 = NO_PREFIX_MATCHED,
+  1 = INVALID_CHECKSUM,
+  2 = INVALID_LENGTH,
+  3 = VALID
+  ```
+
+### Validate an address
+
```
jevonearth, line 15: It would be good to group

- validateAddress (Will validate an implicit (tz1/2/3 or a KT1 address))
- validateKeyHash (Will validate only implicit addresses (tz1/2/3))
- validateContractAddress (Will validate only contract address (KT1))

together, and explain why they are different. 

The naming of these methods could be better, but that's a different matter.

docs/validators.md
```diff
@@ -0,0 +1,146 @@
...
+```js live noInline
+import { validateAddress } from '@taquito/utils';
+
+const pkh = 'tz1L9r8mWmRPndRhuvMCWESLGSVeFzQ9NAWx'
+const validation = validateAddress(pkh)
+println(validation)
+    
+```
+
+Calling the `validateAddress` function with the previous pkh missing its last character will result in an invalid checksum :
```
jevonearth, line 27: Instead of making the address too short, better to make it invalid, as this will impress upon the reader that there's actually an integrity check. 

For example:
`tz1L9r8mWmRPndRhuvMCWESLGSVEFzQ9NAWx` <-- Invalid
`tz1L9r8mWmRPndRhuvMCWESLGSVeFzQ9NAWx` <-- Valid

PR #387, opened 2020-06-25
Added documentation for map and bigMap

roxaneletourneau:
Created a documentation page on the website for map and bigMap.
Added examples where the MichelsonMap class and its get/set methods are used to originate contracts with initial values and to access these values.
Added examples where the key is a pair.


jevonearth:
Replaced by #389

PR #431, opened 2020-09-09
fix(subscribe): subscription should not die on rpc error

### LINKED ISSUE #430: Subscription dies when getBlock rpc call fails
When the getBlock rpc call fails, the observable in PollingSubscribeProvider dies and no event will be reported after that.
I'm submitting a PR fixing this in a minute.

julienmuller-forge:
currently, the observable in ObservableSubsription dies when getBlock rpc call fails. This change
makes it report the error via the error callback, but it continues to poll for new blocks

fix #430

- [X] Your code builds cleanly without any errors or warnings
- [NA] You have added unit tests
- [NA] You have added integration tests (if relevant/appropriate)
- [NA] All public methods or types have TypeDoc coverage with a complete description, and ideally an @example
- [NA] You have added or updated corresponding documentation
- [NA] If relevant, you have written a first draft summary describing the change for inclusion in Release Notes. 

## Release Note Draft Snippet

### DIFF
--- a/example/example-streamer.ts
+++ b/example/example-streamer.ts
@@ -3,7 +3,7 @@ import { Tezos } from '@taquito/taquito';
 
 async function example() {
   const provider = 'https://api.tez.ie/rpc/mainnet';
-  Tezos.setProvider({ rpc: provider });
+  Tezos.setProvider({ rpc: provider, config: {shouldObservableSubscriptionRetry: true} });
   try {
 
     const bakerEndorsementFilter = {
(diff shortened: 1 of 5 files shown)


jevonearth:
Thank you @julienmuller-forge !

roxaneletourneau:
Great work @julienmuller-forge! Would it be possible to add a unit test to prevent regression? Thank you and great catch on this.

roxaneletourneau:
> Great work @julienmuller-forge! Would it be possible to add a unit test to prevent regression? Thank you and great catch on this.

@julienmuller-forge Just to let you know, so that we don't duplicate work, that I am writing the test right now. Thanks again for your contribution.

julienmuller-forge:
@roxaneletourneau , ok great ! I had started to look into it but haven't had much time to do so.
FYI, i was thinking the best way to do it was to just test ObservableSubscription with a specially crafted observable that both emits values and also throws errors, and check that the subscription is not interrupted by errors.

PR #443, opened 2020-10-15
Remove default RPC address

### LINKED ISSUE #388: Remove default RPC address so that users make their own choice
# Problem

Taquito is configured to use a default RPC node. We took this approach so that users can get started quickly, and Taquito should "just work" with minimal fiddling. 

The problem is that users of Taquito can easily punt on the decision and assume the default is safe. This leads to a lot of applications relying on our default node and thus centralising a part of the ecosystem on one node. There's more problems with offering this default node, but that's the gist of it. 

# Action

* Remove the default node configuration, forcing the user to set their own
* Implement a helpful error message so when users don't set it, they get some clear hints on the what/why/how to proceed
* Create a documentation page with links to community-run nodes and to instructions on how to run your own node. Explain the tradeoffs and encourage the Taquito user to DYOR
* Make the change prominent in release announcements so as not to blindside users with this breaking change.

# Alternative

An alternative would have been to replace the default URL (https://api.tez.ie/rpc/carthagenet) with an unopinionated value as http://localhost:8732/, which is the default value when running a node locally. It would have reduced the magnitude of the breaking change by only impacting applications that actually use the default RPC URL. All applications that currently configure their own RPC by doing setProvider({rpc: ...}) or setRpcProvider(..) would not have been affected by the change.

However, we decided to go with the other solution because setting the URL to http://localhost:8732/ does not satisfy many use cases. Also, changing the default URL to localhost would not warn people using the default with compile-time errors; they would get a connection refused error from localhost.

roxaneletourneau:
Fixes #388 
- Removed the default RPC URL in the RpcClient class. It is now required to specify an RPC node in the constructor of the RpcClient, Context, and TezosToolkit.
- The Tezos singleton has been replaced by a function that takes an RPC as a parameter:
```
import { Tezos } '@taquito/taquito';
const tezos = Tezos('https://YOUR_PREFERRED_RPC_URL');
```
- Added integration tests to show what methods in the taquito RPC package, which maps to the RPC endpoints, work on which nodes (Carthagenet so far).
- Added a documentation page on the website about RPC nodes.
- Removed Tezos singleton on the documentation pages of the website.

-Also, I removed the React-component package


jevonearth (approved):
:muscle:

PR #447, opened 2020-10-20
Pr/360

roxaneletourneau:
Added the Missing properties on types ConstantsResponse and corrected some types.
Removed the properties from ConstantsResponse that are no longer return when querying the constants from the RPC node.
Added some test cases in taquito-rpc.spec.ts.


packages/taquito-rpc/src/types.ts
```diff
@@ -581,8 +581,10 @@ export interface OperationContentsAndResultMetadataOrigination {
 export interface ConstantsResponse {
   proof_of_work_nonce_size: number;
   nonce_length: number;
-  max_revelations_per_block: number;
+  max_revelations_per_block?: number; // carthagenet
```
jevonearth, line 584: @roxaneletourneau what do you think about making integration tests that fetch the constant values from a block-level for each protocol? Below is a list of block levels for protos 0 to 6.

|Protocol Hash                                         | Proto Number | Level   |
|------------------------------------------------------|--------------|---------|
|`PrihK96nBAFSxVL1GLJTVhu9YnzkMFiBeuJRPA8NwuZVZCE1L6i` | `0`          | `0`     |
|`PtCJ7pwoxe8JasnHY8YonnLYjcVHmhiARPJvqcC6VfHT5s8k8sY` | `1`          | `10000` |
|`PsYLVpVvgbLhAhoqAkMFUo6gudkJ9weNXhUYCiLDzcUpFpkk8Wt` | `2`          | `100000`|
|`PsddFKi32cMJ2qPjf43Qv5GDWLDPZb3T3bF6fLKiF5HtvHNU7aP` | `3`          | `300000`|
|`Pt24m4xiPbLDhVgVfABUjirbmda3yohdN82Sp9FeuAXJ4eV9otd` | `4`          | `600000`|
|`PsBabyM1eUXZseaJdmXFApDSBqj8YBfwELoxZHHW77EMcAbbwAS` | `5`          | `700000`|
|`PsCARTHAGazKbHtnKfLzQg3kms52kSRpgnDY982a9oYsSXRLQEb` | `6`          | `900000`|

packages/taquito-rpc/src/types.ts
```diff
@@ -596,24 +598,26 @@ export interface ConstantsResponse {
   tokens_per_roll: BigNumber;
   michelson_maximum_type_size: number;
   seed_nonce_revelation_tip: string;
-  origination_burn: string;
+  origination_size: number;
   block_security_deposit: BigNumber;
   endorsement_security_deposit: BigNumber;
-  block_reward: BigNumber;
```
jevonearth, line 602: Removing this field will be problematic for protos < 7.

jevonearth (changes requested):

packages/taquito-rpc/src/types.ts
roxaneletourneau: #449 Created this issue to implement these tests.

packages/taquito-rpc/src/types.ts
roxaneletourneau: This will be resolved by adding union type (in the following commit)

jevonearth (approved):
Approved, just waiting for integration tests to pass, then we can merge to master.

jevonearth (approved):

PR #452, opened 2020-10-22
Removed the Tezos function

roxaneletourneau:
For simplicity, the Tezos function added in the TezosToolkit (to replace the Tezos singleton after removing the default RPC URL) has been removed. So the only way to create a new instance will be :
`const tezos = new TezosToolkit('your_rpc');`

Fixed examples that were using this function.


jevonearth:
@claudebarde do you think it would be worthwhile to go through your tutorial blogs and update them so that they conform with these changes?

jevonearth (approved):

PR #454, opened 2020-10-26
Migration doc v7

roxaneletourneau:
Added a new documentation page to expose each breaking change in version 7 with:
- the reasons that motivated it,
- code examples that show what needs to be changed, and
- potential errors when updating to v7 without making the fixes.

### DIFF
--- a/docs/v7_breaking_changes.md
+++ b/docs/v7_breaking_changes.md
@@ -3,4 +3,170 @@ title: Taquito version 7
 author: Roxane Letourneau
 ---
 
-With this major number update, we decided to refract some parts of the Taquito code. The migration to version 7 introduces breaking changes.
+:::warning Breaking changes
+With this major number update, we decided to refract some parts of the Taquito code. Please note that the migration to version 7 introduces breaking changes. 
+:::
+
+Each breaking change is exposed on this page with:
+- the reasons that motivated it,
+- code examples that show what needs to be changed on your code (how it was in prior versions versus how it needs to be using v7), and
+- potential errors you might get when updating to v7 without making the fixes.
+
+## Removed the default RPC Node URL
+
+Before version 7, Taquito was configured to use our default RPC node; the default value was set in the constructor of the `RpcClient` class. We took this approach so that users can get started quickly, and Taquito should "just work" with minimal fiddling. Users could import a ready-to-use `Tezos` singleton, an instance of the `TezosToolkit` class using the default RPC URL.    
+
+However, in version 7 of Taquito, we decided to remove the default RPC node URL. The reason behind this choice is to avoid that a lot of applications rely on our default node and thus centralize a part of the ecosystem on one node.
+
+This change impacts the following classes, where it is now required to specify an RPC node in their constructor:
+- `RpcClient`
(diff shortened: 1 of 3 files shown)


docs/v7_breaking_changes.md
```diff
@@ -0,0 +1,172 @@
...
+```
+
+### Error when not replacing main with default:
+
+If you update to version 7 without replacing main with default in your code, you might get the following error at runtime:  
+`contract.methods.main is not a function`
+
+## Removed the importKey method from TezosToolkit class
+
+This method has been marked as deprecated in march 2020 in favor of setting the signer provider with `@taquito/signer importKey`.
```
jevonearth, line 85: "march" --> "March"

docs/v7_breaking_changes.md
```diff
@@ -0,0 +1,172 @@
+---
+title: Taquito version 7
+author: Roxane Letourneau
+---
+
+:::warning Breaking changes
+With this major number update, we decided to refract some parts of the Taquito code. Please note that the migration to version 7 introduces breaking changes.
```
jevonearth, line 7: "we decided to refract some parts of the Taquito code." --> "we decided to refactor or change some parts of the Taquito API."

docs/v7_breaking_changes.md
```diff
@@ -0,0 +1,172 @@
+---
+title: Taquito version 7
+author: Roxane Letourneau
+---
+
+:::warning Breaking changes
+With this major number update, we decided to refract some parts of the Taquito code. Please note that the migration to version 7 introduces breaking changes. 
+:::
+
+Each breaking change is exposed on this page with:
```
jevonearth, line 10: "exposed" --> "explained"

docs/v7_breaking_changes.md
```diff
@@ -0,0 +1,172 @@
...
+- the reasons that motivated it,
+- code examples that show what needs to be changed on your code (how it was in prior versions versus how it needs to be using v7), and
+- potential errors you might get when updating to v7 without making the fixes.
+
+## Removed the default RPC Node URL
+
+Before version 7, Taquito was configured to use our default RPC node; the default value was set in the constructor of the `RpcClient` class. We took this approach so that users can get started quickly, and Taquito should "just work" with minimal fiddling. Users could import a ready-to-use `Tezos` singleton, an instance of the `TezosToolkit` class using the default RPC URL.    
+
+However, in version 7 of Taquito, we decided to remove the default RPC node URL. The reason behind this choice is to avoid that a lot of applications rely on our default node and thus centralize a part of the ecosystem on one node.
```
jevonearth, line 19: "The reason behind this choice is to avoid that a lot of applications rely on our default node and thus centralize a part of the ecosystem on one node."

change to:

"The reason behind this choice is to encourage developers to make their own informed choice on which Tezos RPC node (public or private) is best for them. This change also helps avoid dApps using Taquito from centralizing on one public RPC node. Decentralization is an important part of Tezos, and we want to encourage decentralization at the RPC infrastructure level."

jevonearth (changes requested):

PR #465, opened 2020-10-29
Fixed live code examples in the wallet documentation

roxaneletourneau:
Added tezbridge plugin to docusaurus config to fix the Tezbridge wallet example.
Added the import for ThanosWallet.
Made the other examples use the taquito-beacon-wallet package otherwise it was using an ephemeral key to sign.


jevonearth (approved):
:100:

Issue #432, opened 2020-09-10
SPIKE: Design better API for dryrun, estimates, signing and preparation


jevonearth:
The `transfer()` method abstracts many underlying steps for the developer user, which is a good thing for typical usescases. When a developer user wishes to retrieve data from some particular underlying step, they use the lower-level abstractions in taquito.

If a user wishes to get the estimate for a transaction operation on a contract method, they need to prepare the call, fetch the parameters, and then call the `Tezos.estimate.transfer({op...})` API.

A proposal for a new set of methods follow;

(_Controversial, maybe a better way around this?_) Break the api so that `contract.transfer()` becomes a no-op method.

Idea is to add the following new methods;

```
contract.transfer({ to: address, amount: amount }).send()
contract.transfer({ to: address, amount: amount }).estimate()
contract.transfer({ to: address, amount: amount }).dryrun()
contract.transfer({ to: address, amount: amount }).sign()
contract.transfer({ to: address, amount: amount }).prepare()
```

And for smart contract invocations, it may look like this:

```
contract.methods.increment(1).send()
contract.methods.increment(1).estimate()
contract.methods.increment(1).dryrun()
contract.methods.increment(1).sign()
contract.methods.increment(1).prepare() 
```

`send` will complete all steps and inject the method
`estimate` will return an Estimate for the operation
`dryrun` will return the dryrun results, including the updated storage as evaluated by the RPC server
`sign` will return the signed operation but not inject the operation
`prepare` will prepare an operation that is ready for singing.

**Acceptance criteria:**
1. Prepare and agree on design
2. Create follow-up issues to prototype(if needed) and implement

roxaneletourneau:
In addition to the above; Provide the ability to obtain operation hash before sending it to the node

Innkst:
related to previous implementation #995

Innkst:
We need to review the design and have a discussion again. Also, please re-confirm the follow-up tasks and let's what non-breaking changes could be implemented earlier and which ones have to wait till v15

roxaneletourneau:
Created follow-up task issues 2019 and 2020

PR #512, opened 2020-11-22
Make http requests cancelable

serh11p:
### Why?

Taquito has very heavy methods under the hood which make a lot of requests, such as Estimate or ReadViewLambda.
When using it in interfaces, there is often a case when the method needs to be canceled almost immediately.
So, the goal is the network optimization.

My vision of this story is that there is an [RPCOptions](https://github.com/ecadlabs/taquito/blob/master/packages/taquito-rpc/src/taquito-rpc.ts#L47) argument with the `signal` property in all of Taquito's methods that use RPC queries under the hood.

Made for https://developer.mozilla.org/en-US/docs/Web/API/AbortController.

### Example

```ts
const abortCtrl = new AbortController();

const reqPromise = httpBackend.createRequest({ url: "https://...", signal: abortCtrl.signal });

reqPromise
  .then(console.info)
  .catch(console.error);

setTimeout(() => {
  abortCtrl.abort();
}, 100);
```

### Use AbortController in different environments
There are three main environments where your JavaScript code can run in:
- Modern Browsers: You can use AbortController without problems because AbortController was created to work with the DOM by the WHATWG.
- Node: If you want to use AbortController in Node, you can do it using an implementation of the WHATWG AbortController interface. You can use different npm packages to achieve this, for example, [abort-controller](https://www.npmjs.com/package/abort-controller) or [node-fetch](https://www.npmjs.com/package/node-fetch).
- Legacy Browsers: You need to use polyfills, for example, with Babel.

### TODO

- [x] All public methods or types have TypeDoc coverage with a complete description, and ideally an @example


jevonearth:
Thanks for the @serh11p 
We will schedule how to integrate this after we ship tzip16 and beacon v2 work. :) Really appreciate the PR illustrating the way.

roxaneletourneau:
This implementation would require changes in the high-level API. Here is an example where a custom HttpBackend class is used to be able to cancel all requests: https://github.com/ecadlabs/taquito/pull/914. 
I'm closing this for now; an implementation to cancel specific requests will be considered when designing the issue [432](https://github.com/ecadlabs/taquito/issues/432), which might require providing a new API.

PR #513, opened 2020-11-23
Allow numbers as big map keys

claudebarde:
Taquito only allows strings to be used as keys when searching a value in a big map, which can create frictions when the big map keys are numbers. This update allows string and numbers to be used to search in big maps.


roxaneletourneau (changes requested):
This will require more investigation as there are integration tests that are failing.
For example: contract-pair-as-key.spec.ts

PR #617, opened 2021-02-15
Item-580 fix for max_priority parameter is ignored in getBakingRights…

kjaiswal:
… when it's value is zero.

## Release Note Draft Snippet

### DIFF
--- a/packages/taquito-http-utils/src/taquito-http-utils.ts
+++ b/packages/taquito-http-utils/src/taquito-http-utils.ts
@@ -50,7 +50,7 @@ export class HttpBackend {
 
     const str = [];
     for (const p in obj) {
-      if (obj.hasOwnProperty(p) && obj[p]) {
+      if (obj.hasOwnProperty(p) && typeof obj[p] !== 'undefined') {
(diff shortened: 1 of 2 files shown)


packages/taquito-http-utils/src/taquito-http-utils.ts
```diff
@@ -43,14 +43,14 @@ export class HttpRequestFailed implements Error {
 }
 
 export class HttpBackend {
-  private serialize(obj?: { [key: string]: any }) {
```
roxaneletourneau, line 46: Great work on this PR.

Just one question/comment: 
Would it have been possible to keep this method private?
I understand that you made it public to test it and that testing it via the `createRequest` method would not have been ideal. A hacky solution when I want to test a private method is to call it that way: `httpBackend['serialize'](httpRequestQuery)`
Let me know what you think.

kjaiswal: Thank you, I have made changes according to feedback.

roxaneletourneau (approved):

Issue #629, opened 2021-02-18
Add logging to console for taquito integration tests when call to key agent returns 503


michaelkernaghan:
When running Taquito integration tests, either locally or in the CI, a critical call to the keygen agent is made. When the call fails, the tests will fail, but it is not obvious why. 
If we add a log to console to the code where the key agent is called, if the response is 503, we can see immediately that the cause of the test failures is that keygen needs attention.

here is the relevant call to keygen agent in integration-tests/config.ts:

  const key = await httpClient.createRequest<string>({
    url: keyUrl,
    method: 'POST',
    headers: requestHeaders,
    json: false,
  });

roxaneletourneau:
@michaelkernaghan  
I tried something: wrap the call in a try-catch and logging to the console any error that happens (not specific to 503). It will be in my next PR with the fix for the estimation test.

PR #630, opened 2021-02-18
fix integration tests

roxaneletourneau:
- Fixed the estimation test (2 mutez difference).
- Replaced the public gateway for IPFS from 'gateway.ipfs.io' to 'cloudflare-ipfs.com', to see if the requests will timeout less often.
- log error from the keygen in the console: #629


jevonearth (approved):
Thank you!

PR #651, opened 2021-02-24
Try npm jest retries package to retry integration tests that fail due to connection problems

michaelkernaghan:
To try to reduce the failures in integration tests that are due to transients we can retry failed tests to get a better connection on a retry.

This is a sample to see how well this works, using a test that fails frequently.


roxaneletourneau (approved):
It might also be interesting to add a log when a test has retried.

PR #716, opened 2021-03-18
Dynamically set polling interval from constants file base…

kjaiswal:
…d on param

## Release Note Draft Snippet

### DIFF
--- a/packages/taquito/src/context.ts
+++ b/packages/taquito/src/context.ts
@@ -17,10 +17,15 @@ import { ParserProvider } from './parser/interface';
 import { MichelCodecParser } from './parser/michel-codec-parser';
 import { Packer } from './packer/interface';
 import { RpcPacker } from './packer/rpc-packer';
+import BigNumber from 'bignumber.js';
 
 export interface TaquitoProvider<T, K extends Array<any>> {
   new (context: Context, ...rest: K): T;
 }
+
+export const defaultConfirmationCountConst = 1;
+export const confirmationPollingTimeoutSecondConst = 180;
+export const shouldObservableSubscriptionRetryConst = false;
 
 // The shouldObservableSubscriptionRetrythe parameter is related to the observable in ObservableSubsription class. 
 // When set to true, the observable won't die when getBlock rpc call fails; the error will be reported via the error callback, 
@@ -32,11 +37,10 @@ export interface Config {
   shouldObservableSubscriptionRetry?: boolean;
 }
 
-export const defaultConfig: Required<Config> = {
-  confirmationPollingIntervalSecond: 10,
-  defaultConfirmationCount: 1,
-  confirmationPollingTimeoutSecond: 180,
-  shouldObservableSubscriptionRetry: false
+export const defaultConfig: Partial<Config> = {
+  defaultConfirmationCount: defaultConfirmationCountConst,
+  confirmationPollingTimeoutSecond: confirmationPollingTimeoutSecondConst,
+  shouldObservableSubscriptionRetry: shouldObservableSubscriptionRetryConst
 };
 
 /**
@@ -82,11 +86,11 @@ export class Context {
     this._packer = packer? packer: new RpcPacker(this);
   }
 
-  get config(): Required<Config> {
+  get config(): Partial<Config> {
     return this._config as any;
   }
 
-  set config(value: Required<Config>) {
+  set config(value: Partial<Config>) {
     this._config = {
       ...defaultConfig,
       ...value,
@@ -166,6 +170,28 @@ export class Context {
     }
   }
 
+  async getConfirmationPollingInterval() {
+    try {
+      const constants = await this.rpc.getConstants();
+      let confirmationPollingInterval = BigNumber.sum(constants.time_between_blocks[0], 
+             constants.delay_per_missing_endorsement!, 
+             Math.max(0, constants.initial_endorsers! - constants.endorsers_per_block));
+      
+      // Divide the polling interval by a constant 3
+      // to improvise for polling time to work in prod,
+      // testnet and sandbox enviornment.   
+      confirmationPollingInterval = confirmationPollingInterval.dividedBy(3);  
+      this.config.confirmationPollingIntervalSecond = confirmationPollingInterval.toNumber();
+      return this.config.confirmationPollingIntervalSecond;
+    } catch (exception) {
+      console.log(exception);
+    }
+    // Return default value if there is
+    // an issue returning from constants
+    // file.
+    return 10;
+  }
+  
   /**
    * @description Create a copy of the current context. Useful when you have long running operation and you do not want a context change to affect the operation
    */
--- a/packages/taquito/src/context.ts
+++ b/packages/taquito/src/context.ts
@@ -18,6 +18,8 @@ import { MichelCodecParser } from './parser/michel-codec-parser';
 import { Packer } from './packer/interface';
 import { RpcPacker } from './packer/rpc-packer';
 import BigNumber from 'bignumber.js';
+import { throwError } from 'rxjs';
+import { Console } from 'console';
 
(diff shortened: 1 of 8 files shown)


packages/taquito/src/context.ts
```diff
@@ -166,6 +166,27 @@ export class Context {
...
+    try {
+      const constants = await this.rpc.getConstants();
+      let confirmationPollingInterval = BigNumber.sum(constants.time_between_blocks[0], 
+             constants.delay_per_missing_endorsement!, 
+             Math.max(0, constants.initial_endorsers! - constants.endorsers_per_block));
+      
+      // Divide the polling interval by a constant 3
+      // to improvise for polling time to work in prod,
+      // testnet and sandbox enviornment.   
+      confirmationPollingInterval = confirmationPollingInterval.dividedBy(3);
```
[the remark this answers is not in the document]
kjaiswal: This has been fixed.

packages/taquito/src/operations/operations.ts
```diff
@@ -168,11 +172,11 @@ export class Operation {
         .pipe(
           switchMap(() => this.polling$),
           switchMap(() => this.currentHead$),
-          filter(head => head.header.level - this._foundAt >= conf - 1),
+          filter(head => head.header.level - this._foundAt >= conf! - 1),
```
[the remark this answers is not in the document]
roxaneletourneau: I think it is because of the change from Required<Config> to Partial<Config> in the context class. Maybe using Partial there is not the right thing to do, could we use something else?

packages/taquito/src/wallet/operation-factory.ts
```diff
@@ -55,7 +55,7 @@ export const createNewPollingBasedHeadObservable = (
     switchMap(() => sharedHeadOb),
     distinctUntilKeyChanged('hash'),
     timeoutWith(
-      context.config.confirmationPollingTimeoutSecond * 1000,
+      context.config.confirmationPollingTimeoutSecond! * 1000,
```
roxaneletourneau, line 58: Same as the above discussion

packages/taquito/src/wallet/operation.ts
```diff
@@ -153,7 +153,7 @@ export class WalletOperation {
           block: head,
           expectedConfirmation: conf,
           currentConfirmation: head.header.level - foundAtBlock.header.level + 1,
-          completed: head.header.level - foundAtBlock.header.level >= conf - 1,
+          completed: head.header.level - foundAtBlock.header.level >= conf! - 1,
```
roxaneletourneau, line 156: same as the above discussion

packages/taquito/src/operations/operations.ts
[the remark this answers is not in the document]
kjaiswal: I have introduced "throw errors" in this case as suggested.

packages/taquito/src/wallet/operation-factory.ts
kjaiswal: Introduced "throw errors" as suggested.

packages/taquito/src/wallet/operation.ts
kjaiswal: Introduced "throw errors" as suggested.

roxaneletourneau:
The tests timeout because confirmationPollingIntervalSecond is undefined at this line of code:
https://github.com/ecadlabs/taquito/blob/master/packages/taquito/src/operations/operations.ts#L160

We need to call the new getConfirmationPollingInterval() method if the property is undefined:
```
const confirmationPollingIntervalSecond = this.context.config.confirmationPollingIntervalSecond !== undefined 
                                        ? this.context.config.confirmationPollingIntervalSecond 
                                        : await this.context.getConfirmationPollingInterval();
```

kjaiswal:
> The tests timeout because confirmationPollingIntervalSecond is undefined at this line of code:
> https://github.com/ecadlabs/taquito/blob/master/packages/taquito/src/operations/operations.ts#L160
> 
> We need to call the new getConfirmationPollingInterval() method if the property is undefined:
> 
> ```
> const confirmationPollingIntervalSecond = this.context.config.confirmationPollingIntervalSecond !== undefined 
>                                         ? this.context.config.confirmationPollingIntervalSecond 
>                                         : await this.context.getConfirmationPollingInterval();
> ```

Surprising this didn't show up when i run tests locally. Thank you I have re-introduced confirmationPollingIntervalSecond now which was removed while fixing ! expressions.

roxaneletourneau (approved):

packages/taquito/src/context.ts
```diff
@@ -166,6 +166,28 @@ export class Context {
     }
   }
 
+  async getConfirmationPollingInterval() {
+    try {
+      const constants = await this.rpc.getConstants();
+      let confirmationPollingInterval = BigNumber.sum(constants.time_between_blocks[0], 
+             constants.delay_per_missing_endorsement!, 
+             Math.max(0, constants.initial_endorsers! - constants.endorsers_per_block));
```
roxaneletourneau, line 174: One last thing, while debugging earlier, I find out that the value for confirmationPollingIntervalSecond is 11.333333333333334 instead of 10.
The calculus is different from what you shared on slack some time ago. 
Here it is : 30 + 4 + Max(0,-8)) = 34
On slack : 30 + (4*Max(0,-8)) = 30
Which one is correct?

kjaiswal: Yes you are right it needs to be multiply. I will fix it [here.[](url)](url
Screen Shot 2021-03-12 at 10 09 27 AM
)

kjaiswal: This is fixed now.

packages/taquito/src/context.ts
```diff
@@ -17,6 +17,9 @@ import { ParserProvider } from './parser/interface';
 import { MichelCodecParser } from './parser/michel-codec-parser';
 import { Packer } from './packer/interface';
 import { RpcPacker } from './packer/rpc-packer';
+import BigNumber from 'bignumber.js';
+import { throwError } from 'rxjs';
+import { Console } from 'console';
```
roxaneletourneau, line 22: imports used:
import { throwError } from 'rxjs';
import { Console } from 'console';

packages/taquito/src/context.ts
```diff
@@ -166,6 +168,29 @@ export class Context {
...
+      );
+      
+      // Divide the polling interval by a constant 3
+      // to improvise for polling time to work in prod,
+      // testnet and sandbox enviornment.   
+      confirmationPollingInterval = confirmationPollingInterval.dividedBy(3);  
+      this.config.confirmationPollingIntervalSecond = confirmationPollingInterval.toNumber();
+      return this.config.confirmationPollingIntervalSecond;
+    } catch (exception) {
+      console.log(exception);
```
roxaneletourneau, line 186: Should we remove that? we don't want dapp builders seeing unexpected logging in their app

packages/taquito/src/context.ts
kjaiswal: Done

packages/taquito/src/context.ts
kjaiswal: This is fixed.

packages/taquito/src/context.ts
```diff
@@ -17,6 +17,8 @@ import { ParserProvider } from './parser/interface';
 import { MichelCodecParser } from './parser/michel-codec-parser';
 import { Packer } from './packer/interface';
 import { RpcPacker } from './packer/rpc-packer';
+import BigNumber from 'bignumber.js';
+import { throwError } from 'rxjs';
```
roxaneletourneau, line 21: unsued import { throwError } from 'rxjs';

michaelkernaghan (commented):
lines 13 and 60 have a typo :   enviornment  should be environment

PR #721, opened 2021-03-20
Contract type generator

ricklove:
Edit: Replaced by #722 

---

## Release Note Draft Snippet

Generate Types from Contracts


ricklove:
This PR is replaced by #722 which also includes generic type extensions to TezosToolkit

packages/taquito-contract-type-generator/src/generator/contract-parser.ts
```diff
@@ -0,0 +1,339 @@
...
+                isBigMap: node.prim === `big_map`,
+            },
+        };
+    }
+
+    // option
+    if (node.prim === `option`) {
+        return {
+            ...visitType(node.args[0]),
+            optional: true,
```
roxaneletourneau, line 244: In Michelson we can have an optional parameter before a required one, but for these cases, when generating the type, typescript shows the following error: 
```
 set_child_record: (
        address?: address,
        data: MMap<string, ( ... // A required parameter cannot follow an optional parameter.
```
Could we return the type or undefined instead of making the parameter optional?
```
 set_child_record: (
        address: address | undefined,
        data: MMap<string, ( ...
```
When I call this method without an address, I explicitly need to write undefined (or null) for this parameter:
```
await c.methods.set_child_record(
          undefined, // address
          new MichelsonMap(), // data
```
so while getting rid of the typescript error, this change could be more intuitive in showing how to write the optional parameter.

packages/taquito-contract-type-generator/example/example-usage.ts
```diff
@@ -0,0 +1,207 @@
...
+const exampleContractMethods1 = async () => {
+
+    // TODO: Support Generic Type in TezosToolkit Directly
+    // const Tezos = new TezosToolkit<TestContractType>(`https://YOUR_PREFERRED_RPC_URL`)
+    const Tezos = new TezosToolkit(`https://YOUR_PREFERRED_RPC_URL`) as unknown as TezosToolkitTyped<TestContractType>;
+
+    const contract = await Tezos.contract.at(`tz123`);
+
+    contract.methods.bid(tas.nat(0));
+    contract.methods.configure({
```
roxaneletourneau, line 17: I think this example is not right as the contract methods do not expect an object but flat parameters.  We have an internal discussion about accepting an object for the methods (like we do for the storage) as it is easier to reason about, but this would be a breaking change...
But that said, the work you did in this package makes it easier to discover what parameters are expected, this is really helpful.

packages/taquito-contract-type-generator/src/generator/contract-parser.ts
ricklove: @roxaneletourneau good catch, I initially had those inside an object and didn't find that when changing it.

ricklove: @roxaneletourneau though I would argue that taquito would work better if the method calls were wrapped in an object (though not backward compatible, but maybe something to consider for future design)

packages/taquito-contract-type-generator/example/example-usage.ts
ricklove: Right, this is wrong - leftover from how I first did it.

Thanks!

jevonearth:
Superseded by upcoming Taqueria plugin.

PR #728, opened 2021-03-23
Added Teznode to list of community-run nodes

amani1104:
The architecture consists of a nginx load-balancer and two Tezos bare-metal nodes. We are using a whitelist approach to access the node so that dangerous calls are prevented. If a rpc required by developers is currently missing, we are very approachable and happy to add the respective rpc to the whitelist. 

https://letzbake.com/
https://t.me/cryptodad77


jevonearth (approved):

jevonearth:
Thanks @amani1104 !

PR #735, opened 2021-03-23
Issue-601 New RPC endpoints

kjaiswal:

## Release Note Draft Snippet

### DIFF
--- a/packages/taquito-rpc/src/taquito-rpc.ts
+++ b/packages/taquito-rpc/src/taquito-rpc.ts
@@ -41,6 +41,7 @@ import {
   ScriptResponse,
   StorageResponse,
   VotesListingsResponse,
+  VotingPeriodBlockResult,
 } from './types';
 import { castToBigNumber } from './utils/utils';
 
@@ -510,6 +511,8 @@ export class RpcClient {
    * @param options contains generic configuration for rpc calls
    *
    * @description Current period kind.
+   * 
+   * @deprecated Deprecated in favor of getCurrentPeriod
    *
    * @see https://tezos.gitlab.io/api/rpc.html#get-block-id-votes-current-period-kind
    */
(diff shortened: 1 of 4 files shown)


packages/taquito-rpc/src/taquito-rpc.ts
```diff
@@ -779,4 +780,41 @@ export class RpcClient {
     return this.url
   }
 
+  /**
+   *
+   * @param options contains generic configuration for rpc calls
+   *
+   * @description Voting period of current block.
+   *
+   * @see https://tezos.gitlab.io/008/rpc.html#get-block-id-votes-current-period
```
jevonearth, line 789: @kjaiswal 

Can you update the URLs to use `api` in place of the proto number? ie `https://tezos.gitlab.io/api/rpc.html#get-block-id-votes-current-period` 

The API website should always redirect the API to the latest protocol, so it should mean less maintenance for us.

jevonearth: Can you add `@example` props too?

Adding an `@example` prop for the typical case (head) and for a specific block is very helpful to developers.

jevonearth (changes requested):
Looks great, two minor requests Re Typedocs

packages/taquito-rpc/src/types.ts
```diff
@@ -677,7 +687,9 @@ export interface BlockMetadata {
   max_operation_list_length: MaxOperationListLength[];
   baker: string;
   level: Level;
+  level_info: LevelInfo;
```
roxaneletourneau, line 690: I think level, level_info, voting_period_kind, and voting_period_info should be optional to stay compatible.


Deprecated: metadata.level, metadata.voting_period_kind
New properties: metadata.level_info, metadata.voting_period_info

packages/taquito-rpc/src/taquito-rpc.ts
kjaiswal: Thank you for the feedback I have added @example props and replaced 008 with api.

packages/taquito-rpc/src/types.ts
kjaiswal: Thank you for the feedback, I have made these properties optional.

jevonearth (approved):

PR #739, opened 2021-03-24
mk-738-add-retry-to-wallet-api-integration-test

### LINKED ISSUE #738: add retry to intermittent failing integrations  tests wallet-api.spec.ts
**Description**
Sometimes the integration. tests will fail in CI for the tests in wallet-api.spec.ts
These failures can be reduced with the addition of a retry command to the test.

michaelkernaghan:
mk-738-add-retry-to-wallet-api-integration-test

- [x ] Your code builds cleanly without any errors or warnings

Close #738


michaelkernaghan:
I think actually retries are the wrong fix for these test. Better might be to split the sequences of send calls into distinct tests

PR #743, opened 2021-03-24
chore: remove delphi, carthage and babylon integration tests

jevonearth:
Replaces #737


integration-tests/batch-api.spec.ts
```diff
@@ -29,7 +29,7 @@ CONFIGS().forEach(({ lib, rpc, setup, knownBaker, createAddress }) => {
     })
 
     it('Simple transfers with origination using with', async (done) => {
-      const op = await Tezos.batch([
+      const op = await Tezos.contract.batch([
```
roxaneletourneau, line 32: Should we keep this test as it until we removed the deprecated batch function?

contract.batch() is covered in the file contract-batch.spec.ts

integration-tests/contract-originate-contract-with-sapling.spec.ts
```diff
@@ -4,17 +4,16 @@ import { Protocols } from "@taquito/taquito";
...
-  
   describe(`Test origination of contracts with sapling using: ${rpc}`, () => {
 
     beforeEach(async (done) => {
       await setup();
       done()
     })
-
-     edonet('Originates a contract having an empty sapling state in its storage', 2, async (done: () => void) => {
+    //  deepcode ignore only-arrow-functions: not sure how to do it otherwise
```
roxaneletourneau, line 15: Do you think this line can be removed if the following one is written this way: 
edonet('Originates a contract having an empty sapling state in its storage', 2, async (done: () => void) => {

integration-tests/edo-deploy-contract-having-ticket.spec.ts
```diff
@@ -4,17 +4,16 @@ import { Protocols } from "@taquito/taquito";
...
 
   describe(`Test origination of a token contract using: ${rpc}`, () => {
 
     beforeEach(async (done) => {
       await setup();
       done()
     })
-
-    edonet('Originates a contract having ticket with init and the wallet api', 2, async (done: () => void) => {
+    //  deepcode ignore only-arrow-functions: not sure how to do it otherwise
```
roxaneletourneau, line 15: Same as above

integration-tests/contract-originate-contract-with-sapling.spec.ts
michaelkernaghan: i will change it

michaelkernaghan: made the change

integration-tests/edo-deploy-contract-having-ticket.spec.ts
michaelkernaghan: i will try it

PR #748, opened 2021-03-25
Item-724 Subscription retries limited to 10 with 1 second time interval

kjaiswal:

## Release Note Draft Snippet

### DIFF
--- a/packages/taquito/src/subscribe/observable-subscription.ts
+++ b/packages/taquito/src/subscribe/observable-subscription.ts
@@ -1,6 +1,6 @@
-import { Observable, Subscription as RXJSSubscription, Subject, of, NEVER } from 'rxjs';
+import { Observable, Subscription as RXJSSubscription, Subject, timer, throwError } from 'rxjs';
 import { Subscription } from './interface';
-import { takeUntil, tap, catchError, retryWhen, delay, take } from 'rxjs/operators';
+import { takeUntil, tap, catchError, retryWhen, take, mergeMap } from 'rxjs/operators';
 
 export class ObservableSubscription<T> implements Subscription<T> {
   private errorListeners: Array<(error: Error) => void> = [];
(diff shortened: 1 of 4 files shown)


packages/taquito/src/subscribe/observable-subscription.ts
```diff
@@ -23,7 +24,7 @@ export class ObservableSubscription<T> implements Subscription<T> {
             this.call(this.closeListeners);
           }
         ),
-        this.shouldRetry ? retry() : tap(),
+        this.shouldRetry ? retryWhen(errors => errors.pipe(delay(1000), take(10))) : tap(),
```
roxaneletourneau, line 27: Should the retry strategy configurable by the user instead of hard-coding 10 retries with a 1-second delay?

kjaiswal: I have made these user configurable.

packages/taquito/test/subscribe/observable-subscription.spec.ts
```diff
@@ -32,7 +31,7 @@ describe('Observable subscription test', () => {
         flush();
 
         expect(stub).toBeCalledTimes(3);
-        expect(errStub).toBeCalledTimes(2);
```
roxaneletourneau, line 35: From my understanding, the stub is called 3 times (a,b,c), and then an error happens (#). 
It seems that nothing happens after the error (d-e are never called.. I would have expected the stub to have been called 5 times with this change). This test was to make sure that the observable retries on error when the property "shouldRetry" is set to true, but it seems that the retry is not tested.

kjaiswal: This is fixed.

roxaneletourneau:
The `stream` property of the `TezosToolikit` is using the `ObservableSubscription` class. There is an example [here](https://github.com/ecadlabs/taquito/blob/master/example/example-streamer.ts) where it shows how to configure it to retry on an error with the `shouldObservableSubscriptionRetry` parameter. 
With the current changes, is it possible to configure the number of retries and the delay at this level? Can the example be modified to reflect it?

PR #759, opened 2021-03-30
Allows to fetch bigmap with a key of type string, number or object

roxaneletourneau:
Replace #513 

Breaking change for `BigMapAbstraction.getMultipleValues()` and `RpcContractProvider.getBigMapKeysByID()` that will return a `MichelsonMap` instead of an object to support keys of type object (michelson pair)

## Release Note Draft Snippet

### DIFF
--- a/packages/taquito/src/contract/rpc-contract-provider.ts
+++ b/packages/taquito/src/contract/rpc-contract-provider.ts
@@ -1,5 +1,5 @@
 import { HttpResponseError, STATUS_CODE } from '@taquito/http-utils';
-import { Schema } from '@taquito/michelson-encoder';
+import { BigMapKeyType, MichelsonMap, MichelsonMapKey, Schema } from '@taquito/michelson-encoder';
 import { ScriptResponse } from '@taquito/rpc';
 import { encodeExpr } from '@taquito/utils';
 import { OperationBatch } from '../batch/rpc-batch-provider';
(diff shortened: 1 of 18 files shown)


packages/taquito/src/contract/rpc-contract-provider.ts
```diff
@@ -124,61 +124,46 @@ export class RpcContractProvider extends OperationEmitter
...
-          throw ex;
-        }
-      }
-      Object.assign(bigMapValues, {[keys[0]]: val});
+  async getBigMapKeysByID<T>(id: string, keys: Array<BigMapKeyType>, schema: Schema, block?: number): Promise<MichelsonMap<MichelsonMapKey, T | undefined>> {
+    const level = await this.getBlockForRequest(keys, block)
+    const bigMapValues = new MichelsonMap<MichelsonMapKey, T | undefined>();
+
+    // Execute batch of promises in series
+    const batchSize = 5;
```
carte7000, line 135: Should that be configurable?

packages/taquito/src/contract/rpc-contract-provider.ts
```diff
@@ -124,61 +124,46 @@ export class RpcContractProvider extends OperationEmitter
...
+    let position = 0;
+    let results: Array<(T | undefined)> = [];
+
+    while (position < keys.length) {
+      const keysBatch = keys.slice(position, position + batchSize);
+      const batch = keysBatch.map((keyToEncode) => this.getBigMapValueOrUndefined<T>(keyToEncode, id, schema, level))
```
carte7000, line 141: Do we have a test for the negative case? If the key is not found we return undefined. What happens when there is an unexpected exception? Should we test that these exceptions bubble up to the caller?

packages/taquito/src/contract/rpc-contract-provider.ts
roxaneletourneau: Good idea, I made the change.

packages/taquito/src/contract/rpc-contract-provider.ts
roxaneletourneau: I added a test case, thank you!

PR #790, opened 2021-04-15
Issue-724 Added exponential slow down of requests

kjaiswal:

## Release Note Draft Snippet


roxaneletourneau:
Would it give better flexibility to the user if we let them pass an operator of their choice for the retry strategy and we keep the previous infinite retry() as the default value? Let me know what you think.

kjaiswal:
> Would it give better flexibility to the user if we let them pass an operator of their choice for the retry strategy and we keep the previous infinite retry() as the default value? Let me know what you think.

Yes I am still undecided on number of retries part, I am okay with keeping infinite retry(), can you please provide an example to explain the operator and how it would be used in retry strategy.

roxaneletourneau:
> Yes I am still undecided on number of retries part, I am okay with keeping infinite retry(), can you please provide an example to explain the operator and how it would be used in retry strategy.

Instead of having `private retries = 10, private delayInMS = 1000` in the constructor of `ObservableSubscription`, it could accept an [OperatorFunction](https://rxjs-dev.firebaseapp.com/api/index/interface/OperatorFunction). This operator could be set to `retry()` by default. Users will be able to pass an operator that suits their needs if they don't want the infinite retry.

packages/taquito/src/context.ts
```diff
@@ -32,15 +33,17 @@ export interface Config {
...
 }
 
 export const defaultConfig: Partial<Config> = {
   defaultConfirmationCount: 1,
   confirmationPollingTimeoutSecond: 180,
   shouldObservableSubscriptionRetry: false,
-  observableSubscriptionRetries: 10,
-  observableSubscriptionRetryDelay: 1000
+  observableSubscriptionRetryFunction: retry,
+  observableSubscriptionRetryDelay: 1000, // 1 minute
```
itkach, line 45: comment says "1 minute", but it looks like it's actually 1 second? Perhaps it's better not to have a comment here at all, and either make units obvious from the name (like confirmationPollingTimeoutSecond does) or stick to the library conventions if applicable (maybe all timeouts/delays are consistently specified in the library in milliseconds ?)

kjaiswal: Fixed comment, thank you.

packages/taquito/src/subscribe/observable-subscription.ts
```diff
@@ -24,8 +29,13 @@ export class ObservableSubscription<T> implements Subscription<T> {
...
           }
         ),
-        this.shouldRetry ? retryWhen(errors => errors.pipe(take(retries))) : tap(),
-        catchError(() => timer(delayInMS))
+        this.shouldRetry ? operatorFunction() : tap(),
+        catchError(() => {
+          delayInMS = errorFlag ? delayInMS*2 : delayInMS;
+          delayInMS = delayInMS < maxDelayInMS ? delayInMS : maxDelayInMS;
+          errorFlag = true;
+          return timer(delayInMS);
```
roxaneletourneau, line 37: By accepting an operator as a param, I was thinking that we should let the user decide his own retry logic, I don't understand why we need to include this logic in the catch block. Can you explain, please?

Issue #448, opened 2020-10-20
Add reveal and activateAndReveal method to the RpcTzProvider class


roxaneletourneau:
At the moment, there is a check to see if the account is revealed when sending an operation of type `transaction`, `delegation`, or `origination`. If the account is not revealed, the reveal is done at the same time as sending the operation. 
Adding `reveal` and `activateAndReveal` methods to the RpcTzProvider class would make it possible to reveal an account without having to send another transaction.

PR #794, opened 2021-04-16
Include reveal fees in estimate

### LINKED ISSUE #772: Estimate for reveal operation are not returned on batch estimation
**Description**
The implicit reveal data are not returned when doing estimation on batched operations.
For example, during a batch with two transfers, the estimate array returned only contains my two transfers estimation when there should be also a reveal estimation.

**Steps To Reproduce**
Using an unrevealed account, the following code:
```
const batch = await Tezos.estimate
    .batch([
        {
            kind: OpKind.TRANSACTION,
            to: 'tz1ZfrERcALBwmAqwonRXYVQBDT9BjNjBHJu',
            amount: 1
        },
        {
            kind: OpKind.TRANSACTION,
            to: 'tz1ZfrERcALBwmAqwonRXYVQBDT9BjNjBHJu',
            amount: 1
        }
    ])

console.log(batch)
```
returns:
```
[
      Estimate {
        _milligasLimit: 1427000,
        _storageLimit: 0,
        opSize: 106,
        minimalFeePerStorageByteMutez: 250,
        baseFeeMutez: 100
      },
      Estimate {
        _milligasLimit: 1427000,
        _storageLimit: 0,
        opSize: 106,
        minimalFeePerStorageByteMutez: 250,
        baseFeeMutez: 100
      }
    ]
```
**Expected behavior**
It should also return an estimate for the reveal operation.

**Additional context**
The reveal estimation should be included in the array but it gets removed by this line of code: https://github.com/ecadlabs/taquito/blob/master/packages/taquito/src/contract/rpc-estimate-provider.ts#L133.

roxaneletourneau:
fix #772 

This change will lower the fees for reveal operation. The fees were set to default values and were overestimated :
before: https://edo.tzstats.com/ooM44VuZZKiz4x8FHhAjKX4iRn8iQcxmrvnzshMPfB6MK3Wu41a
after:
https://edo.tzstats.com/opKqLy2NJcu1jJb9depoUPUjDQdVhBXBpEGwoQx8rXwuXTwGBiU

I added a reveal operation on the contract API (this relates to #448).


packages/taquito/test/contract/rpc-estimate-provider.spec.ts
```diff
@@ -96,7 +97,7 @@ describe('RPCEstimateProvider test', () => {
   });
 
   describe('originate', () => {
-    it('should produce a reveal and origination operation', async done => {
+    it('should produce an origination operation, no reveal', async done => {
```
kjaiswal, line 100: By default, reveal is done on operations. Why is it that here origination operation has no reveal. Also this test doesn't test for if there is any reveal (Line 122).

packages/taquito/test/contract/rpc-estimate-provider.spec.ts
```diff
@@ -124,14 +125,35 @@ describe('RPCEstimateProvider test', () => {
   });
 
   describe('transfer', () => {
-    test('return the correct estimate for multiple internal origination', async done => {
+    test('return the correct estimate for multiple internal origination, reveal needed', async done => {
```
kjaiswal, line 128: Do we need to test for reveal op. here, whether its executed? This test only test for estimate values Line 137

kjaiswal (commented):
Maybe no changes are needed with the review and only explanation. Thank you!

packages/taquito/src/contract/rpc-estimate-provider.ts
```diff
@@ -150,15 +174,22 @@ export class RPCEstimateProvider extends OperationEmitter implements EstimationP
...
-      ...rest,
-      ...mergeLimits({ fee, storageLimit, gasLimit }, DEFAULT_PARAMS),
-    }));
-    return (await this.createEstimate({ operation: op, source: pkh }))[0];
+        ...rest,
+        ...mergeLimits({ fee, storageLimit, gasLimit }, DEFAULT_PARAMS),
+      })
+    );
+    const isRevealNeeded = await this.isRevealOpNeeded([op], pkh);
+    const ops = isRevealNeeded ? await this.addRevealOp([op], pkh) : op;
```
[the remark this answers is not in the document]
roxaneletourneau: The isRevealOpNeeded method validates if the account (tz1..) is already revealed or not by making a call to the RPC. If the account is not revealed we need to include a reveal operation because we simulate the operation with another call to the RPC (here: const { opResponse } = await this.simulate(operation);) If the account is not revealed and we do not include a reveal operation, this call fail. So we need to include a reveal operation everytime if it is needed.

packages/taquito/src/contract/rpc-estimate-provider.ts
```diff
@@ -9,29 +10,39 @@ import {
...
 }
 
+interface EstimateProperties {
+  milligasLimit: number,
+  storageLimit: number,
+  opSize: number,
+  minimalFeePerStorageByteMutez: number,
+  baseFeeMutez?: number
+}
+
```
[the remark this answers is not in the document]
roxaneletourneau: When we do the estimation, the fee is calculated, have a look at the suggestedFeeMutez method of the estimate class. We simulate the operation using the rpc, this gives us the gas that will be consumed by the operation, we use this value in the calculus and the operation size in bytes. I added a constant for the size of the reveal operation as this will not change, but the gas could change over the protocol.
I am not sure I understand your question correctly, don't hesitate to ask again if I don't answer the question.

integration-tests/contract-estimation-including-reveal-tests.spec.ts
```diff
@@ -0,0 +1,370 @@
...
+
+  const edonet = (protocol === Protocols.PtEdo2Zk) ? test : test.skip;
+  const florencenet = (protocol === Protocols.PsFLorena) ? test : test.skip;
+
+  describe(`Estimate scenario using: ${rpc}`, () => {
+    let LowAmountTez: TezosToolkit;
+    let contract: Contract;
+    const amt = 2000000 + DEFAULT_FEE.REVEAL;
+
+    beforeAll(async (done) => {
```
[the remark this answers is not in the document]
roxaneletourneau: Not the way I did it. Let's say that we do an origination operation, that a reveal is needed, and that we want to include the reveal fee in the estimation, I merged both fees (reveal and origination) in one Estimate object because I did not want to break the API... Open to suggestion

packages/taquito/test/contract/rpc-estimate-provider.spec.ts
roxaneletourneau: Reveal is done only if the account needs a reveal. This test simulates that the account does not need a reveal, so there is no reveal operation added. There is a case with an origination where a reveal is needed at line 241 below.

packages/taquito/test/contract/rpc-estimate-provider.spec.ts
roxaneletourneau: The estimation includes the fees needed for the reveal op.
This mock means that the account does not have a manager, so it needs to be revealed: mockRpcClient.getManagerKey.mockResolvedValue(null);

packages/taquito/src/contract/rpc-estimate-provider.ts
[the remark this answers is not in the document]
kjaiswal: The estimate has a total fee property that includes reveal fee, so this should be okay.

integration-tests/contract-batch-include-reveal.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+import { CONFIGS } from './config';
+
+CONFIGS().forEach(({ lib, rpc, setup, knownBaker }) => {
+    const Tezos = lib;
+    describe(`Test contract.batch using: ${rpc}`, () => {
+        beforeEach(async (done) => {
+            await setup(true);
+            done();
+        });
+        it('Batch estimate including reveal', async (done) => {
```
carte7000, line 11: Consider adding a test where `includeRevealOperation` is true but the account do not require a reveal and assert that the fees are the same as if `includeRevealOperation` was false

packages/taquito/src/contract/interface.ts
```diff
@@ -55,8 +61,19 @@ export interface EstimationProvider {
...
    */
-  registerDelegate(params?: RegisterDelegateParams): Promise<Estimate>;
-  batch(params: ParamsWithKind[]): Promise<Estimate[]>;
+  registerDelegate(params?: RegisterDelegateParams, estimationOptions?: EstimationOptions): Promise<Estimate>;
+
+  /**
+   *
+   * @description Estimate gasLimit, storageLimit and fees for a reveal operation
+   *
+   * @returns An estimation of gasLimit, storageLimit and fees for the operation or undefined if the account is already revealed
```
carte7000, line 70: What if we want to estimate a reveal operation after having an account revealed?

integration-tests/contract-estimation-tests.spec.ts
```diff
@@ -39,8 +39,8 @@ CONFIGS().forEach(({ lib, setup, knownBaker, createAddress, protocol, rpc }) =>
       }
     })
 
-    florencenet('Estimate transfer with allocated destination', async (done) => {
-      const estimate = await LowAmountTez.estimate.transfer({ to: await Tezos.signer.publicKeyHash(), amount: 1.9 });
+     florencenet('Estimate transfer with allocated destination', async (done) => {
+      const estimate = await LowAmountTez.estimate.transfer({ to: await Tezos.signer.publicKeyHash(), amount: 1.9 }, { includeRevealOperation: false });
```
carte7000, line 43: It looks like defaulting to false instead minimize the impact of this change. What are the reason we preferred defaulting to true?

packages/taquito/src/batch/rpc-batch-provider.ts
```diff
@@ -152,10 +156,16 @@ export class OperationBatch extends OperationEmitter {
    * @param params Optionally specify the source of the operation
    */
   async send(params?: { source?: string }) {
-    const estimates = await this.estimator.batch(this.operations);
+    const publicKeyHash = await this.signer.publicKeyHash();
+    const publicKey = await this.signer.publicKey();
+    const estimates = await this.estimator.batch(this.operations, { includeRevealOperation: false });
```
carte7000, line 161: I see that this change also includes a new `reveal` function in the contract API, should we consider supporting the reveal as an operation that can be added to the batch API?

If that's the case maybe it's not necessary to change this part of the code. However I understand that in the current state we have to change this in order to compensate for the lack of reveal operation support from the batch API.

packages/taquito/src/contract/rpc-estimate-provider.ts
```diff
@@ -127,16 +150,17 @@ export class RPCEstimateProvider extends OperationEmitter implements EstimationP
...
+    if (containRevealOp) {
+      numberOfOps--;
+    };
 
     return opResponse.contents.map(x => {
-      return this.createEstimateFromOperationContent(
+      return this.getEstimationPropetiesFromOperationContent(
         x,
-        opbytes.length / 2 / opResponse.contents.length,
+        // TODO: Calculate a specific opSize for each operation.
```
carte7000, line 162: We can create an issue and link it here to avoid losing track of this

packages/taquito/src/contract/rpc-estimate-provider.ts
```diff
@@ -247,14 +298,97 @@ export class RPCEstimateProvider extends OperationEmitter implements EstimationP
...
+        ...{ fee: DEFAULT_FEE.REVEAL, gasLimit: DEFAULT_GAS_LIMIT.REVEAL, storageLimit: DEFAULT_STORAGE_LIMIT.REVEAL }
+      },
+        pkh,
+        await this.signer.publicKey())
+    )
+    return op;
+  }
+
+  // Create an Estimate instance that include both reveal and operation fees
+  private createEstimateObject(estimateProperties: EstimateProperties[]) {
```
carte7000, line 366: Should we move this to the Estimate class? What if we have to duplicate this logic in the native estimate class or any other implementation?

packages/taquito/src/contract/rpc-estimate-provider.ts
```diff
@@ -247,14 +298,97 @@ export class RPCEstimateProvider extends OperationEmitter implements EstimationP
...
+      if (estimate.baseFeeMutez) {
+        baseFeeMutez = baseFeeMutez ? baseFeeMutez + estimate.baseFeeMutez : estimate.baseFeeMutez;
+      }
+    })
+    return new Estimate(milligasLimit, storageLimit, opSize, minimalFeePerStorageByteMutez, baseFeeMutez);
+  }
+
+  // Create separate instances of Estimate for each operation from the batch operation
+  private createEstimateObjectsArray(estimateProperties: EstimateProperties[]) {
+    const estimates: Estimate[] = [];
```
carte7000, line 387: Consider using `Array.map` https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Array/map

packages/taquito/src/contract/interface.ts
roxaneletourneau: This would result in an error from the RPC saying that the account is already revealed when simulating the operation.

packages/taquito/src/contract/rpc-estimate-provider.ts
roxaneletourneau: Makes sense, here is the issue: https://github.com/ecadlabs/taquito/issues/829

packages/taquito/src/contract/rpc-estimate-provider.ts
roxaneletourneau: I made use of Array.map, thank you for suggesting.

packages/taquito/src/contract/rpc-estimate-provider.ts
roxaneletourneau: Done, thanks for pointing this.

integration-tests/contract-batch-include-reveal.spec.ts
roxaneletourneau: A test case has been added, thank you.

integration-tests/contract-estimation-tests.spec.ts
roxaneletourneau: I switched this back to false by default as it makes more sense in my opinion. 

When a user specifies his own fees when doing an operation, the fees should not include fees related to the reveal operation as Taquito will take care of adding them if needed. If we returned an estimate that contains the reveal fee and the user uses this when sending an operation, it will be ending up adding two times the fees for the reveal operation.

packages/taquito/src/batch/rpc-batch-provider.ts
roxaneletourneau: I am not sure of the value added by supporting this as Taquito adds a reveal operation in the batch by itself when needed.

Issue #638, opened 2021-02-19
Ability to find version and git hash from which Taquito was built


jevonearth:
**As a developer, I want to get Taquito version, so that I can display it in Dapps or add conditional logic to my Dapp.**

**Acceptance criteria:**
- Expose version tag and git sha from the TezosToolkit instance

Note, when this is implemented, all Taquito demo dapps will display the version of Taquito that is in use for testing purposes.

- To implement, the build process will have to update constants in the Taquito source code. 
- The `getVersionInfo` method should be implemented in all the Taquito packages.
-

AlirezaHaghshenas:
I have a few questions to make sure my understanding of this issue is right:
1. In CI, some new constants in `src/constants.ts` should be updated
2. These constants be exposed in a new method in `src/taquito.ts`
3. The changes made in CI to the `src/constants.ts` will not be pushed back to the repository (so we understand the real source code that produces the package is not available, one side effect of this is that if we do a manual publish outside of CI/CD, these values are not right, unless explicitly updated manually)

AlirezaHaghshenas:
OK. I have a better way for hash: build CI should pass it as build-time environment variable.

jevonearth:
Hi @AlirezaHaghshenas 

1. I wonder could we revert the change back after the build in a post build hook? (Is this a horrible thing to do? :thinking_face: )
2. We could do this action only when making releases or pre-view builds, so it doesn't happen when a developer is building locally (this will negate idea 1 above)
3. Let's not commit the updated hash, we can set it at build, but in git it should only ever be a placeholder
4. Sounds good, but for preview builds you might have to handle it differently. You have two cases, real releases (which we still do semi-manually) and preview builds. For preview builds, I think the version should contain the branch name and git short hash.

PR #812, opened 2021-04-26
Issue-638 Removed post install step

kjaiswal:

## Release Note Draft Snippet


roxaneletourneau (commented):
The issue #638 mentions :

> The getVersionInfo method should be implemented in all the Taquito packages.

Looking at the PR, I understand that the version will be accessible for @taquito/taquito package because of the `getVersionInfo` method added on the `TezosToolkit` class, but how will we retrieve the version if we install another package?

kjaiswal:
> The issue #638 mentions :
> 
> > The getVersionInfo method should be implemented in all the Taquito packages.
> 
> Looking at the PR, I understand that the version will be accessible for @taquito/taquito package because of the `getVersionInfo` method added on the `TezosToolkit` class, but how will we retrieve the version if we install another package?

As per the release process in notion we do _npm build_ before we deploy packages to npm. This creates version.ts file in src folder of each package which is then used by dapp to display version info.

Issue #673, opened 2021-03-05
Taquito Docs  upgrade guide page needs to be updated from V7 to V8


michaelkernaghan:
The Taquito docs page  for Upgrading to v7/delphi needs to be updated to Upgrading to v8/edo.

1. Go to https://tezostaquito.io/docs/v7_breaking_changes
2. This page should be updated for info on changing to V8

PR #813, opened 2021-04-26
Upgrading v9

roxaneletourneau:
#673 

Changed page `Upgrading to v7/delphi` to `Upgrading Guide` and added instructions for v8 and v9


michaelkernaghan (approved):
Looks good to me!

jevonearth (approved):

PR #879, opened 2021-05-19
Issue-817 Issue-862 Rounded off polling interval to ceiling and inclu…

### LINKED ISSUE #862: Error: Interval must be more than 0
Since v8.1.1 a new dynamic polling interval was intoduced. Using taquito with a local development chain sets the interval to 0, which throws this Error: `Error: Interval must be more than 0`

The error occurred for me trying to originate a contract
```typescript
const contract = await this.tezos.contract
  .originate({ code, storage })
  .then((originationOperation) => originationOperation.contract());
```

setting the polling interval manually fixes this issue: 
```typescript
this.tezos.setProvider({ signer: this.signer, config: { confirmationPollingIntervalSecond: 1 } })
```

kjaiswal:
Resolves #862

### DIFF
--- a/packages/taquito/src/context.ts
+++ b/packages/taquito/src/context.ts
@@ -173,7 +173,11 @@ export class Context {
   async getConfirmationPollingInterval() {
     try {
       const constants = await this.rpc.getConstants();
-      let confirmationPollingInterval = BigNumber.sum(constants.time_between_blocks[0], 
+      let blockTime = constants.time_between_blocks[0];
+      if (constants.minimal_block_delay !== undefined) {
+        blockTime = constants.minimal_block_delay;
+      }
+      let confirmationPollingInterval = BigNumber.sum(blockTime, 
         new BigNumber(constants.delay_per_missing_endorsement!)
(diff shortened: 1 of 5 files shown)


packages/taquito/src/context.ts
```diff
@@ -183,6 +187,10 @@ export class Context {
       // testnet and sandbox enviornment.   
       confirmationPollingInterval = confirmationPollingInterval.dividedBy(3);  
       this.config.confirmationPollingIntervalSecond = confirmationPollingInterval.toNumber();
+      
+      if (this.config.confirmationPollingIntervalSecond === 0) {
+        return 0.1;
+      }
```
roxaneletourneau, line 193: We should do the validation (that looks if the value is 0) before setting it in the config, otherwise, we will still use the 0 for the pollinginterval in the subsequent requests.
If `confirmationPollingInterval.toNumber() === 0`, we need to set the `confirmationPollingIntervalSecond` to 0.1 and not 0.

roxaneletourneau (changes requested):

packages/taquito/src/context.ts
kjaiswal: Thanks, this is fixed.

PR #880, opened 2021-05-19
Added a test cases with 150 operations in a batch

roxaneletourneau:
With the changes in galpha2net, the estimation is not working with a batch having a high number of operations. 
Temporarily fix is to explicitly specify the gasLimit instead of relying on the estimate.

### DIFF
--- a/integration-tests/contract-batch-high-number-of-operations.spec.ts
+++ b/integration-tests/contract-batch-high-number-of-operations.spec.ts
@@ -1,0 +1,42 @@
+import { CONFIGS } from './config';
+import { b58cencode, Prefix, prefix } from '@taquito/utils';
+import { InMemorySigner } from '@taquito/signer';
+const crypto = require('crypto');
+
+CONFIGS().forEach(({ lib, rpc, setup }) => {
+    const Tezos = lib;
+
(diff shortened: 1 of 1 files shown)


integration-tests/contract-batch-high-number-of-operations.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+            dests.forEach(({ pkh }) => {
+                // We currently need to bypass the estimate for gapha2net as the hard_gas_limit_per_block is exceeded
+                batch.withTransfer({ to: pkh, amount: 0.001, gasLimit: 1527});
+            })
+
+            const op = await batch.send();
+            await op.confirmation();
+
+            expect(op.status).toEqual('applied');
+            console.log(op.hash)
```
kjaiswal, line 38: Do you want to keep this log?

kjaiswal (changes requested):

integration-tests/contract-batch-high-number-of-operations.spec.ts
roxaneletourneau: Removed, thanks!

kjaiswal (approved):

PR #909, opened 2021-06-04
Added tokens Bls12-381 to the Michelson Encoder

### LINKED ISSUE #888: Add a token Bls12-381 to the Michelson Encoder

The type Bls12-381 was added in protocol 008, but is missing in the Michelson Encoder.

**Additional context**
https://tezos.gitlab.io/protocols/008_edo.html#bls12-381

roxaneletourneau:
closes #888 

## Release Note Draft Snippet

### DIFF
--- a/packages/taquito-michelson-encoder/src/tokens/bls12-381-fr.ts
+++ b/packages/taquito-michelson-encoder/src/tokens/bls12-381-fr.ts
@@ -1,0 +1,31 @@
+import { Token, TokenFactory } from './token';
+
+export class Bls12381frToken extends Token {
+  static prim = 'bls12_381_fr';
+
+  constructor(
+    protected val: { prim: string; args: any[]; annots: any[] },
+    protected idx: number,
+    protected fac: TokenFactory
+  ) {
+    super(val, idx, fac);
+  }
+
+  Encode(args: any[]): any {
+    const val = args.pop();
+    return { bytes: val.toString() };
+  }
+
+  EncodeObject(val: any): any {
+    return { bytes: val.toString() };
+  }
+
+  Execute(val: any): string {
+      console.log('execute', val)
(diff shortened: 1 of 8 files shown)


packages/taquito-michelson-encoder/src/tokens/bls12-381-fr.ts
```diff
@@ -0,0 +1,62 @@
...
+    protected fac: TokenFactory
+  ) {
+    super(val, idx, fac);
+  }
+
+  private isValid(val: any): Bls12381frValidationError | null {
+    if (typeof val === 'string' && /^[0-9a-fA-F]*$/.test(val) && val.length % 2 === 0) {
+      return null;
+    } else {
+      return new Bls12381frValidationError(val, this, `Invalid bytes: ${val}`);
```
kjaiswal, line 26: Would it provide more value if we add condition (out of above 3) on which validation failed to the returned error?

kjaiswal:
What is the difference between 3 token types except their prim values. Is there a way we can avoid 3 files and make one that defines the 3 tokens?

roxaneletourneau:
> What is the difference between 3 token types except their prim values. Is there a way we can avoid 3 files and make one that defines the 3 tokens?

These are 3 different Michelson types that I represented with 3 different classes that implement the token class. Can you give more detail on the solution you have in mind?

packages/taquito-michelson-encoder/src/tokens/bls12-381-fr.ts
roxaneletourneau: I'll make the change, thank you!

kjaiswal:
I reckon each token type in Michelson has its own type file. So this should be okay.

kjaiswal (approved):

Issue #602, opened 2021-02-05
Consider implementing a sapling_state abstraction


roxaneletourneau:
When accessing a sapling_state in the storage, a int value is returned. (similar with big_map)
Then `<block_id>/context/sapling/<sapling_state_id>/get_diff` returns the root and a diff of state.

Examples: 
The contract `KT1KnF4vZMgTRgy9hniwV9pYpPj4b7HEV8fU` is having `{"int":"56"}` as `sapling_state` value in its storage.
The following value is returned when calling
https://api.tez.ie/rpc/edonet/chains/main/blocks/head/context/sapling/56/get_diff : 
```
{
  "root":"fbc2f4300c01f0b7820d00e3347c8da4ee614674376cbc45359daa54f9b5493e",
  "commitments_and_ciphertexts":[ ],
  "nullifiers":[ ]
}
```
See also: https://api.tez.ie/rpc/edonet/chains/main/blocks/head/context/contracts/KT1KnF4vZMgTRgy9hniwV9pYpPj4b7HEV8fU/single_sapling_get_diff

Issue #586, opened 2021-01-24
Sapling Integration


RomarQ:
**As a developer, I would like to originate and/or retrieve a sapling contract, so that I can start using private transaction feature(sapling) of Edo protocol.**

**Acceptance criteria:**
- Add support of two new Michelson types 'sapling_state' and 'sapling_transaction'
- Add support of two new Michelson instructions called SAPLING_VERIFY_UPDATE and SAPLING_EMPTY_STATE (see the Michelson [reference ](https://tezos.gitlab.io/008/michelson.html#michelson-008) for more details).
- Contract using the above instructions and types can be originated and retrieved through Taquito.
- Add new RPC to RPCClient:
   - `../<block_id>/context/contracts/<contract_id>/single_sapling_get_diff` (GET)
   - `../<block_id>/context/sapling/<sapling_state_id>/get_diff` (GET)

OUT OF SCOPE:
- No additional abstractions or ability to decrypt Sapling transactions would be available with this ticket

NOTE:
When originating a sapling contract, the operation is injected successfully but **taquito-michelson-encoder** fails to parse the contract after the first confirmation:
`Malformed data expected a value with a valid prim property`
This is most likely caused by the missing `sapling` tokens.

[Example contract](https://edo.tzstats.com/KT1QHwX65rW6pdVuk3KrT3gQHsSCBzcGLgaq)

This issue extends #528

Innkst:
short description on sapling state and overall workflow: https://www.reddit.com/r/tezos/comments/l378dx/test_edonet_sapling/gke48bc?utm_source=share&utm_medium=web2x&context=3

PR #935, opened 2021-06-18
Basic Sapling Integration 

roxaneletourneau:
#586
#602 

- Deploy/interact with a contract having sapling types
- Access the sapling diff via the SaplingStateAbstraction class when fetching the storage of a contract (that has a type sapling_state in its storage)
- Added the methods getSaplingDiffByContract and getSaplingDiffById on the RpcClient to retrieve a sapling state diff.

## Release Note Draft Snippet


packages/taquito/src/contract/semantic.ts
```diff
@@ -18,6 +19,14 @@ export const smartContractAbstractionSemantic: (p: ContractProvider) => Semantic
       return new BigMapAbstraction(new BigNumber(val.int), schema, provider);
     }
   },
+  sapling_state: (val: MichelsonV1Expression) => {
+    if (!val || !('int' in val) || val.int === undefined) {
+      // Return an empty object in case of missing sapling state ID
+      return {};
```
kjaiswal, line 25: Should we provide an error if id is missing?

roxaneletourneau: I don't think we should because this code is executed when we fetch the storage of a contract from the chain and provide an abstraction over it. In fact, this condition should not even happen.. but if it does for any reason, I think it is better to provide an empty sapling state and still return the other values of the storage instead of throwing an exception which the user doesn't have control over. Let me know what you think.

kjaiswal: Sounds good, thanks for explaining that!

kjaiswal (approved):

PR #990, opened 2021-07-23
Issue-960 Changes for ExtractSchema for token type list

### LINKED ISSUE #960: Schema.ExtractSchema() returns "list"
**Description**

The method [`ExtractSchema()` of `Schema`](https://tezostaquito.io/docs/michelson_encoder/#the-extractschema-method) returns string `"list"` for a complex Michelson starting with a `{ prim: 'list' }`. Or am I wrong and this is an expected exception?

**Steps To Reproduce**
1. Run following code.
    ```typescript
    const storageType = {
        prim: 'list',
        args: [
            {
                prim: 'pair',
                args: [
                    { prim: 'address', annots: ['%from'] },
                    { prim: 'address', annots: ['%to'] },
                ],
            },
        ],
    };
    const storageSchema = new Schema(storageType);
    const extractSchema = storageSchema.ExtractSchema();
    println(JSON.stringify(extractSchema, null, 2));
    ```
2. It prints (returns) string `"list"`.

**Expected behavior**
Expected to return complex object like:
```typescript
[
    {
        "from": "address",
        "to": "address"
    }
]
```

kjaiswal:
Resolves #960

### DIFF
--- a/packages/taquito-michelson-encoder/src/tokens/list.ts
+++ b/packages/taquito-michelson-encoder/src/tokens/list.ts
@@ -69,8 +69,7 @@ export class ListToken extends Token {
   public ExtractSchema() {
     const valueSchema = this.createToken(this.val.args[0], this.idx);
     return {
-      prim: ListToken.prim,
-      value : valueSchema.ExtractSchema()
+      [ListToken.prim] : valueSchema.ExtractSchema()
     }
   }
 }
--- a/packages/taquito-michelson-encoder/src/tokens/list.ts
+++ b/packages/taquito-michelson-encoder/src/tokens/list.ts
@@ -67,6 +67,10 @@ export class ListToken extends Token {
   }
 
   public ExtractSchema() {
-    return ListToken.prim;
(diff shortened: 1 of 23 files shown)


packages/taquito-michelson-encoder/src/tokens/list.ts
```diff
@@ -67,6 +67,10 @@ export class ListToken extends Token {
   }
 
   public ExtractSchema() {
-    return ListToken.prim;
+    const valueSchema = this.createToken(this.val.args[0], this.idx);
+    return {
```
roxaneletourneau, line 71: The main purpose of MichelsonEncoder is to abstract Michelson language and "prim" is really something proper to the Michelson language. Thus, I don't think we should return a "prim" in the schema.
An alternative could be:
```
return {
ListToken.prim: valueSchema.ExtractSchema()
}
```

kjaiswal: This has been fixed.

roxaneletourneau (approved):

PR #1044, opened 2021-09-08
Key ordering

skubarenko:

## Release Note Draft Snippet
Fixed key sorting in literal sets and maps when these collections have mixed key types.

## Details
Before this fix, we couldn't send key sets containing multiple key types (Ed25519, Secp256k1, P256) to RPC, e.g.
```json
[
   "edpkvS5QFv7KRGfa3b87gg9DBpxSm3NpSwnjhUjNBQrRUUR66F7C9g",
   "sppk7aVdgAmezMCRTcHciVkVZoGNnhSdKEYcn5pYaqt4PvLjgFbLRxo",
   "p2pk65shUHKhx7zUSF7e8KZ2inmQ5aMS4jRBUmK6aCis4oaHoiWPXoT"
]
```
If we try to use this set we'll get an error `michelson_v1.unordered_set_literal`, as the michelson encoder sorts the set using the default compare method, method of the `ComparableToken` class. In other words, the encoder sorts the set using alphabetic sorting.
Unfortunately, I've found no description of the expected behavior in docs for this case, but the [current implementation says](https://gitlab.com/tezos/tezos/-/blob/master/src/lib_crypto/signature.ml#L268-277):
1. Ed25519 keys should go first, then Secp256k1 keys, and P256 keys finally.
2. if two keys are the same type:
    * if the type is Ed25519  -> compare keys using the alphabetical comparison;
    * is Secp256k1 ->  compare keys using the alphabetical comparison;
    * is P256  -> ~~alphabetical comparison~~ use the custom comparison, I wrote details below;

### Comparison algorithm of P256 keys
If we try to use the following set as literal:
```
[
   "p2pk67c5b5THCj5fyksX1C13etdUpLR9BDYvJUuJNrxeGqCgbY3NFpV",
   "p2pk66xmhjiN7LpfrDGFwpxPtJxkLtPjQ6HUxJbKmRbxSR7RMpamDwi",
   "p2pk65shUHKhx7zUSF7e8KZ2inmQ5aMS4jRBUmK6aCis4oaHoiWPXoT"
]
```
We also get the error, `michelson_v1.unordered_set_literal`. The P256 keys use the custom comparison that I haven't found an implementation of. I suppose the implementation is [located somewhere here](https://gitlab.com/tezos/tezos/-/blob/master/src/lib_crypto/p256.ml), but I don't know OCaml :) 
After several tests I guess the algorithm probably is the following:
1. Decode keys from b58;
2. Compare receiving bytes ignoring the first byte;
Maybe I'm wrong, but I've tried to test on different keys, and I couldn't reproduce the error.

<details>
  <summary>Source Code of the test contract (ReasonLIGO)</summary>

```reason
type storage = {
    keySet: set(key),
    keyMap: map(key, int)
};

type parameter =
    | SetSet (set(key))
    | SetMap (map(key, int));

type main_result = (list (operation), storage);
   
let main = ((action, storage) : (parameter, storage)) : main_result => (
 ([]: list(operation)),
 switch (action) { 
  | SetSet(new_set) => { ...storage, keySet: new_set }; 
  | SetMap(new_map) => { ...storage, keyMap: new_map }; 
 }
)
```
</details>

### DIFF
--- a/integration-tests/contract-with-key-collections.spec.ts
+++ b/integration-tests/contract-with-key-collections.spec.ts
@@ -1,0 +1,77 @@
+import { CONFIGS } from "./config";
+import { MichelsonMap } from "@taquito/taquito";
+import { contractWithKeyCollections } from "./data/contract-with-key-collections";
+
+CONFIGS().forEach(({ lib, rpc, setup }) => {
+  const Tezos = lib;
+
+  describe(`Test contract with key collections using: ${rpc}`, () => {
+    type Storage = { keySet: string[], keyMap: MichelsonMap<string, number>; }
+
+    beforeEach(async (done) => {
+      await setup();
+      done();
+    });
+
+    it('Originate a contract with set and map of keys and change them using corresponding methods', async (done) => {
+      const initialStorage: Storage = {
+        keySet: [
+          'edpkvS5QFv7KRGfa3b87gg9DBpxSm3NpSwnjhUjNBQrRUUR66F7C9g',
+          'edpkuNjKKT48xBoT5asPrWdmuM1Yw8D93MwgFgVvtca8jb5pstzaCh',
+          'sppk7aVdgAmezMCRTcHciVkVZoGNnhSdKEYcn5pYaqt4PvLjgFbLRxo',
+          'p2pk67c5b5THCj5fyksX1C13etdUpLR9BDYvJUuJNrxeGqCgbY3NFpV',
+          'p2pk65shUHKhx7zUSF7e8KZ2inmQ5aMS4jRBUmK6aCis4oaHoiWPXoT',
+          'p2pk6842BMz2Se9XuMxQDe7yVdaEvpjNGQ7DYjKTQwzMUtryaHmZcV9',
+        ],
+        keyMap: MichelsonMap.fromLiteral({
+          'edpkuNjKKT48xBoT5asPrWdmuM1Yw8D93MwgFgVvtca8jb5pstzaCh': 10,
+          'p2pk65shUHKhx7zUSF7e8KZ2inmQ5aMS4jRBUmK6aCis4oaHoiWPXoT': 40,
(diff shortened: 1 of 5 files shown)


roxaneletourneau:
Thank you very much @skubarenko for submitting this PR. The explanations and test cases make it very enjoyable to review :)
I did some more testing regarding the custom comparison for p2pk, and it seems to work as expected.
However, before merging, I'll do some research to try to validate that ignoring the first byte is the way to go.

skubarenko:
@roxaneletourneau Yeah, sure. If you'll find a clearer algorithm of the P256 comparison, please, let me know, I'll update the PR.
Because the current one is really strange :)

roxaneletourneau:
@skubarenko  After some digging, it makes sense to me to remove the first byte based on the following:
The value returned by b58cdecode corresponds to the "compressed public key".
In elliptic curve cryptography, public keys are points (coordinates x,y) on the curve.
The compressed public key corresponds to the x coordinate to which a parity byte is appended based on the parity of the y coordinate; 02 if even or 03 if odd.

So, the byte that is ignored [here](https://github.com/ecadlabs/taquito/pull/1044/files#diff-199cd712f7062221034dd81e3b5cec35c2c9e7ffc6b9183ffe1bab2fc3227fbdR103) corresponds to this added parity byte.

By looking in the code, the following part seems to be the one responsible for removing the first byte: https://gitlab.com/tezos/tezos/-/blob/master/src/lib_hacl_glue/js/src/hacl.ml#L622
thanks to @mweichert for digging into the codebase :)  

Thanks again for your contribution.

roxaneletourneau (approved):

skubarenko:
@roxaneletourneau great, thanks for the explanation. I see that some tests have failed, as I understand this is not related to my changes. But if some actions on my part are needed ping me :)

roxaneletourneau:
The fix has been released in version 10.1.3 :)

PR #1080, opened 2021-09-28
RpcClientCache

roxaneletourneau:

## Release Note Draft Snippet


docs/rpc-cache.md
```diff
@@ -0,0 +1,22 @@
+---
```
kjaiswal, line 1: Few points to consider - 

1. Please consider providing a third parameter to cache constructor which could be an array of rpc endpoints which we don't want to be cached. It would be useful to avoid presenting stale data to user in cases where data on chain changes rapidly.
2. Consider putting a test case to test data before and after ttl expires.

Rest looks great.

roxaneletourneau: The third param is a good idea. However, data can not really been updated faster than the time between blocks and the TTL of 1 second is much lower. I will let it as it for now and we can consider adding the parameter later on as a separate task.

PR #1182, opened 2021-11-16
Refactor(michel-codec) use newest protocol behaviour by default

roxaneletourneau:
I changed the logic of the if/else statements where we act differently according to the protocol hash.

The behavior regarding macro expansion is different since the Edo protocol. There were checks in the codebase (using ProtoGreaterOfEqual) to see if the protocol used was higher than Edo and if so to apply the "Edo Since" behavior. However, if a user was using Taquito with a new protocol that is not yet supported (not part of the Protocol enum), the default behavior to expand macro was the "before Edo". I made a change so the  "Edo Since" behavior is used by default unless the protocol used is inferior to Edo (which is less likely).


e-asphyx (approved):
Yes, sounds very reasonable. Thank you, Roxane.

PR #1185, opened 2021-11-17
(Docs) warning about signing data packed with the RPC

roxaneletourneau:

## Release Note Draft Snippet


docs/signing.md
```diff
@@ -124,33 +124,40 @@ const { signature } = signedPayload;
...
 ```
 
 First, you provide the Michelson code to be signed as a string along with its type.  
 Then, you create a new instance of the `michel-codec` parser and call the `parseMichelineExpression` on it to get the JSON representation of the Michelson code and type.  
-Once done, you can pack the data using the `packData` method available on `Tezos.rpc` (after creating an instance of the `TezosToolkit`).  
+Once done, you can pack the data using the `packDataBytes` function available in the `@taquito/michel-codec` package.  
 To finish, use one of the methods presented above to sign the packed data (with the `InMemorySigner` like in this example or with the Beacon SDK).
 
+:::caution
+In the previous example, the data are packed locally using the `packDataBytes` function of the `@taquito/michel-codec` package instead of the RPC. Users should be cautious and verify the packed bytes before signing them if they are packed using the RPC. This precaution is to avoid signing unwanted operations in cases where the RPC could be malicious or compromised.
```
jevonearth, line 158: Change to:

> In the previous example, the data is packed locally in Taquito using the `packDataBytes` function of the `@taquito/michel-codec` package instead of the RPC. You should always verify the packed bytes before signing or requesting that they be signed when using the the RPC to pack. This precaution helps protect you and your applications users from RPC nodes that have been compromised. A node that is operated by a bad actor, or compromised by a bad actor could return a fully formed operation that does not correspond to the input provided to the RPC endpoint.

packages/taquito-rpc/src/taquito-rpc.ts
```diff
@@ -801,9 +801,13 @@ export class RpcClient implements RpcClientInterface {
    * @param options contains generic configuration for rpc calls
    *
    * @description Computes the serialized version of a data expression using the same algorithm as script instruction PACK
+   * Note: Users should be cautious if signing data packed with the RPC. They should always verify the packed bytes before signing them. 
+   * This precaution is to avoid signing unwanted operations in cases where the RPC could be malicious or compromised. 
+   * The safest solution to pack and sign data would be to use the `packDataBytes` function available in the `@taquito/michel-codec` package.
```
jevonearth, line 806: Use same rewrite as I provided above please.

jevonearth (changes requested):

docs/signing.md
roxaneletourneau: Thank you, I made the change

packages/taquito-rpc/src/taquito-rpc.ts
roxaneletourneau: Thank you, I made the change

PR #1200, opened 2021-11-19
fix contract reference in doc page

michaelkernaghan:
There is a link to a contract on BCD that is for the wrong chain and needs to be updated.


roxaneletourneau (changes requested):
The storage of thecontract KT1CfFBaLoUrgv93k8668KCCcu2hNDNYPz4L does not match the description: "a map that contains addresses as keys and tez as values"

PR #1201, opened 2021-11-19
Global constants provider

### LINKED ISSUE #1219: Expand global constants with the michel-codec parser
Provide the ability to expand global constant in a contract script.

Add an optional configuration to the parser that will allow replacing global constant hashes in a script with their corresponding Michelson values.

### LINKED ISSUE #1220: Support contract origination using the storage property when there are global constants in the storage part of the contract code
As a user, I want to deploy a smart contract using the new Tezos global constant feature and still be able to use the storage property provided by Taquito. When there is a global constant in the storage part of the contract code, it is currently not possible to deploy the contract using the storage property and the init property needs to be used instead.

Expand the global constant in the script using the michel-codec parser. Use this expanded script with the Michelson-Encoder which transforms the storage property into Michelson data.

**Describe alternatives you've considered**
Instead of expanding the global constants using the michel-codec parser, expand them in place in the GlobalConstantToken of the Michelson-Encoder. This becomes rapidly complex when there are multiple constants because of how the package deals with annotations. The easiest/safest solution is to provide an expanded script to the Michelson-Encoder.

roxaneletourneau:
Closes #1219 
Closes #1220

## Release Note Draft Snippet


packages/taquito-michel-codec/test/helpers/global-constants-helpers.ts
```diff
@@ -0,0 +1,65 @@
+import { Expr } from '../../src/micheline';
+
```
dsawali, line 2: very minor nitpick, but what do you think about renaming the file to `global-constants-helper.ts` instead for consistency. I think generally file types in names tend to be singular instead of plurals

packages/taquito-michel-codec/test/global-constant.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+  const parserOptions: ParserOptions = {
+    expandGlobalConstant: {
+      constantHashInt: registeredIntExprJSON,
+      constantHashDrop: registeredDropExprJSON,
+    },
+  };
+
+  const p = new Parser(parserOptions);
+
+  it('Expands global constants in script using parseJSON', () => {
```
dsawali, line 22: this might just be my experience as an SDET speaking, but I think the way our describe blocks are written it's better to use the 'Should...' model of writing test descriptions.

So it will be written like 'Should expand global constants in script using parseJSON'. Seeing that most of our tests already have that description format, is that something we are willing to continue adopting?

packages/taquito/src/parser/michel-codec-parser.ts
```diff
@@ -32,6 +34,25 @@ export class MichelCodecParser implements ParserProvider {
         parsedParams.code = await this.formatCodeParam(params.code);
         if (params.init) {
             parsedParams.init = await this.formatInitParam(params.init);
+        } else if (params.storage) {
+            const storageType = (parsedParams.code as Expr[]).find((p): p is Prim => ('prim' in p) && p.prim === 'storage');
+            if (storageType?.args === undefined) {
```
dsawali, line 39: we might be able to shorten this to just 
`if (!storageType?.args)`

packages/taquito/test/global-constants/default-global-constants-provider.spec.ts
```diff
@@ -0,0 +1,113 @@
+import { DefaultGlobalConstantsProvider } from '../../src/global-constants/default-global-constants-provider';
+
+describe('DefaultGlobalConstantsProvider tests', () => {
+    it('DefaultGlobalConstantsProvider is instantiable', () => {
+        expect(new DefaultGlobalConstantsProvider()).toBeInstanceOf(DefaultGlobalConstantsProvider);
+    });
+
+    it('register a global constant and retrieve the value by its hash', async (done) => {
+        const hash = 'expru5X5fvCer8tbRkSAtwyVCs9FUCq46JQG7QCAkhZSumjbZBUGzb';
+        const michelineExpr = {
```
dsawali, line 10: looks like this michelineExpression is being reused a few times in the file, maybe we could initialize it once above the describe block and reuse that in the it blocks?

packages/taquito-michel-codec/test/helpers/global-constants-helpers.ts
roxaneletourneau: Thanks, I made the change.

packages/taquito-michel-codec/test/global-constant.spec.ts
roxaneletourneau: Good catch, I made the changes. Better to be consistent.

packages/taquito/src/parser/michel-codec-parser.ts
roxaneletourneau: Thanks, I made the change, it is cleaner like that.

packages/taquito/test/global-constants/default-global-constants-provider.spec.ts
roxaneletourneau: Good catch :)

PR #1205, opened 2021-11-22
rewrite check balances function in docs live code examples origination script

### LINKED ISSUE #1202: Rewrite check_balances in the Taquito Docs live code examples origination script - example/deploy-docs-live-code-contracts
**Description**
in the script /home/mike/taquito/example/deploy-docs-live-code-contracts.ts
the async function checkBalances()  uses asynchronous calls inside a forEach loop
and when run trips over itself by requesting several operations at the same time.

The function should be rewritten along these lines:

> "The idea is to use Promise.all and use .map instead of .forEach
> The map returns an array of promises and Promise.all waits until all the promises are resolved
> But if one is rejected, everything fails
> If you expect 1 or multiple promises to reject but you don’t want the whole thing to fail, you use Promise.allSettled instead"

michaelkernaghan:
Closes #1202

### DIFF
--- a/example/deploy-docs-live-code-contracts.ts
+++ b/example/deploy-docs-live-code-contracts.ts
@@ -35,78 +35,94 @@ export const tezos = new TezosToolkit(provider);
 
 let contract_catalogue = new Map();
 
+const users: Array<string> = [
+  'tz1bwsEWCwSEXdRvnJxvegQZKeX5dj6oKEys',
+  'tz1VSUr8wwNhLAzempoch5d6hLRiTh8Cjcjb',
+  'tz1h3rQ8wBxFd8L9B3d7Jhaawu6Z568XU3xY',
+  'tz3WXYtyDUNL91qfiCJtVUX746QpNv5i5ve5',
+  'tz1KqTpEZ7Yob7QbPE4Hy4Wo8fHG8LhKxZSx',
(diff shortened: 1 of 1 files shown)


example/deploy-docs-live-code-contracts.ts
```diff
@@ -35,80 +35,95 @@ export const tezos = new TezosToolkit(provider);
...
-  contract_catalogue.set('Tzip16OffChainTwo', await originateTzip16OnChainTwo());
-  contract_catalogue.set('WalletContract', await originateWalletOriginateContractTransfer());
-  contract_catalogue.set('WalletAreYouThereContract', await originateWalletOriginateAreYouThere());
-  contract_catalogue.set('TokenContract', await originateTokenContract());
-  contract_catalogue.set('BigMapPackContract', await originateBigMapPackContract());
-  contract_catalogue.set('MichelsonMap', await originateMichelsonMap());
-
-  json_contract_catalogue();
-
+  checkBalances(users);
```
roxaneletourneau, line 98: I suggest awaiting here or the script will start deploying contracts before the checkBalances is done.
`await checkBalances(users);`

michaelkernaghan: I have made the change, thanks

Issue #1109, opened 2021-10-11
Schema.Execute() fails if timestamp as unix time string


psmolinsky:
## Description

When there is a timestamp (unit time seconds) in Michelson in a string then `Schema`/`ParameterSchema` fails to `Execute(...)` it.

## Steps To Reproduce

1. Create `Schema` object.
  ```typescript
    const schema = new Schema({ "prim":"timestamp" });
  ```
2. Prepare the Michelson data.
  ```typescript
    const data = { "string":"1613034908" };
  ```
3. Run `Execute(...)`.
  ```typescript
    const value = schema.Execute(data);
  ```
4. It throws an error.
  ```
  RangeError: Invalid time value
    at Date.toISOString
    at TimestampToken.Execute
  ```

## Additional context

An example in actual blockchain is block `BL9uEsioAdcpbj71GZ2DzBWM9yyAq6g3inZhvEXHZDzimSdzPuj` with operation `ooXpuCVoChXXLgzLpsjFUJpcB8yAd9a3974jJZakSckUVLGLrz9` and its transaction parameters, respective part is: `{ "string":"1613034908" }` which corresponds to schema: `{ "prim":"timestamp", "annots": ["%update_time"] }`.

Corresponding Taquito code is [TimestampToken class](https://github.com/ecadlabs/taquito/blob/master/packages/taquito-michelson-encoder/src/tokens/comparable/timestamp.ts#L15-L16).

PR #1232, opened 2021-12-03
1109 unix timestamp

dsawali:

## Release Note Draft Snippet


dsawali:
This PR addresses [issue #1109](https://github.com/ecadlabs/taquito/issues/1109)

Added extra typecheck and added relevant tests to the timestamp class

packages/taquito-michelson-encoder/test/tokens/timestamp.spec.ts
```diff
@@ -0,0 +1,41 @@
+import { TimestampToken } from '../../src/tokens/comparable/timestamp';
+
+describe('Timestamp token', () => {
+  let token: TimestampToken;
+
+  beforeEach(() => {
+    token = new TimestampToken({ prim: 'timestamp', args: [{ string: '1638554817'}], annots: [] }, 0, null as any);
```
roxaneletourneau, line 7: The args should be empty here as the token only represents the Michelson type, not the value.

packages/taquito-michelson-encoder/test/tokens/timestamp.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+  });
+
+  describe('Execute', () => {
+    it('Should return the RFC3339 time format when type is int and value is UNIX timestamp', () => {
+      const result = token.Execute({ int: 1638555790 } as any);
```
roxaneletourneau, line 12: { int: 1638555790 } should be { int: '1638555790' } and thus no need to cast to any.

packages/taquito-michelson-encoder/test/tokens/timestamp.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+      expect(result).toEqual('2021-12-03T18:23:10.000Z');
+    });
+
+    it('Should return the RFC3339 time format when type is string and value is UNIX timestamp', () => {
+      const result = token.Execute({ string: '1638555790' } as any);
```
roxaneletourneau, line 17: I don't think there is a need to cast to any here as the expected type of the parameter is `{ string?: string; int?: string }`.

packages/taquito-michelson-encoder/test/tokens/timestamp.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+      expect(result).toEqual('2021-12-03T18:23:10.000Z')
+    });
+
+    it('Should return an error when value is not a valid Date or timestamp format', () => {
+      expect(() => {
+        token.Execute({ string: 'not valid' } as any);
```
roxaneletourneau, line 23: Same as the precedent comment, no need to cast to any

packages/taquito-michelson-encoder/src/tokens/comparable/timestamp.ts
```diff
@@ -12,7 +12,10 @@ export class TimestampToken extends ComparableToken {
   }
 
   public Execute(val: { string?: string; int?: string }) {
-    if (val.string) {
+    if (val.string && /^\d{10}(\.\d)?$/.test(val.string)) {
+      const timestamp = +val.string;
+      return new Date(Number(timestamp) * 1000).toISOString();
```
roxaneletourneau, line 17: "+" and "Number" have the same behavior of converting the string to a number. I suggest keeping only one of them.

packages/taquito-michelson-encoder/src/tokens/comparable/timestamp.ts
```diff
@@ -12,7 +12,10 @@ export class TimestampToken extends ComparableToken {
   }
 
   public Execute(val: { string?: string; int?: string }) {
-    if (val.string) {
+    if (val.string && /^\d{10}(\.\d)?$/.test(val.string)) {
```
roxaneletourneau, line 15: If we have a Unix timestamp that refers to the 2000s or prior, the regex will fail as the number of digits will be < 10.
Should the check be less restrictive?

roxaneletourneau (changes requested):

packages/taquito-michelson-encoder/src/tokens/comparable/timestamp.ts
dsawali: you are correct, I missed the `Number()` casting

packages/taquito-michelson-encoder/src/tokens/comparable/timestamp.ts
dsawali: Good point, I can make it less restrictive. Do you think 9-10 digits is enough? or should we be making a larger range like 8-10?

packages/taquito-michelson-encoder/test/tokens/timestamp.spec.ts
dsawali: oops nice catch, I meant them to be empty.

packages/taquito-michelson-encoder/src/tokens/comparable/timestamp.ts
roxaneletourneau: What do you think about just validating that the string is made of digits regardless of the length?

dsawali: we can definitely do that too, I can make that change then

roxaneletourneau (approved):
Great work :)

PR #1251, opened 2021-12-09
On-chain view documentation page

### LINKED ISSUE #1117: Abstraction to execute contract views
We want to provide a new feature on the ContractAbstraction class that will allow users to execute the new views proposed in the Hangzhou protocol.

The user creates an instance of ContractAbstraction using the `at` method of the Contract or Wallet API. The `contractViews` member of the `ContractAbstraction` instance is dynamically populated with methods that match the on-chain view names. _`contractViews` is an object where the key is the view name, and the value is a function that returns an instance of the OnChainView class._
The view parameter is expected in the JS object format and not as flattened arguments.
_Note for reference, the flattened arguments are the expected format when calling a contract entrypoint using the methods member, but we plan to move away from this format and favor the object one (see the methodsObject vs methods members of the ContractAbstraction)._
A method named `getSignature` on the `OnChainView` class allows inspecting the parameter and the returned type of the view. The `executeView` method of the `OnChainView` class allows simulating the view. It takes as a parameter a viewCaller which represents the contract address which is the caller of the view and an optional source which is the public key hash of the account that initialized this view execution.

Here is an example where a contract script contains an on-chain view called fib:
```
const contract = await Tezos.contract.at('KT1...');
contract.contractViews.fib().getSignature();
// getSignature returns: { parameter: 'nat', result: 'nat' }
await contract.contractViews.fib(10).executeView({ viewCaller: 'KT1... });
// executeView returns the BigNumber 55
```

The implementation to execute the view can be very similar to the Michelson view from the tzip16 package.

**Note: on-chain view should not be confused with the tzip4 views which are represented by ContractAbstraction.views.**

roxaneletourneau:
Closes #1117 

## Release Note Draft Snippet

### DIFF
--- a/docs/on_chain_views.md
+++ b/docs/on_chain_views.md
@@ -130,7 +130,7 @@ Tezos.contract.at(contractCallFib)
               .then(() => {
                 return contract.storage()
                   .then((finalStorage) => {
-                    println(`The storage is now ${finalStorage} which correspond to the value of the Fibonacci sequence at position ${fibPosition}.`);
+                    println(`The storage is now ${finalStorage} which corresponds to the value of the Fibonacci sequence at position ${fibPosition}.`);
                   })
               })
           })
@@ -177,4 +177,4 @@ Tezos.contract.at(contractTopLevelViews)
 On-chain views should not be confused with lambda views which are also available on the ContractAbstraction class. See the documentation for [lambda_view](lambda_view.md).
 :::
 
-Follow this link for more information about on-chain views: https://tezos.gitlab.io/011/michelson.html#michelsonviews-011
+Follow this link for more information about on-chain views: https://tezos.gitlab.io/active/michelson.html#operations-on-views
--- a/docs/on_chain_views.md
+++ b/docs/on_chain_views.md
@@ -1,0 +1,179 @@
+---
+title: On-chain views
+author: Roxane Letourneau
+---
+
+## On-chain views description
+
+- The `On-chain views` feature was introduced in the Hangzhou protocol. 
+- Views are meant to be called by a contract. It can be the contract where the view is defined or another contract.
+- Views take arguments as input and may depend on the contract's storage declaring the view.
+- Views are read-only and won't modify the contract's storage where the view is defined.
+- Views return a result as output which is immediately available on the stack of the caller contract.
+- The types `ticket`, `operation`, `big_map`, and `sapling_state` are not allowed in the argument or return types of views.
+
+## Declaring an on-chain view
+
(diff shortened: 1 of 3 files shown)


docs/on_chain_views.md
```diff
@@ -0,0 +1,179 @@
...
+       VIEW "fib" nat;
+       IF_SOME {NIL operation ; PAIR;} { FAIL }
+     }
+```
+
+**Example of calling a contract entrypoint that makes a call to a view using Taquito:**
+
+The following live code example shows a contract (`contractCallFib`) calling the view `fib` of another contract (`contractTopLevelViews`). 
+
+The example first shows the initial storage of the contract `contractCallFib`. It calls the default entry point of `contractCallFib` with the value of its storage + 1 and the address of the contract `contractTopLevelViews`. A call is made the `fib` view of `contractTopLevelViews` with the `storage + 1` as argument. The view returns the value of the Fibonacci sequence at the position represented by `storage + 1`. The storage of `contractCallFib` is updated to the result of the view.
```
michaelkernaghan, line 112: "A call is made the "  ->  "A call is made to the .."

docs/on_chain_views.md
```diff
@@ -0,0 +1,179 @@
...
+  })
+  .catch((error) => println(`Error: ${JSON.stringify(error, null, 2)}`));
+```
+
+## How to simulate a view execution using Taquito
+
+Taquito offers the ability to simulate the result of on-chain views. 
+
+The user can create an instance of `ContractAbstraction` using the `at` method of the Contract or Wallet API with the contract's address that defines the views. The `contractViews` member of the `ContractAbstraction` instance is dynamically populated with methods that match the on-chain view names. 
+
```
michaelkernaghan, line 146: " view arguments in parameter and "...    

should it be "the parameter",  ?

michaelkernaghan (commented):
very nice. just one, maybe two very minor typos.
There is an article about the views,  "https://medium.com/@keefertaylor/introducing-views-in-the-harbinger-price-oracle-aeee793e4577" in which they say views help to avoid the need to use callbacks in contracts. Could we mention something about that? or will readers already know about it?

docs/on_chain_views.md
roxaneletourneau: Fixed, thanks!

docs/on_chain_views.md
roxaneletourneau: I think my sentence is not really clear :P

It is better:
*`contractViews` is an object where the key is the view name, and the value is a function that takes the view arguments as a parameter and returns an instance of `OnChainView` class.*

roxaneletourneau:
> views help to avoid the need to use callbacks in contracts

thanks, @michaelkernaghan, I added a note about it.

Issue #1215, opened 2021-11-25
Activate rules for eslint


roxaneletourneau:
Many rules are disabled for ESLint, which should be enabled (for example, no-explicit-any).
See: https://github.com/ecadlabs/taquito/blob/master/.eslintrc.json#L16

roxaneletourneau:
I started inital work to enable the rules.
The following rules still need to be enable and required code fixes:

- [ ] "@typescript-eslint/no-unused-vars"

- [ ] "@typescript-eslint/no-explicit-any"

- [ ] "@typescript-eslint/ban-types"

- [ ] "@typescript-eslint/no-non-null-assertion"

- [ ] "@typescript-eslint/no-var-requires"

PR #1301, opened 2022-01-11
style(eslint lint rules): enabled @typescript-eslint/ban-types rule

roxaneletourneau:
re #1215

## Release Note Draft Snippet

### DIFF
--- a/packages/taquito/src/signer/interface.ts
+++ b/packages/taquito/src/signer/interface.ts
@@ -8,7 +8,7 @@ export interface Signer {
    * @param magicByte Magic bytes 1 for block, 2 for endorsement, 3 for generic, 5 for the PACK format of michelson
    */
   sign(
-    op: {},
+    op: string,
     magicByte?: Uint8Array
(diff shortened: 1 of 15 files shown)


packages/taquito/src/signer/interface.ts
```diff
@@ -8,7 +8,7 @@ export interface Signer {
    * @param magicByte Magic bytes 1 for block, 2 for endorsement, 3 for generic, 5 for the PACK format of michelson
    */
   sign(
-    op: {},
+    op: string,
```
dsawali, line 11: this changes `op` type from `object` to `string`

Is that intended?

roxaneletourneau: Yes, if we look at the classes that implement the Signer interface, each of them has the type string for this property. It makes sense because `op` is the forged bytes (a string). 
I am not too sure why the interface was defined to `{}`.

dsawali (approved):

PR #1310, opened 2022-01-12
Update to the README for the integration tests

### LINKED ISSUE #1306: update the instructions on how to use taquito integration tests
Review and refresh/update the instructions on running the Taquito integration tests: 
 https://github.com/ecadlabs/taquito/blob/master/integration-tests/README.md

We have a number of environment variables that are available, that could be documented in that readme too.

Validate that the instructions, when followed rote, on a fresh checkout, actually work.
Yes, update from delphi to hangzhounet
How to run the tests with a faucet key, instead of using the keygen api
How to run a single specific test
How to run against a specific RPC URL
How to run against a specific testnet
All those latter items, IIRC, are controlled by environment variables.

michaelkernaghan:
CLOSE #1306

Update to the README for the integration test

light editing, clarification of some points,
addition of other ways to modify the test run

### DIFF
--- a/integration-tests/README.md
+++ b/integration-tests/README.md
@@ -3,7 +3,7 @@
 The `taquito/integration-tests` directory contains Taquito's integration test suite. These tests exercise many use cases of Taquito against a live Tezos testnet.
 
 ## How to run the tests
-
+ 
 To run the tests, you must have:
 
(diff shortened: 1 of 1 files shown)


jevonearth (approved):
TY!

PR #1315, opened 2022-01-13
Mk 1294 deprecated sub instruction

### LINKED ISSUE #1294: Replace deprecated instruction SUB in test
The contracts used in the integration tests `contract-call-with-xtz-amount.spec.ts`, `wallet-unit-as-param.spec.ts` and `contract-unit-as-param.spec.ts` use the deprecated instruction `SUB`.
A fix is needed for these contract on ithacanet.

> The SUB instruction on type mutez is deprecated. It can be replaced by SUB_MUTEZ; ASSERT_SOME (and SUB; DROP can be replaced by ASSERT_CMPGE).

**Additional context**
https://tezos.gitlab.io/protocols/012_ithaca.html#michelson

michaelkernaghan:
Close #1294 

revised contract and updated tests to use SUB_MUTEZ in place of SUB in mutex type contexts.


roxaneletourneau (commented):
Could we skip the test and put the following config back :` const forgers: ForgerType[] = [ForgerType.COMPOSITE]` until the localForger supports the new `SUB_MUTEZ`  instruction?

michaelkernaghan:
> Could we skip the test and put the following config back :` const forgers: ForgerType[] = [ForgerType.COMPOSITE]` until the localForger supports the new `SUB_MUTEZ` instruction?

restored it and I made a way to ignore the affected tests with easy change back when Forging is ready

roxaneletourneau (approved):

PR #1331, opened 2022-01-17
changes for new op codes - support SUB_MUTEZ in composite forger

### LINKED ISSUE #1292: Support for the new instruction in the local-forging package
**Acceptance criteria:**
- Add support for the new SUB_MUTEZ instruction and its corresponding opcode in the local-forging.
- Add related unit test.
- Re-activate the following integration tests: 
  - contract-call-with-xtz-amount.spec.ts, 
  - wallet-unit-as-param.spec.ts and 
  - contract-unit-as-param.spec.ts

**Additional context**
https://tezos.gitlab.io/protocols/012_ithaca.html#michelson

michaelkernaghan:
CLOSE #1292

### DIFF
--- a/integration-tests/contract-call-with-xtz-amount.spec.ts
+++ b/integration-tests/contract-call-with-xtz-amount.spec.ts
@@ -62,13 +62,15 @@ CONFIGS().forEach(({ lib, rpc, setup, protocol }) => {
     skipHangzhouAndIthaca(
       //restore to skipHangzhou when forger supports new sub_mutez for Ithaca
       'originates a contract on Ithaca with SUB MUTEZ and sends base layer tokens when calling contract methods',
+     
       async (done) => {
+        try {
         const op = await Tezos.contract.originate({
           balance: '0',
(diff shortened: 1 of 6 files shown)


integration-tests/contract-call-with-xtz-amount.spec.ts
```diff
@@ -84,7 +86,9 @@ CONFIGS().forEach(({ lib, rpc, setup, protocol }) => {
         balance = await Tezos.tz.getBalance(contract.address);
         expect(balance.toString()).toEqual('1000001');
         done();
-      }
+      } catch (ex) {
+        console.error(ex);
```
roxaneletourneau, line 90: In case of an error, the test should fail instead of logging the error.

michaelkernaghan: sorry forgot to remove the added debugging console log line.  Removed.

roxaneletourneau (approved):

PR #1341, opened 2022-01-21
Update setup-node and cache GH actions to their latest versions and fix caching in pipelines

### LINKED ISSUE #1340: Current caching step is a bottleneck for most GH Action pipelines
**Description**
In most of our Github Action pipelines we're currently caching the entire `.npm` folder. This appears to be major bottleneck as the cache weight 2.2Gb and takes ~5m just to restore it.
Image from iOS

danielelisi:
Closes #1340 

# Description
The current `actions/cache` step takes ~5m to restore cache in most jobs due to the size of the cache ~2.2Gb. 
Image from iOS

Turns out it takes longer to update and restore `node_modules` cache than to install them in every job.

What we can cut down on is caching the compiled typescipt code in `packages/*/dist` and skip having to recompile it on every pipeline. We can detect changes in those files by comparing the hash of the ts files in the `src` folder of each package

This PR also updates Github Action `cache` step and `setup-node` to `v2`. These new version don't present any breaking changes.

# Test Plan

Previous to this change the `build-and-lint` job would take 13m. About 4m to restore cache and  ~3m for compiling typescript
Screenshot from 2022-01-21 13-17-40

After this change the `build-and-lint` job only takes 4m 
Screenshot from 2022-01-21 13-19-13


GImbrailo (approved):
Other than the discussed glob pattern the PR LGTM

danielelisi:
Turns out it takes longer to update and restore `node_modules` cache than to install them in every job.

What we can cut down on is caching the compiled typescipt code in `packages/*/dist` and skip having to recompile it on every pipeline. We can detect changes in those files by comparing the hash of the ts files in the `src` folder of each package

PR #1343, opened 2022-01-21
Add Types for Contract methods, storage, views, etc.

ricklove:
These changes are required to allow for improved type support:

See https://github.com/ecadlabs/taqueria/pull/91

### Before

Using taquito with the generated contract types:

The at method can be called providing a type with a utility method that can be provided:
```ts 
const contract = await Tezos.contract.at(`tz123`, contractAbstractionComposer<TestContractType>());

// methods can now use typed parameters
// methodsObject will be able to use type parameters
```

This can work the same with a wallet
```ts 
const contract = await Tezos.wallet.at(`tz123`, walletAbstractionComposer<TestContractType>());
```

Alternatively, this can be done with a cast:
```ts
const contract = await Tezos.contract.at(`tz123`) as ContractProviderFromContractType<TestContractType>;
```

The originate contract does not have any way to provide a type, so this requires a cast:
```ts
const originationResult = await Tezos.contract.originate({...});
const contract = await originationResult.contract(5) as ContractProviderFromContractType<TestContractType2>;
```

For accessing storage, there is no way to pass the type through the contract class, 
so it requires providing the type again:
```ts
const contract = await Tezos.contract.at(`tz123`) as ContractProviderFromContractType<TestContractType>;
const storage = await contract.storage<StorageFromContractType<TestContractType>>();
```

### After

Extending ContractAbstraction with additional generic types:

The at method can be called with the contract type provided:
```ts 
const contract = await Tezos.contract.at<TestContract>(`tz123`);

// methods can now use typed parameters
// methodsObject will be able to use type parameters
// storage will be able to use type parameters

```

This can work the same with a wallet
```ts 
const contract = await Tezos.wallet.at<TestWalletContract>(`tz123`);
```

The originate contract now accepts a type:
```ts
const originationResult = await Tezos.contract.originate({...});
const contract = await originationResult.contract<TestContract2>(5);
```

The contract type now also provides the default storage type:
```ts
const contract = await Tezos.contract.at<TestContract>(`tz123`);
const storage = await contract.storage();
```

---

## Release Note Draft Snippet


roxaneletourneau (commented):
These changes look good to me for the methods and views!
For the storage, should a generic type be added to the originate methods? 
```
const op = await Tezos.contract.originate<storageType>({
        code,
        storage: {// type used here}
      })
```

packages/taquito/src/contract/interface.ts
```diff
@@ -158,7 +158,7 @@ export interface ContractProvider extends StorageProvider {
    *
    * @param OriginationOperation Originate operation parameter
    */
-  originate(contract: OriginateParams): Promise<OriginationOperation>;
+  originate<TContract extends DefaultContractType = DefaultContractType>(contract: OriginateParams<ContractStorageType<TContract>>): Promise<OriginationOperation>;
```
roxaneletourneau, line 161: I think the same should be added here for origination with batch and wallet:
- https://github.com/ecadlabs/taquito/blob/master/packages/taquito/src/batch/rpc-batch-provider.ts#L115
- https://github.com/ecadlabs/taquito/blob/master/packages/taquito/src/wallet/wallet.ts#L80
- https://github.com/ecadlabs/taquito/blob/master/packages/taquito/src/wallet/wallet.ts#L183

ricklove:
Fixing the bug first

packages/taquito/src/contract/interface.ts
ricklove: batch operations cannot be typed because the method accepts an array of operations:

There is no way to require a specific origination operation type at a specific index of the array (as a typed array). Each item of the array must be any of the possible types.

However, a utility method could be provided for this - see the `example-usages.ts` in the related PR.

ricklove: Oh, I see how .withOrigination works --- looking at that now

ricklove: Add wallet `originate()` and `withOrigination()` types for contract and wallet `batch()`

ricklove:
Verified that the CI failure for `integration-tests-ithacanet` is actually occurring in the master branch (and not from these changes):

https://github.com/ecadlabs/taquito/pull/1370

roxaneletourneau (approved):
Looks good to me, great work :)

PR #1345, opened 2022-01-21
Limit GH Action workflows concurrency

### LINKED ISSUE #1342: Outdated GH Action workflows aren't cancelled when new commits are pushed to a PR
**Description**
While testing #1341 I noticed that old pipelines would keep running even if new commits are pushed to the branch. This behavior fills up GH runners queue with jobs we don't care about.

New commit to a branch/PR should be detected tocancel the outdated pipelines and start running pipelines based on the latest commits.

danielelisi:
Closes #1342 

# Description
When working on PR or when merging to `master` we should always make sure to run pipeline against the latest code.

Currently when we push new commits to a PR the new pipelines for the latest code are queued up and need to wait for previous pipelines to complete.

By limiting workflow concurrency old pipelines will be cancel automatically when a newer commit is pushed to a PR.

# Test Plan
Testing this changes to show that when new commits are pushed old pipelines are automatically cancelled.

Peek 2022-01-21 14-03

### DIFF
--- a/.github/workflows/algolia.yml
+++ b/.github/workflows/algolia.yml
@@ -3,6 +3,11 @@ name: Algolia indexing
 on:
   schedule:
     - cron: '0 0 * * *'
+
+concurrency:
+  group: algolia-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   delete-current-algolia-index:
     env:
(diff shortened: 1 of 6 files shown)


BGMurphy (approved):
Looks great!

PR #1349, opened 2022-01-21
Run Taquito integration tests from public GH runners

danielelisi:
# Description

We've been running Taquito integration tests with our self-hosted runners for some times now; however due to our current infrastructure we can't scale self-hosted runner on demand the way we would like to. Furthermore the operational cost for maintaining the runners is starting to put some weight on the SRE team.

As Github offers unlimited minutes when using runners for public repositories I suggest we switch all our CI/CD jobs to public runners. Since we make use of endpoint and services internal to our network I'm using the [tailscale/github-action](https://github.com/tailscale/github-action/actions) to authenticate the runner to our internal network. Worth noting that I configured Tailscale ACLs to restrict communication only to required services.  

# Test Plan
All pipelines succeeded :green_circle:


GImbrailo (approved):
LGTM

PR #1357, opened 2022-01-25
Mk 1353 update rpc node integration tests for ithaca

michaelkernaghan:
Update config and tests for change to ithaca2


integration-tests/config.ts
```diff
@@ -87,41 +87,41 @@ const hangzhounetEphemeral = {
...
+          "argue",
+          "knee",
+          "pluck",
+          "remove",
+          "scheme"
+        ],
+        "email": "noriqgjl.gtsyulgy@teztnets.xyz",
+        "password": "st3sZBRLWF",
+        "amount": "118887604096",
+        "activation_code": "7d414378d9071328313cca699d6922f1b59d076a"
```
roxaneletourneau, line 118: activation_code should be secret

integration-tests/config.ts
```diff
@@ -87,41 +87,41 @@ const hangzhounetEphemeral = {
...
     }
   },
 };
 
 const hangzhounetFaucet = {
-  rpc: process.env['TEZOS_RPC_HANGZHOUNET'] || 'https://hangzhounet.api.tez.ie',
+  rpc: process.env['TEZOS_RPC_HANGZHOUNET'] || 'https://hangzhounet.smartpy.io/',
```
roxaneletourneau, line 124: By curiosity, why is the rpc URL is changed to smartpy

example/deploy-docs-live-code-contracts.ts
```diff
@@ -99,41 +99,41 @@ async function checkBalances(users: string | any[]) {
 }
 
 async function originateTheContracts() {
-  await checkBalances(users);
+  //await checkBalances(users);
```
roxaneletourneau, line 102: Not sure if it is intended that the file is commented

packages/taquito/src/constants.ts
```diff
@@ -41,7 +42,8 @@ export const protocols = {
   '010': [Protocols.PtGRANADs],
   '011': [Protocols.PtHangz2], // hangzhou v2,
   '012': [Protocols.PsiThaCa], //Ithaca
-  '013': [Protocols.ProtoALpha],
+  '013': [Protocols.Psithaca2], //Ithaca2
+  '014': [Protocols.ProtoALpha],
 };
 
 export enum DefaultLambdaAddresses {
```
roxaneletourneau, line 49: A lambda contract and the chain_id for ithaca2 are missing in this file.

roxaneletourneau (changes requested):

integration-tests/config.ts
michaelkernaghan: a mistake! Thanks. fixed.
I'm not sure I changed it to smartpy though :)

example/deploy-docs-live-code-contracts.ts
michaelkernaghan: Fixed.

packages/taquito/src/constants.ts
michaelkernaghan: originated the Lambda contract and added the address here
added the chon_id

roxaneletourneau (approved):

Issue #1365, opened 2022-01-26
Github Workflow cache isn't being busted  


danielelisi:
**Description**
@roxaneletourneau reported recent failing workflows. The workflows are skipping the `build` step causing problems at later stages when trying to reference required libraries.

The `build` step is skipped whenever we find a valid cache of compiled Taquito libraries from `taquito/packages/**/dist`. This caching logic was implemented in #1341 to speed up workflow runs. We need to tweak this logic as it seems it isn't busting cache as I originally expected to.

Link to a failed Github workflow caused by this issue: https://github.com/ecadlabs/taquito/runs/4958740714?check_suite_focus=true

![Screenshot from 2022-01-26 15-05-09](https://user-images.githubusercontent.com/22307776/151261977-4d4b7524-75dc-4f0e-a0c5-d3b3dbde70ca.png)

danielelisi:
I'll remove the caching logic for now to fix the workflows. Further investigation and testing is required when we want to re-implement caching to speedup workflow runs

danielelisi:
Completed in https://github.com/ecadlabs/taquito/pull/1366

PR #1366, opened 2022-01-26
Remove Caching steps in Taquito workflows

danielelisi:
Ref #1365  

# Summary
Remove caching steps in all Taquito workflows due to the problem mentioned in the referenced Issue


danielelisi:
Workflows are now executing correctly. Merging this PR

Issue #1379, opened 2022-02-02
Wallet Originated Contract - not working with BeaconWallet signing of methods


ricklove:
**Description**
After originating a contract, the resulting contract is not usable (it seems to be missing some context or something).

**Steps To Reproduce**

Context: chrome browser using BeaconWallet (with temple extension)

```
const wallet = new BeaconWallet({
    name: "Example Dapp",
});
await wallet.requestPermissions({ 
    network,
});

const userAddress = await wallet.getPKH();
Tezos.setWalletProvider(wallet);


const origination = await Tezos.wallet.originate<ExampleWalletType>({
    code: ExampleCode.code,
    storage: tas.int(42),
}).send();

// ---- ISSUE HERE ----
// 1. Not working - origination.contract() -> calling methods will cause signing error
// const contract = await origination.contract();

// 2. Working - using Tezos.wallet.at again to get the wallet contract from the address
const contractAddress = (await origination.contract()).address;
const contract = await Tezos.wallet.at<ExampleWalletType>(contractAddress);
// ---- ^ ISSUE HERE ----

// Below will cause a signing error with method 1.
const sendResult = await contract.methodsObject.increment(tas.int(42)).send();
await sendResult.confirmation(5);
```

When using the first method to get the `contract()`. Calls using `methods` does not use the BeaconWallet signer, and instead the following error occurs:

`No signer has been configured. Please configure one by calling setProvider({signer}) on your TezosToolkit instance.`

**Expected behavior**

```
const contract = await origination.contract();
``` 

should have the same result as (i.e. it is possible to call .methods, etc. and it works with the BeaconWallet for signing):

```
const contractAddress = (await origination.contract()).address;
const contract = await Tezos.wallet.at<ExampleWalletType>(contractAddress);
```

**Screenshots**
![image](https://user-images.githubusercontent.com/703880/152185239-b7703c2b-1863-4676-bde4-b40f03a49562.png)

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: chrome
 - Version: latest
 
 - Taquito version: `11.2.0-beta-RC.0`

PR #1380, opened 2022-02-02
Context clone missing some fields

ricklove:
Fixes Bug: https://github.com/ecadlabs/taquito/issues/1379

---

## Release Note Draft Snippet

FIX Wallet Originated Contract methods usage should work as expected

### DIFF
--- a/packages/taquito/src/context.ts
+++ b/packages/taquito/src/context.ts
@@ -247,7 +247,10 @@ export class Context {
       this._config,
       this.forger,
       this._injector,
-      this.packer
+      this.packer,
+      this._walletProvider,
+      this._parser,
+      this._globalConstantsProvider,
     );
   }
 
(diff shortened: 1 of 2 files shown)


ricklove:
Tested this fix, and it does fix the original issue.

Question: is there any side effects (where clone is expected to ignore the last 3 items)?

ricklove:
@roxaneletourneau I tried to look for an integration test that covers it. 

The problem seems to be that none of the integration test configs use a cotnext that uses a wallet provider. I'm not sure the best way to create a test that would cover this case:

- Tezos.setWalletProvider
- wallet.originate
- use contract

But it seems that all the wallet integration tests should be using a config that uses this type of setup.

roxaneletourneau (approved):

roxaneletourneau:
@ricklove 
These 3 parameters should have been included in the clone method, thank you for the fix.

There are integration tests with the contract and others with the wallet API. However, wallet tests depend on the [LegacyWalletProvider](https://github.com/ecadlabs/taquito/blob/11.2.0/packages/taquito/src/wallet/legacy.ts#L11) which in fact fall back to a signer.

roxaneletourneau (approved):

PR #1386, opened 2022-02-03
add wallet examples to live code in docs

### LINKED ISSUE #493: Update all website examples to show contract and wallet API example variants.
@roxaneletourneau implemented this pattern on the https://tezostaquito.io/docs/originate doc page.

All code examples should be updated to show usage with wallet API and contract API

Related to: #486

michaelkernaghan:
Closes #493

### DIFF
--- a/docs/ledger_signer.md
+++ b/docs/ledger_signer.md
@@ -123,7 +123,7 @@ Tezos.contract
 </TabItem>
   <TabItem value="walletAPI">
 
-```js live noInline wallet
+```js
 const amount = 0.5;
 const address = 'tz1h3rQ8wBxFd8L9B3d7Jhaawu6Z568XU3xY';
 
--- a/docs/ledger_signer.md
+++ b/docs/ledger_signer.md
@@ -123,7 +123,7 @@ Tezos.contract
 </TabItem>
   <TabItem value="walletAPI">
 
-```js
+```js live noInline wallet
 const amount = 0.5;
 const address = 'tz1h3rQ8wBxFd8L9B3d7Jhaawu6Z568XU3xY';
 
--- a/docs/making_transfers.md
+++ b/docs/making_transfers.md
@@ -67,24 +67,12 @@ fetch('https://api.tez.ie/keys/hangzhounet/', {
 // import { TezosToolkit } from '@taquito/taquito';
 // const Tezos = new TezosToolkit('https://hangzhounet.api.tez.ie');
 
-render(`Fetching a private key...`);
-fetch('https://api.tez.ie/keys/hangzhounet/', {
-  method: 'POST',
-  headers: { Authorization: 'Bearer taquito-example' },
-})
-  .then((response) => response.text())
-  .then((privateKey) => {
-    render(`Importing the private key...`);
-    return importKey(Tezos, privateKey);
-  })
-  .then(() => {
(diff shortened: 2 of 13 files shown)


docs/on_chain_views.md
```diff
@@ -138,6 +148,37 @@ Tezos.contract.at(contractCallFib)
   })
   .catch((error) => println(`Error: ${JSON.stringify(error, null, 2)}`));
 ```
+</TabItem>
+  <TabItem value="walletAPI">
+
+```js live noInline
```
roxaneletourneau, line 154: In order to use the BeaconWallet, the live examples need a `wallet` props: `js live noInline wallet` instead of  `js live noInline`. Otherwise, the LegacyWallet is used which depends on a signer.

michaelkernaghan: i have fixed all of these and checked that they call the wallet when run.

the samples for "estimate" did not work ith the js live online wallet - this needs to be investigated

michaelkernaghan: put the estimate example back to how it was. It won't work with a wallet as it is calling a contract

docs/ledger_signer.md
```diff
@@ -123,7 +123,7 @@ Tezos.contract
 </TabItem>
   <TabItem value="walletAPI">
 
-```js
+```js live noInline wallet
```
roxaneletourneau, line 126: The wallet props refer to BeaconWallet, but in the current example, we want to demonstrate the ledger signer. It might be more appropriate to put back ```js for this example.

docs/making_transfers.md
```diff
@@ -50,6 +60,40 @@ fetch('https://api.tez.ie/keys/hangzhounet/', {
...
 
+</TabItem>
+  <TabItem value="walletAPI">
+
+```js live noInline wallet
+// import { TezosToolkit } from '@taquito/taquito';
+// const Tezos = new TezosToolkit('https://hangzhounet.api.tez.ie');
+
+render(`Fetching a private key...`);
+fetch('https://api.tez.ie/keys/hangzhounet/', {
```
roxaneletourneau, line 71: No need to fetch a private key when using the wallet

```
// import { TezosToolkit } from '@taquito/taquito';
// const Tezos = new TezosToolkit('https://hangzhounet.api.tez.ie');

    const amount = 0.5;
    const address = 'tz1h3rQ8wBxFd8L9B3d7Jhaawu6Z568XU3xY';

    render(`Transfering ${amount} ꜩ to ${address}...`);
    return Tezos.wallet.transfer({ to: address, amount: amount }).send()
.then((op) => {
    render(`Waiting for ${op.opHash} to be confirmed...`);
    return op.confirmation(1).then(() => op.opHash);
  })
  .then((hash) => render(`Operation injected: https://hangzhou.tzstats.com/${hash}`))
  .catch((error) => render(`Error: ${JSON.stringify(error, null, 2)}`));
```

roxaneletourneau (changes requested):

docs/ledger_signer.md
michaelkernaghan: restored.

docs/making_transfers.md
michaelkernaghan: changed

PR #1394, opened 2022-02-08
Liquidity baking doc and permit testing

### LINKED ISSUE #1095: Rewrite the contract-permit integration to include Claude's findings about permit contracts.
**Description**
The original contract-permit test was derived from examples that use a trick to get the key that will be used for the permit. As Claude pointed out however this trick would not be acceptable in production environments and so he wrote an example permit implementation that does not require the trick.

we should rewrite the permit integration to work without the trick of inferring the key.

This issues also includes Liquidity Baking documentation

claudebarde:
This pull request would add 1 new page to the documentation about the liquidity baking contract and would update the integration tests with the necessary code for permits.

CLOSE #1095


integration-tests/contract-permits.spec.ts
```diff
@@ -13,6 +14,73 @@ const errors_to_missigned_bytes = (errors: any[]) => {
...
+          { prim: "address" }
+        ]
+      },
+      {
+        prim: "pair",
+        args: [{ prim: "nat" }, { prim: "bytes" }]
+      }
+    ]
+  };
+  // const sigParamPacked = await Tezos.rpc.packData({
```
michaelkernaghan, line 73: can we remove the commented-out code?

michaelkernaghan (changes requested):
Looks good

maybe remove some commented out code

Issue #1113, opened 2021-10-14
Use of the deprecated Buffer() by one taquito's dependencies


callistonianembrace:
**Description**
Scripts using taquito display the following warning

`(node:14846) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues.`

On inspecting further, it turns out Taquito uses xhr2-cookies which seems to use `new Buffer()` in [multiple places](https://github.com/souldreamer/xhr2-cookies/search?q=new+buffer). As per official recommendations, it is advisable to move away from `new Buffer()` to it's alternatives. See https://nodejs.org/api/buffer.html#buffer_new_buffer_array.

**Steps To Reproduce**
Firing any API using `taquito-http-utils` 

**Expected behavior**
Not trigger the Nodejs security warning.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser N/A
 - Version 10.2.1

**Smartphone (please complete the following information):**
N/A

**Additional context**
N/A

PR #1396, opened 2022-02-08
1113 fix dependency

dsawali:
Fixes: #1113 

## Release Note Draft Snippet


packages/taquito-http-utils/src/taquito-http-utils.ts
```diff
@@ -83,77 +79,56 @@ export class HttpBackend {
...
+        method: method ?? 'GET',
+        headers: headers,
+        responseType: resType,
+        transformResponse,
+        timeout: timeout,
+        data: data,
+      });
+    } catch (err: any) {
+      throw new HttpResponseError(
+        `Http error response: (${err.response.status}) ${JSON.stringify(err.response.data)}`,
```
roxaneletourneau, line 124: On error, it seems that `err.response.data` is sometimes a string, sometimes an object. I see that a change was needed in the following test file: [integration-tests/contract-call-with-xtz-amount.spec.ts](https://github.com/ecadlabs/taquito/pull/1396/files#diff-89b7d3fb841cede12ae4c302aad898f559ab136b37ff7f3ff3467e1c2c863f40). However, `err.response.data` is a string, if not doing JSON.stringify(err.response.data), we keep the same error format as before.

The solution I suggest is to do a validation on the type, and only call `JSON.stringify` if typeof `err.response.data === 'object'`

packages/taquito-http-utils/src/taquito-http-utils.ts
```diff
@@ -83,77 +79,56 @@ export class HttpBackend {
...
+        url: url + this.serialize(query),
+        method: method ?? 'GET',
+        headers: headers,
+        responseType: resType,
+        transformResponse,
+        timeout: timeout,
+        data: data,
+      });
+    } catch (err: any) {
+      throw new HttpResponseError(
```
roxaneletourneau, line 123: Not all errors will have a `response` property. It is the case of a time-out error for example.

I suggest doing something like:
```
catch (err){
if (err.response){
throw new HttpResponseError ...
} else if (// error is a timeout) {
throw error saying `Request timed out after: ${request.timeout}ms`
} else {
throw HttpRequestFailed
}
```

roxaneletourneau (changes requested):

roxaneletourneau (approved):

Issue #1389, opened 2022-02-04
USER STORY: Read interface


Innkst:
**As Taquito user, I want to be able to use Indexers instead of RPC, so that I get results faster and reduce number of RPC interactions**

**Use cases:**
DApp developers want to read data from the Indexer, i.e. tzkt or TezGraph

**Is this user story related to an existing problem? Please describe.**
A clear and concise description of what the problem is.

**Acceptance criteria:**
- Add the ability to substitute the usage of RCPClient with the ability to read data from the Indexer.

**Non-functional:**
- consider performance, scalability, security

OUT OF SCOPE:
N/A

**Additional context**

PR #1397, opened 2022-02-08
Read Interface

roxaneletourneau:
#1389 

When using Taquito, all data retrieved from the blockchain are obtained using an RPC by default. For example, all required data for preparing an operation are fetched by doing different queries to the RPC node. Those data could be obtained using a different medium (an Indexer), which would reduce the load on the nodes. With this in mind, we defined a new interface in Taquito named `TzReadProvider` and a new provider on the `TezosToolkit` named `readProvider`. The `readProvider` default to the RPC as before. The goal would be to have the different indexers implement the `TzReadProvider` interface allowing users to configure their TezosToolkit to fetch data from indexers instead of RPC if needed.

Another change has been done regarding the confirmation method of the operations. The confirmation is now done using the SubscribeProvider set on the TezosToolkit. By default, the SubscribeProvider is set to an instance of PollingSubscribeProvider, which polls on the RPC for the head blocks as before. This change is intended to make it easier to use a different strategy for operation confirmation (for example, it could use streaming based on an indexer instead of polling on the RPC head block)

This change to the confirmation methods, unfortunately, includes a breaking change. The polling interval for the confirmation and the streamer configuration has been moved from the TezosToolkit to the PollingSubscribeProvider class, where they belong logically.
BREAKING CHANGE: The confirmationPollingIntervalSecond and the ConfigStreamer are removed from the
TezosToolkit. Configuration for the PollingSubscribeProvider needs to be passed in its constructor.:
Before:
```
Tezos.setProvider({ config: { confirmationPollingIntervalSecond: 5 }});
```
Now:
```
Tezos.setStreamProvider(Tezos.getFactory(PollingSubscribeProvider)({ pollingIntervalMilliseconds:5000 }));
```

These changes are preliminary work to better support indexers in Taquito; more to come on this in the near future.


packages/taquito/src/context.ts
```diff
@@ -206,24 +219,9 @@ export class Context {
...
-          Math.max(0, constants.initial_endorsers! - constants.endorsers_per_block)
-        )
-      );
-
-      // Divide the polling interval by a constant 3
-      // to improvise for polling time to work in prod,
-      // testnet and sandbox enviornment.
-      confirmationPollingInterval = confirmationPollingInterval.dividedBy(3);
+      const constants = await this.readProvider.getProtocolConstants('head');
+      const blockTime = constants.minimal_block_delay ? constants.minimal_block_delay : constants.time_between_blocks ? constants.time_between_blocks[0] : new BigNumber(defaultInterval);
```
dsawali, line 223: can I get a brief explanation on what blockTime is used for?

packages/taquito/src/read-provider/rpc-read-adapter.ts
```diff
@@ -0,0 +1,179 @@
+import { EntrypointsResponse, MichelsonV1Expression, MichelsonV1ExpressionExtended, SaplingDiffResponse, ScriptedContracts } from "@taquito/rpc";
+import BigNumber from 'bignumber.js';
+import { Context } from "../context";
+import { BigMapQuery, BlockIdentifier, SaplingStateQuery, TzReadProvider } from "./interface";
+
+/**
+     * @description Converts calls from TzReadProvider into calls to the wrapped RpcClient in a format it can understand.
```
dsawali, line 7: minor nitpick, but the description comment block seems misaligned here

packages/taquito/src/context.ts
roxaneletourneau: It is used to calculate the polling interval used during the operation confirmation

dsawali (approved):

PR #1409, opened 2022-02-15
Reduce integration test transaction costs

### LINKED ISSUE #1405: Chore: reduce the amount of tez used in transactions by the integration tests
There are occasions where a third party would like to use Taquito integration tests and may do so on Mainnet where there are actual costs. therefore the transaction costs in the integration tests should be minimized.

Revise all integration transactions to be on the order of 0.01 or 0.02 tez where possible without breaking existing tests.

michaelkernaghan:
CLOSE #1405


integration-tests/contract-batch.spec.ts
```diff
@@ -5,18 +5,17 @@ import { MANAGER_LAMBDA, OpKind } from '@taquito/taquito';
...
 CONFIGS().forEach(({ lib, rpc, setup, knownBaker, knownContract, createAddress }) => {
     const Tezos = lib;
-    const test = require('jest-retries');
     describe(`Test contract.batch using: ${rpc}`, () => {
         beforeEach(async (done) => {
             await setup();
             done();
         });
-        test('Simple transfers with origination (where the code in JSON Michelson format)', 2, async (done: () => void) => {
+        test('Simple transfers with origination (where the code in JSON Michelson format)', async (done: () => void) => {
```
roxaneletourneau, line 13: I think that `: () => void` was added for jest-retries and can be removed

`async (done)` instead of `async (done: () => void)`

michaelkernaghan: I have made this fix in several places where it is needed.

michaelkernaghan:
I redid one of the tests after we lowered the amount that keygen gives to new addresses.

Issue #1298, opened 2022-01-10
Use the RPC run_view to execute lambda views


roxaneletourneau:
A new RPC endpoint `helpers/scripts/run_view` was introduced in the Hangzhou protocol. It allows simulating a call to a view following the TZIP-4 standard.

Currently, Taquito simulates those views by calling a lambda contract that always fails. The result of invoking the always-failing lambda contract is an error from the blockchain which contains the requested information.

**Describe the solution you'd like**
Support the new `helpers/scripts/run_view` endpoint in the RpcClient class.
Use the `helpers/scripts/run_view` endpoint to simulate the tzip-4 views instead of the lambda contract.
This change will remove the need of deploying the lambda contract on the different testnets or having to deploy it when using a sandbox.

**Additional context**
[Add any other context or screenshots about the feature request here.](https://gitlab.com/tezos/tezos/-/blob/master/docs/api/hangzhou-openapi.json)

PR #1412, opened 2022-02-18
1298 implement run_view support in Taquito

dsawali:
#1298 

## Release Note Draft Snippet

### DIFF
--- a/docs/rpc_package.md
+++ b/docs/rpc_package.md
@@ -96,8 +96,9 @@ const packedData = await client.packData({ data: { string: 'test' }, type: { pri
 console.log('-- Packed data:', packedData);
 ```
 
-You can simulate the `PACK` instruction from Michelson with the `packData` method.
+You can simulate the `PACK` instruction from Michelson with the `packData` method.   
 
+This function will execute Tzip4 views normally referred to as 'Lambda Views'. You can learn more about Tzip4 [here](https://gitlab.com/tezos/tzip/-/blob/master/proposals/tzip-4/tzip-4.md#view-entrypoints)
 ```js
 // runs view
 const view = await client.runView({
--- a/integration-tests/rpc-nodes.spec.ts
+++ b/integration-tests/rpc-nodes.spec.ts
@@ -86,7 +86,7 @@ CONFIGS().forEach(
           done();
         });
 
-        it('runView', async (done) => {
+        it('Executes tzip4 views by calling runView ', async (done) => {
           let chainId: string;
 
           if (protocol === Protocols.Psithaca2) {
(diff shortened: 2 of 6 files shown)


docs/rpc_package.md
```diff
@@ -98,6 +98,18 @@ console.log('-- Packed data:', packedData);
 
 You can simulate the `PACK` instruction from Michelson with the `packData` method.
 
+```js
+// runs view
```
roxaneletourneau, line 102: I suggest adding a note that this function is the execute tzip4 views (refer as lambda views) and maybe add a link to the tzip4: https://gitlab.com/tezos/tzip/-/blob/master/proposals/tzip-4/tzip-4.md#view-entrypoints

To avoid confusing between all kind of views (on-chain views, off-chain views, lambda-views)

integration-tests/rpc-nodes.spec.ts
```diff
@@ -86,6 +86,30 @@ CONFIGS().forEach(
...
+          let chainId: string;
+
+          if (protocol === Protocols.Psithaca2) {
+            chainId = ChainIds.ITHACANET2
+          } else {
+            chainId = ChainIds.HANGZHOUNET
+          }
+
+          const params: RPCRunViewParam = {
+            contract: 'KT1LNMrk8orMQ85zbwK25996dPhDxfSicvKh',
```
roxaneletourneau, line 99: The contract is deployed on hangzhounet, so the test will fail on ithacanet. I suggest using the knownBigMapContract from the config file which has a getBalance entrypoint too.

packages/taquito-rpc/test/taquito-rpc.spec.ts
```diff
@@ -2562,6 +2563,23 @@ describe('RpcClient test', () => {
     });
   });
 
+  describe('runView', () => {
+    it('query the right url and data', async (done) => {
+      const testData = {
+        unparsing_mode: "Readable"
+      };
+
+      await client.runView(testData as any);
```
roxaneletourneau, line 2572: Can you use valid param in the test, `testData` is missing all the required properties. That way it won't need to be casted as any.

roxaneletourneau: A test also showing that we can override the default unparsing_mode can be great also.

packages/taquito-rpc/src/types.ts
```diff
@@ -954,6 +954,21 @@ export type RunCodeResult = {
...
 
+export type RPCRunViewParam = {
+  contract: string;
+  entrypoint: string;
+  input: MichelsonV1Expression;
+  chain_id: string;
+  source?: string;
+  payer?: string;
+  gas?: BigNumber;
+  unparsing_mode?: UnparsingModeEnum;
```
roxaneletourneau, line 965: "unparsing_mode" is required by the RPC if I am not mistaken. If so, I think we should make this property required in the type.

packages/taquito-rpc/src/rpc-client-modules/rpc-cache.ts
```diff
@@ -798,6 +800,17 @@ export class RpcClientCache implements RpcClientInterface {
...
   }
 
+  /**
+   * @param viewParams Parameters of the view to run
+   * @param options contains generic configuration for rpc calls
+   *
+   * @description Simulate a call to a view following the TZIP-4 standard. See https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-4/tzip-4.md#view-entrypoints.
+   *
+   */
+   async runView(viewParams: RPCRunViewParam, { block }: RPCOptions = defaultRPCOptions): Promise<RunViewResult> {
```
roxaneletourneau, line 810: consider adding a default value for unparsing_mode as in the RpcClient class. 
same for the rpc-wrapper

roxaneletourneau (changes requested):

packages/taquito-rpc/test/taquito-rpc.spec.ts
dsawali: will amend that!

integration-tests/rpc-nodes.spec.ts
dsawali: good catch, will do!

roxaneletourneau (approved):

PR #1417, opened 2022-02-23
Beaconwallet singleton

claudebarde:
Added new documentation about the BeaconWallet singleton


roxaneletourneau (changes requested):
Documentation file is missing

roxaneletourneau (approved):
Looks good to me :)

docs/beaconwallet-singleton.md
```diff
@@ -0,0 +1,289 @@
...
+
+In modern JavaScript frameworks, this can be achieved by passing the instance to the components through their props or by keeping the instance in the context (or state) of the dapp so that every component has access to the same instance.  
+
+This is how this can be accomplished with 3 of the main JavaScript frameworks, [React](https://reactjs.org/), [Vue](https://vuejs.org/) and [Svelte](https://svelte.dev/).
+
+## With React
+### - Using "prop drilling"
+It is possible to create a single instance of the `BeaconWallet` in one of the parent components and pass it down to its children component. This method can be preferred when there aren't many components down the tree. It becomes difficult to track when the instance is passed to many components and their props become redundant.
```
michaelkernaghan, line 18: children -> child?

docs/beaconwallet-singleton.md
```diff
@@ -0,0 +1,289 @@
...
+        if(context.wallet.getActiveAccount()){
+            ...
+        }
+    }, [])
+}
+```
+
+## With Vue
+### - Passing the instance through props
+It is possible to do "prop drilling" in Vue and pass an instance of the `BeaconWallet` down the children of a component. This method requires updating the value returned by the `data` method and catching the prop in the `props` array of the child component:
```
michaelkernaghan, line 91: down -> down to?

docs/beaconwallet-singleton.md
```diff
@@ -0,0 +1,289 @@
...
+            ...
+        }
+    })
+</script>
+
+<div>Something</div>
+```
+
+### - Using a store
+A Svelte store is the recommended way to store and use an instance of the `BeaconWallet`. You just need to import the store in the components that need access to the instance and you can be sure you have a single instance:
```
michaelkernaghan, line 248: in the?   not sure i understand the sentence here

docs/beaconwallet-singleton.md
michaelkernaghan: or child compenents?

claudebarde: Done

docs/beaconwallet-singleton.md
claudebarde: Which sentence don't you understand?

docs/tezos_domains.md
```diff
@@ -18,7 +18,7 @@ If you have an address and you want to find the domain tied to it, the storage o
 import { TezosToolkit } from "@taquito/taquito";
 import { bytes2Char } from "@taquito/utils";
 
-const domainContractAddress = "KT1PfBfkfUuvQRN8zuCAyp5MHjNrQqgevS9p";
+const domainContractAddress = "KT1GBZmSxmnKJXGMdMLbugPfLyUPmuLSMwKS";
```
roxaneletourneau, line 21: As the domainContractAddress is deployed on mainnet, could you replace the RPC URL passed in the TezosToolkit in each example to use mainnet instead of "https://testnet-tezos.giganode.io"?

docs/beaconwallet-singleton.md
michaelkernaghan: It's ok it reads fine now that I look at it again. Thanks;

PR #1422, opened 2022-02-28
feat(localforger): add protocol hash to the constructor of LocalForger

roxaneletourneau:
In preparation for Ithaca support, where some operation schema has changed (ie endorsement), the protocol hash will be passed to the constructor of the LocalForger in order to select the right encoder, decoder.

## Release Note Draft Snippet


packages/taquito-local-forging/src/taquito-local-forging.ts
```diff
@@ -25,10 +26,10 @@ export function getCodec(codec: CODEC) {
   };
 }
 
-export { VERSION } from './version';
-
 export class LocalForger implements Forger {
-  private codec = getCodec(CODEC.MANAGER);
+  constructor(public readonly protocolHash = ProtocolsHash.PtHangz2) {}
```
dsawali, line 30: will this be something we'd change in the future in accordance to protocol updates?

roxaneletourneau: Yes, we will need to update the default value to match new protocol hash

dsawali: maybe passing a config that is named `currentProtocol` or something like that would be more descriptive, and instead we can just modify the value of `currentProtocol` and not pass them directly to the constructor

roxaneletourneau: I thought about it a little more, and I think that I will change my precedent response. The schema of endorsement is different in hangzhou and ithaca. So at the moment, we need to know which protocol is used to forge this kind of operation properly. The distinction is needed until the migration from hangzhou to ithaca on mainnet. After that, it won't really make sense to forge using the old schema and we should remove the distinction to keep the code cleaner.

PR #1423, opened 2022-02-28
made changes to lambda view to use run_view endpoint

### LINKED ISSUE #1298: Use the RPC run_view to execute lambda views

dsawali:
- Modified lambda view implementation to use run_view
- Updated relevant documentation (and removed outdated ones)
- Added unit tests
- Removed unnecessary examples

Closes #1298 

## Release Note Draft Snippet


packages/taquito/test/contract/contractView.spec.ts
```diff
@@ -137,4 +139,23 @@ describe('ContractView test', () => {
     }
     done();
   });
+
+  it('Should be able to execute tzip4 views', async (done) => {
+    mockRpcClient.runView.mockResolvedValue({
+      "data": {"int": "100"}
+    });
+
+    const result = await mockRpcClient.runView({
```
roxaneletourneau, line 148: It seems that the test is more testing the mock behavior than the tzip4 execution itself.

Some suggestions:
The test should call the `read` method, which will internally call `mockRpcClient.runView`.
The test should ensure that the returned value is formatted properly (`"data": {"int": "100"} `will be transform to `100`)

Note that if no `chainId` is passed in the `read` method, a mock will be needed for  `mockRpcClient.getChainId`.
It might be great to have a test case where a `chainId `is passed in the `read` method and another where it is no.
The first test will call `mockRpcClient.getChainId` and the second will not.

_The advantage of accepting the chainId as a param is to reduce the number of calls made to the RPC_

roxaneletourneau (changes requested):

packages/taquito/test/contract/contractView.spec.ts
dsawali: that's a good point, will amend

roxaneletourneau (approved):
Looks good to me !

PR #1427, opened 2022-03-01
add tests for some sample sapling contracts 

michaelkernaghan:
There are contracts used in tests in Tezos. I have made tests here that originate those contracts in Tezos and then check features of the contracts.


integration-tests/contract-originate-contract-with-sapling.spec.ts
```diff
@@ -43,5 +61,132 @@ CONFIGS().forEach(({ lib, rpc, setup }) => {
...
+      const op = await Tezos.contract.originate({
+        code: saplingContractDouble,
+        init: `(Pair {} {})`,
+      });
+      await op.confirmation();
+      expect(op.hash).toBeDefined();
+      expect(op.includedInBlock).toBeLessThan(Number.POSITIVE_INFINITY);
+      const contract = await op.contract();
+
+      const saplingDiffById = await rpcClient.getSaplingDiffById('168');
```
roxaneletourneau, line 75: The sapling state id 168 is not related to the current contract, should this be removed from the test?

integration-tests/contract-originate-contract-with-sapling.spec.ts
```diff
@@ -43,5 +61,132 @@ CONFIGS().forEach(({ lib, rpc, setup }) => {
...
+      });
+
+      await op.confirmation();
+      expect(op.hash).toBeDefined();
+      expect(op.includedInBlock).toBeLessThan(Number.POSITIVE_INFINITY);
+
+      done();
+    });
+
+    hangzhounet('Should fail to originate a Push Sapling State contract', async (done) => {
```
roxaneletourneau, line 140: This test and the following one could be combined into one by matching the error message only to `michelson_v1.unexpected_lazy_storage`

integration-tests/contract-originate-contract-with-sapling.spec.ts
michaelkernaghan: removed.

integration-tests/contract-originate-contract-with-sapling.spec.ts
michaelkernaghan: changed to be one test. Thanks that is much better :)

roxaneletourneau (approved):

Issue #1300, opened 2022-01-10
[Chore] Add Git pre-commit hook to run linting and checks


dsawali:
Add a git pre-commit hook to ensure linting is done before a commit.

relevant documentation: https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks

PR #1428, opened 2022-03-02
added husky pre-commit hook to lint

dsawali:
#1300

Added husky pre-commit hook to lint before a commit
 
## Release Note Draft Snippet


.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npm run lint
```
roxaneletourneau, line 4: Should we run `npm run precommit` instead to apply prettier configuration too?
see https://github.com/ecadlabs/taquito/blob/master/packages/taquito-beacon-wallet/package.json#L37

.github/workflows/deploy_edge.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-node@v2
       with:
-        node-version: ${{ matrix.node }}
+        node-version: ${{ matrix.nodet }}
```
roxaneletourneau, line 26: Can you explain what this change is doing?

package.json
```diff
@@ -15,7 +15,8 @@
     "build-docs": "typedoc",
     "gh-pages": "gh-pages",
     "netlify:preview": "netlify deploy --site $NETLIFY_SITE_ID --auth $NETLIFY_AUTH_TOKEN --json --alias=$short_sha --message 'Deployment for $GITHUB_REF'",
-    "netlify:publish": "netlify deploy --site $NETLIFY_SITE_ID --auth $NETLIFY_AUTH_TOKEN --json --prod --message 'Deployment for $GITHUB_REF'"
+    "netlify:publish": "netlify deploy --site $NETLIFY_SITE_ID --auth $NETLIFY_AUTH_TOKEN --json --prod --message 'Deployment for $GITHUB_REF'",
+    "prepare": ""
```
roxaneletourneau, line 19: Is the prepare script needed if it is empty?

dsawali: it's needed so it doesn't get run in the CI, I was following this part of the docs https://typicode.github.io/husky/#/?id=disable-husky-in-cidockerprod

.github/workflows/deploy_edge.yml
dsawali: this might be a mistake, I will fix this

package.json
roxaneletourneau: got it, thanks!

.husky/pre-commit
dsawali: sure, that  sounds good. Do you think we should run both lint and precommit? Or will only precommit be suffecient?

roxaneletourneau: I think precommit will be sufficient as it will call `eslint --fix`

dsawali: gotcha, I'll remove the lint stage then

roxaneletourneau (approved):

Issue #1387, opened 2022-02-04
Origination of complex contracts fail


asbjornenge:
Hi there 😊

I'm having some trouble originating complex contracts using the wallet API.

I'm in a React app using `@taquito/taquiot@11.2.0` and `@taquito/beacon-wallet@11.2.0` and connecting Temple wallet.

The following code works fine if I use a very simple contract:

```
export async function originate(contract, storage) {
  const op = await Tezos.wallet.originate({
    code: contract,
    init: storage,
  }).send()
  await op.confirmation(1)
  const _contract = await op.contract()
  return _contract.address
}
```

Contract and storage are JSONs produced by a SmartPy compilation of this simple contract:

```
import smartpy as sp

class Simple(sp.Contract):

  def __init__(self, admin):
    self.init(
      admin = admin
    )

  @sp.entry_point
  def default(self):
    pass

  @sp.entry_point
  def setAdmin(self, admin):
    sp.set_type(admin, sp.TAddress)
    self.data.admin = admin
```

However, if I try to originate say a FA2 token contract using the exact same approach (compile via smartpy into jsons and importing them), I get this error:

```
[
  {
    "kind": "temporary",
    "id": "proto.011-PtHangz2.storage_exhausted.operation"
  }
]
```

NOTE: The contract originates fine using smartpy.

🤔🤔🤔

So it seems to me to be some kind of storage and / or gas fee calculation that is failing?

I have tried using the `toolkit.estimate` API but it complains that `No signer has been configured` 😅 And I cannot seem to find a way to set a signer using the BeaconWallet. I do call `.setWalletProvider` ofcourse, but it does not seem to set it...?

Any and all help would be greatly appreciated 🙌

thanks!

asbjornenge:
FYI: Using the exact same code, contracts and versions above, works totally fine in Node.js 😅

Only difference is I use `importKey` instead of `setWalletProvider` 🤔 I guess that sets the signer properly or something 😅

roxaneletourneau:
Hi @asbjornenge 
Did you try increasing the storage limit before confirming the operation with the Temple wallet:
![image](https://user-images.githubusercontent.com/65877347/157491979-948d48ec-b728-4867-948f-325470bfd41c.png)

It is currently only possible to estimate based on a signer. We made a fix so an estimate can be run based on a configured wallet too. It will be part of release 12.0.0.

asbjornenge:
Hi @roxaneletourneau :blush: 

[...]

PR #1438, opened 2022-03-04
feat(estimate): fix estimate to work with wallet

roxaneletourneau:
Estimation is done with the configured signer. If there is no configured signer, estimation is done
using the public key hash of the configured wallet. Taquito does not have access to the public key
of the wallet (only the pkh), so won't estimate reveal operation. If the account is unrevealed, an
exception is throwned.

re #1387

## Release Note Draft Snippet

### DIFF
--- a/packages/taquito/src/estimate/rpc-estimate-provider.ts
+++ b/packages/taquito/src/estimate/rpc-estimate-provider.ts
@@ -64,7 +64,7 @@ export class RPCEstimateProvider extends OperationEmitter implements EstimationP
   private readonly ORIGINATION_STORAGE = 257;
   private readonly OP_SIZE_REVEAL = 128;
 
-  private async getSignerOrWallet(): Promise<{
+  private async getKeys(): Promise<{
     publicKeyHash: string;
     publicKey?: string;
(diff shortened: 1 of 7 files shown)


packages/taquito/src/estimate/rpc-estimate-provider.ts
```diff
@@ -63,6 +64,19 @@ export class RPCEstimateProvider extends OperationEmitter implements EstimationP
   private readonly ORIGINATION_STORAGE = 257;
   private readonly OP_SIZE_REVEAL = 128;
 
+  private async getSignerOrWallet(): Promise<{
```
dsawali, line 67: I think this name might get misunderstood in contrast to what it actually returns. Maybe something a bit more general is fine since we don't necessarily have to expose why or how the pkh and the public key gets returned?

My suggestion would be something like `getKeys` or something in a similar light.

roxaneletourneau: Thanks for the suggestion. Indeed, getKeys is more descriptive, I will make the change.

PR #1444, opened 2022-03-07
Ithaca support for michel-codec

e-asphyx:

## Release Note Draft Snippet


packages/taquito-michel-codec/michelson.json
```diff
@@ -2853,187 +2728,258 @@
...
       "ty_args": "mutez",
       "related_ops": [
-        "ADD", "AMOUNT", "BALANCE", "CREATE_CONTRACT", "EDIV", "MUL", "SUB",
+        "ADD",
+        "AMOUNT",
+        "BALANCE",
+        "CREATE_CONTRACT",
+        "EDIV",
+        "MUL",
+        "SUB",
```
roxaneletourneau, line 2926: The SUB instruction on type mutez is deprecated

jevonearth: Even if it is deprecated, does that mean it should be forbidden here? I'm not sure it's our place to prevent people from using it?

roxaneletourneau: I could not find where this file is used in the code base, so I thought it was there for informative purposes only (Correct me if I am wrong please). I  pointed out the deprecated instruction so we keep the file up to date.

packages/taquito-michel-codec/test/typecheck.spec.ts
```diff
@@ -1,185 +1,194 @@
...
+      args: [{ prim: 'pair', args: [{ prim: 'string' }, { prim: 'nat' }] }],
+    };
+    const data: MichelsonData = [
+      { prim: 'Pair', args: [{ string: 'aaa' }, { int: '0' }] },
+      { prim: 'Pair', args: [{ string: 'bbb' }, { int: '1' }] },
+    ];
+    assertDataValid(data, typedef);
+  });
+
+  // it('assertDataValid: unsorted set', () => {
```
roxaneletourneau, line 89: Why this test and 'assertDataValid: unsorted map' are commented out?

packages/taquito-michel-codec/michelson.json
e-asphyx: It's deprecated, see
```
 const s = ProtoGreaterOfEqual(proto, Protocol.PsiThaCa)
          ? args(0, ['nat', 'int', 'timestamp'], ['nat', 'int', 'timestamp'])
          : args(0, ['nat', 'int', 'timestamp', 'mutez'], ['nat', 'int', 'timestamp', 'mutez']);
```

And `michelson.json` was left there at some point for reference but it's never used

packages/taquito-michel-codec/src/binary.ts
```diff
@@ -1,1188 +1,1403 @@
...
+} from './utils';
+
+type PrimID =
+  | MichelsonTypeID
+  | MichelsonDataID
+  | MichelsonSectionID
+  | MichelsonInstructionID
+  | 'Elt';
+
+const primitives: PrimID[] = [
```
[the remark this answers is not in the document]
e-asphyx: I failed to merge the first implementation and reapplied everything on top of the new branch forgetting this change. Thank you for pointing.

packages/taquito-michel-codec/test/typecheck.spec.ts
e-asphyx: Well, the Tezos test set has a contract which has a clearly unsorted data but supposed to be correct at the same time. I suppose my assumption about the sorting criteria was incorrect from the beginning so I just disabled this ordering check until I figure out how to do it properly. At least I'm pretty sure that key hash comparison based on base58 form is not how it supposed to work.

Issue #1472, opened 2022-03-23
[Chore] Review and fix Error classes in Taquito


dsawali:
### Description
Currently many error classes in Taquito returns a regular `Error` class.

Evaluate these Error class instances and adjust them to use custom errors accordingly on a case by case basis.

This is to provide better error handling experience for our users. 

- Identify and evaluate Error classes, and create a custom error class for them if needed
- Update Typedoc documentation

List of packages to evaluate: 

- [x] taquito/taquito
- [x] taquito/beacon-wallet
- [x] taquito/contracts-library
- [x] taquito/http-utils
- [x] taquito/ledger-signer
- [ ] taquito/local-forging
- [ ] taquito/michel-codec
- [x] taquito/michelson-encoder
- [x] taquito/remote-signer
- [x] taquito/rpc
- [x] taquito/signer
- [x] taquito/tezbridge-signer
- [x] taquito/tezbridge-wallet
- [x] taquito/tzip12
- [x] taquito/tzip16
- [x] taquito/utils

PR #1487, opened 2022-03-25
evaluated error classes for user clarity

dsawali:
#1472 
- Updated some error classes to throw a custom error instead of a plain `Error` object
- Updated `TezosOperationError` to have an extra field `errorDetails` for developers to pass where the error is being thrown from
- Updated some error messages to be more specific and give more info to the users
- Added `@category` and `@description` annotations above error classes to distinguish them from regular classes in TypeDoc

## Release Note Draft Snippet


packages/taquito/src/operations/operation-emitter.ts
```diff
@@ -251,7 +254,7 @@ export abstract class OperationEmitter {
     const errors = flattenErrors(results);
 
     if (errors.length) {
-      throw new TezosOperationError(errors);
+      throw new TezosOperationError(errors, 'Error occurred during operation preparation');
```
roxaneletourneau, line 257: We can specify that the error happens during the simulation of the validation of the operation

packages/taquito-michelson-encoder/src/tokens/createToken.ts
```diff
@@ -2,6 +2,10 @@ import { tokens } from './tokens';
 import { Token } from './token';
 import { PairToken } from './pair';
 
+/**
+ *  @category Error
+ *  @description Error that indicates a token having malformed data
```
roxaneletourneau, line 7: This error can indicate that a script contains an invalid type or a type that is not supported by the Michelson Encoder. 

Maybe we can add a note saying that if the type is valid but not supported, we encourage the user to open an issue and we will add support for it. This can happen if there are new types in a new protocol and we forget to add support for them..

packages/taquito-michelson-encoder/src/tokens/token.ts
```diff
@@ -1,6 +1,10 @@
 import { MichelsonV1Expression } from '@taquito/rpc';
 import { TokenSchema } from '../schema/types';
 
+/**
+ *  @category Error
+ *  @description Error that indicates an invalid token being passed or used
```
roxaneletourneau, line 6: I don't think that the users are necessarily familiar with the term token; tokens are created by the Michelson Encoder to represent Michelson types and are used internally. We can add precision on the description.  We throw a TokenValidationError when the MichelsonEncoder fails to encode invalid/incorrect data. Encoding here means transforming js data into Michelson JSON format. For example, if an address is expected but a number is received, then we throw a ValidationError.

packages/taquito/src/contract/errors.ts
```diff
@@ -12,6 +16,10 @@ export class InvalidParameterError extends Error {
   }
 }
 
+/**
+ *  @category Error
+ *  @description Error that indicates an an undefined lambda contract
+ */
 export class UndefinedLambdaContractError extends Error {
```
roxaneletourneau, line 23: We can delete this error; it is not used anymore.

packages/taquito/src/error.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ *  @category Error
+ *  @description Error that indicates invalid confirmation count has been returned
```
roxaneletourneau, line 3: has been passed or configured

packages/taquito/src/error.ts
```diff
@@ -0,0 +1,32 @@
...
+export class InvalidConfirmationCountError extends Error {
+  public name = 'InvalidConfirmationCountError';
+  constructor(public message: string) {
+    super(message);
+  }
+}
+
+/**
+ *  @category Error
+ *  @description Error that indicates undefined confirmation has been returned
```
roxaneletourneau, line 14: The number of confirmation counts has not been specify or configured

packages/taquito/src/operations/operation-errors.ts
```diff
@@ -22,16 +22,19 @@ const isErrorWithMessage = (error: any): error is TezosOperationErrorWithMessage
   return 'with' in error;
 };
 
+/**
+ *  @category Error
+ *  @description Generic tezos error that will be thrown when a mistake occurs when doing an operation
```
roxaneletourneau, line 27: We can add a link to Tezos error: https://tezos.gitlab.io/api/errors.html

packages/taquito/src/wallet/operation.ts
```diff
@@ -14,9 +14,14 @@ import { Context } from '../context';
 import { Receipt, receiptFromOperation } from './receipt';
 import { validateOperation, ValidationResult, InvalidOperationHashError } from '@taquito/utils';
 import { BlockIdentifier } from '../read-provider/interface';
+import { InvalidConfirmationCountError, ConfirmationUndefinedError } from '../error';
 
 export type OperationStatus = 'pending' | 'unknown' | OperationResultStatusEnum;
 
+/**
+ *  @category Error
+ *  @description Error that indicates a missed block when waiting for confirmation from the blockchain
```
roxaneletourneau, line 23: Missed a block when polling to retrieve new head block. This can happen if the polling interval is greater than the time between block.

packages/taquito-michelson-encoder/src/tokens/token.ts
dsawali: that's a good point. I can update it to have a friendlier description

Issue #1506, opened 2022-04-06
Update Error classes (continued)


dsawali:
Update error classes in Taquito to standardize and remove the use of generic `Error` classes.

Continuation of #1472 

- Update and adjust Error classes accordingly
- Update or add TypeDoc annotations

PR #1523, opened 2022-04-08
added more custom error classes and updated error descriptions

dsawali:
#1506 

## Release Note Draft Snippet


packages/taquito-local-forging/src/proto12-ithaca/schema.ts
```diff
@@ -36,6 +37,6 @@ export const operationDecoderProto12 =
         ...decodedObj,
       };
     } else {
-      throw new Error(`Unsupported operation ${op[0]}`);
+      throw new UnsupportedOperationError(op[0].toString());
```
roxaneletourneau, line 40: When there is data manipulation like `toString` in this case, I think it would be valuable to add a unit test just to make sure that the message renders as expected. (This applies to other errors following a similar pattern)

roxaneletourneau (changes requested):

packages/taquito-local-forging/src/schema/operation.ts
```diff
@@ -127,7 +128,7 @@ export const schemaEncoder =
         const values = value[key];
 
         if (!Array.isArray(values)) {
-          throw new Error(`Exepected value to be Array ${JSON.stringify(values)}`);
+          throw new OperationDecodingError(`Exepected value to be Array ${JSON.stringify(values)}`);
```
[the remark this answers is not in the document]
dsawali: you are right, I misread decoding/encoding

packages/taquito-michelson-encoder/src/schema/error.ts
```diff
@@ -20,3 +20,86 @@ export class InvalidScriptError extends Error {
...
+/**
+ *  @category Error
+ *  @description Error that indicates an invalid RPC response being passed or used
+ */
+export class InvalidRpcResponseError extends Error {
+  public name = 'InvalidRpcResponseError';
+  constructor(public script: any) {
+    super(`
+      Invalid RPC response passed as argument(s): \n
+      ${JSON.stringify(script)}
```
roxaneletourneau, line 33: Showing the script in the error message can make the error message very long. It is better to not include it in the error message?

packages/taquito-michelson-encoder/src/schema/parameter.ts
```diff
@@ -55,7 +56,7 @@ export class ParameterSchema {
         throw ex;
       }
 
-      throw new Error(`Unable to encode parameter. ${ex}`);
+      throw new ParameterEncodingError('Unable to encode parameter', args.toString(), ex);
```
roxaneletourneau, line 59: parsing of args to string could have unit testing with different args

packages/taquito-michelson-encoder/src/schema/parameter.ts
```diff
@@ -67,7 +68,7 @@ export class ParameterSchema {
         throw ex;
       }
 
-      throw new Error(`Unable to encode parameter object. ${ex}`);
+      throw new ParameterEncodingError('Unable to encode parameter object', _value, ex);
```
roxaneletourneau, line 71: Consider adding a unit test for this error message to.

roxaneletourneau (changes requested):

packages/taquito-michelson-encoder/src/schema/error.ts
dsawali: that's a good point. I will remove it but keep the property

roxaneletourneau (approved):

PR #1530, opened 2022-04-11
add support for transaction rollup contract addresses

zamrokk:
I just authorized "txr1" rollup addresses to the list of valid contract prefixes.
rollups on Mondaynet are protocol smart contract that acts like any user smart contract, taquito should allow to forge a Contract object from a rollup contract address

This is a minor change but it is blocking my development on a dapp for the Tezos rollups

### DIFF
--- a/packages/taquito-utils/src/constants.ts
+++ b/packages/taquito-utils/src/constants.ts
@@ -96,7 +96,7 @@ export const prefix = {
   [Prefix.TXI]: new Uint8Array([79, 148, 196]),
   [Prefix.TXM]: new Uint8Array([79, 149, 30]),
   [Prefix.TXC]: new Uint8Array([79, 148, 17]),
-  [Prefix.TXMR]: new Uint8Array([18, 7, 206, 087]),
+  [Prefix.TXMR]: new Uint8Array([18, 7, 206, 87]),
   [Prefix.TXRL]: new Uint8Array([79, 146, 82]),
   [Prefix.TXW]: new Uint8Array([79, 150, 72]),
    
--- a/packages/taquito-utils/src/constants.ts
+++ b/packages/taquito-utils/src/constants.ts
@@ -4,6 +4,7 @@ export enum Prefix {
   TZ3 = 'tz3',
   KT = 'KT',
   KT1 = 'KT1',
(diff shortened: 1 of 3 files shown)


packages/taquito-utils/src/constants.ts
```diff
@@ -45,6 +46,7 @@ export const prefix = {
   [Prefix.TZ3]: new Uint8Array([6, 161, 164]),
   [Prefix.KT]: new Uint8Array([2, 90, 121]),
   [Prefix.KT1]: new Uint8Array([2, 90, 121]),
+  [Prefix.TXR1]: new Uint8Array([2, 90, 121]),
```
roxaneletourneau, line 49: I am wondering if the prefix should rather be [1, 128, 120, 031]. See [TXR1](https://gitlab.com/tezos/tezos/-/blob/master/src/proto_alpha/lib_protocol/tx_rollup_prefixes.ml#L35)

zamrokk: you are right, I fixed it and added other definition from rollups

Thank you  =)

packages/taquito-utils/src/constants.ts
```diff
@@ -79,14 +91,25 @@ export const prefix = {
   [Prefix.TZ]: new Uint8Array([2, 90, 121]),
 
   [Prefix.VH]: new Uint8Array([1, 106, 242]),
+
+  [Prefix.TXR1]: new Uint8Array([1, 128, 120, 31]),
+  [Prefix.TXI]: new Uint8Array([79, 148, 196]),
+  [Prefix.TXM]: new Uint8Array([79, 149, 30]),
+  [Prefix.TXC]: new Uint8Array([79, 148, 17]),
+  [Prefix.TXMR]: new Uint8Array([18, 7, 206, 087]),
```
roxaneletourneau, line 99: Hi @zamrokk 
The build fails because there should be no '0' before '87'.

zamrokk:
Sorry I was out for holidays

Thank you for the unit tests fix, btw

What do you need to pass the last steps ?

roxaneletourneau:
all good, thanks @zamrokk 
It will be included in the next release shortly.

PR #1549, opened 2022-04-15
J protocol 

### LINKED ISSUE #1539: Taquito Integration Tests to Protocol J - contract-permits.spec.ts - storage_exhausted.operation
On Monday net

the test suite

contract-permits.spec.ts

fails with

```
    TezosOperationError: (temporary) proto.alpha.storage_exhausted.operation

      200 |     // Fail early in case of errors
      201 |     if (errors.length) {
    > 202 |       throw new TezosOperationError(errors, 'Error occurred during estimation');
          |             ^
      203 |     }
      204 |
      205 |     let numberOfOps = 1;

      at RPCEstimateProvider.<anonymous> (../packages/taquito/src/estimate/rpc-estimate-provider.ts:202:13)
      at fulfilled (../packages/taquito/dist/taquito.umd.js:41:62)
```

### LINKED ISSUE #1540: Taquito Integration Tests to Protocol J - contract-estimation-tests.spec.ts - balance too low
On Monday net

contract-estimation-tests.spec.ts fails due to balance too low

this test can be fixed 

to fix, modify line 329:   if (protocol === Protocols.Psithaca2 || Protocols.ProtoALpha) {

### LINKED ISSUE #1541: Taquito Integration Tests to Protocol J - local-forging.spec.ts - forging result bytes differ in two digits
On Monday net the test suite local-forging.spec.ts has several tests that fail

in each case the difference is a 94 changed to a 86 somewhere in the bute string:
 example
    Expected: "a99b946c97ada0f42c1bdeae0383db7893351232a832d00d0cd716eb6f66e5616d0035e993d8c7aaa42b5e3ccd86a33390ececc73abd904e010a0a000000000011020000000c0500036c0501036c0502039600000002030b"
----------------------------------------------------------------------^

    Received: "a99b946c97ada0f42c1bdeae0383db7893351232a832d00d0cd716eb6f66e5616d0035e993d8c7aaa42b5e3ccd86a33390ececc73abd904e010a0a000000000011020000000c0500036c0501036c0502038400000002030b"
----------------------------------------------------------------------^

the following tests fail this way:

    X Should give the same result as when forging with the rpc: Origination operation (sapling_transaction) (http://mondaynet.ecadinfra.com:8732) (192 ms)
    ✕ Should give the same result as when forging with the rpc: Origination with sapling_transaction in storage (http://mondaynet.ecadinfra.com:8732) (201 ms)
    ✕ Should give the same result as when forging with the rpc: Origination where storage has a pair of 3 annotated args whose 2 sapling_transaction (http://mondaynet.ecadinfra.com:8732) (197 ms)
    ✕ Should give the same result as when forging with the rpc: Origination where storage is a pair of 2 annotated sapling_transaction and parameter contains a pair of 3 args (http://mondaynet.ecadinfra.com:8732) (192 ms)
    ✕ Should give the same result as when forging with the rpc: Origination with SAPLING_EMPTY_STATE instruction (http://mondaynet.ecadinfra.com:8732) (194 ms)
    ✕ Should give the same result as when forging with the rpc: Origination with optional sapling_transaction in storage (http://mondaynet.ecadinfra.com:8732) (194 ms)
    ✕ Should give the same result as when forging with the rpc: Origination where parameter is an annotated pairs of 18 args (http://mondaynet.ecadinfra.com:8732) (198 ms)
    ✕ Should give the same result as when forging with the rpc: Origination where parameter is a pairs of 18 args without annotation (http://mondaynet.ecadinfra.com:8732) (187 ms)

roxaneletourneau:
In the J protocol, the op code for sapling_transaction is 96 instead of 84. 84 is now the op code for the 
type sapling_transaction_deprecated. This breaking change is handled by choosing the appropriate encoders/decoders based on the protocol hash.

Closes #1541
Fix #1539 
Fix #1540


packages/taquito-local-forging/src/taquito-local-forging.ts
```diff
@@ -30,19 +36,27 @@ export function getCodec(codec: CODEC, proto: ProtocolsHash) {
         return decodersProto12[codec](consumer) as any;
       },
     };
-  } else {
+  } else if (ProtoInferiorTo(proto, ProtocolsHash.Psithaca2)) {
```
dsawali, line 39: will this be something we need to update in the future with new protocol releases?

roxaneletourneau: Yes, when there are breaking changes regarding forging between the protocols, we will need to support protocols we are using in the integration tests

packages/taquito-local-forging/src/proto13-jakarta/michelson-proto13/codec-proto13.ts
```diff
@@ -0,0 +1,176 @@
...
+    return primEncoderProto13(value);
+  } else if (isBytes(value)) {
+    return bytesEncoder(value);
+  } else if (isString(value)) {
+    return stringEncoder(value);
+  } else if (isInt(value)) {
+    return intEncoder(value);
+  }
+
+  throw new Error('Unexpected value');
```
dsawali, line 57: I believe we have added a new error class in the error pattern PR called `UnexpectedMichelsonValueError` for this case

packages/taquito-local-forging/src/taquito-local-forging.ts
dsawali: I see. I am approving, but I think it might be worth looking into streamlining this as a config so that we won't have to update the code each time a new protocol drops

PR #1554, opened 2022-04-19
Corrected Signing doc and created doc for FAILWITH errors

claudebarde:
Corrected Signing doc and created doc for FAILWITH errors

### DIFF
--- a/docs/failwith_errors.md
+++ b/docs/failwith_errors.md
@@ -1,0 +1,83 @@
+---
+title: Catching error messages
+id: failwith_errors
+author: Claude Barde
+---
+
+# Catching `FAILWITH` errors with Taquito
+
+When a contract call fails because of the `FAILWITH` instruction, an error is returned by the node. This error is made available by Taquito as an instance of the `TezosOperationError` class.
+
+```ts
+try {
+    // a contract call that fails
+} catch (err) {
+    if (err instanceof TezosOperationError) {
+      // process the error
+    }
+}
+```  
+&nbsp;  
+If the type of value returned by the contract is a `string`, the value will be available in the `message` property of the `TezosOperationError` instance, for example:
+```ts
+try {
+    const contractAddress = "KT1V4jzQATak5UYDLqqnU3buakcsbcj1vAXQ";
+    const contract = await Tezos.contract.at(contractAddress);
+    const op = await contract.methods.fail_with_string("error").send();
+    await op.confirmation();
+    console.log(op.hash);
+} catch (err) {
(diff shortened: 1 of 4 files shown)


docs/failwith_errors.md
```diff
@@ -0,0 +1,83 @@
...
+    if (err instanceof TezosOperationError) {
+      console.log(err.errors[1])
+    }
+}
+```
+will log the following:
+```ts
+{
+  kind: 'temporary',
+  id: 'proto.011-PtHangz2.michelson_v1.script_rejected',
```
michaelkernaghan, line 58: maybe upgrade the example to be ithaca?

docs/failwith_errors.md
```diff
@@ -0,0 +1,83 @@
...
+} catch (err) {
+    if (err instanceof TezosOperationError) {
+      console.log(err.message) // will log "error"
+    }
+}
+```
+&nbsp;  
+However, if the value is another type, the error message will be available in the `errors` property.
+
+Among other information, the `TezosOperationError` instance has a property called `errors` whose value is an array of 2 elements. The second element (at index 1) holds the error message returned by the `FAILWITH` instruction, i.e. the last value that was on the stack before the contract call failed. This value is an object with a `with` property where you can find the Michelson expression of the last value on the stack.
```
roxaneletourneau, line 38: I am not sure the array always has 2 elements. I think it can be more. It can be safer to say that the error message returned by the `FAILWITH` instruction will be in the last element of `errors`.

docs/failwith_errors.md
```diff
@@ -0,0 +1,83 @@
...
+```ts
+try {
+    const contractAddress = "KT1V4jzQATak5UYDLqqnU3buakcsbcj1vAXQ";
+    const contract = await Tezos.contract.at(contractAddress);
+    const op = await contract.methods.fail_with_int(2).send();
+    await op.confirmation();
+    console.log(op.hash);
+} catch (err) {
+    if (err instanceof TezosOperationError) {
+      console.log(err.errors[1])
```
roxaneletourneau, line 50: console.log(err.errors[err.errors.length - 1])

docs/failwith_errors.md
claudebarde: Done

docs/failwith_errors.md
claudebarde: Corrected

docs/failwith_errors.md
claudebarde: Corrected

PR #1583, opened 2022-05-02
fix sequential big map and contract schema

zainen:
preload example-activate with unused faucet.
comb through other non working examples

all the additions to package-locks in different packages. should be removed from this PR? 

## Release Note Draft Snippet


example/contract-schema.ts
```diff
@@ -1,20 +1,45 @@
...
+        console.log(await contract.storage());
+        console.log("checking balance")
+        const balance = await tezos.tz.getBalance(await signer.publicKeyHash());
+        const balanceMin = new BigNumber(100);
+        if (balance < balanceMin) {
+            console.log("topping up funds")
+            console.log("Balance before", balance)
+            const sendMutez = await tezosLender.contract.transfer({ to: await signer.publicKeyHash(), amount: 100 });
+            await sendMutez.confirmation();
+            console.log("Balance after")
```
roxaneletourneau, line 40: Seems like the intention was to log the balance after, but it is missing in the log

roxaneletourneau (commented):
Looks good to me, just one small comment :)

example/contract-schema.ts
zainen: 👍 thanks for the catch :D

PR #1584, opened 2022-05-02
update no unused vars, strict

zainen:
linter throw error if contains unused variables / uses require() instead of import

ignores arguments and variables that start with `_`

only one disable comment for require() in patch.js

## Release Note Draft Snippet

### DIFF
--- a/packages/taquito/src/contract/contract.ts
+++ b/packages/taquito/src/contract/contract.ts
@@ -80,9 +80,6 @@ const isView = (entrypoint: MichelsonV1Expression): boolean => {
 export type Contract = ContractAbstraction<ContractProvider>;
 export type WalletContract = ContractAbstraction<Wallet>;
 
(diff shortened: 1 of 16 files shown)


packages/taquito/src/contract/contract.ts
```diff
@@ -80,7 +80,7 @@ const isView = (entrypoint: MichelsonV1Expression): boolean => {
 export type Contract = ContractAbstraction<ContractProvider>;
 export type WalletContract = ContractAbstraction<Wallet>;
 
-const isContractProvider = (variableToCheck: any): variableToCheck is ContractProvider =>
+const _isContractProvider = (variableToCheck: any): variableToCheck is ContractProvider =>
```
roxaneletourneau, line 83: This is not used anymore and could be deleted.

roxaneletourneau (approved):

PR #1585, opened 2022-05-04
Start eslint any

zainen:
some changes as a start so the PR isnt very large 

all tests pass and build properly but definitely would like more eyes on it to make sure I havent changed too much 😄 

## Release Note Draft Snippet

### DIFF
--- a/packages/taquito-local-forging/src/validator.ts
+++ b/packages/taquito-local-forging/src/validator.ts
@@ -1,3 +1,5 @@
+import { OpKind } from './../../taquito-rpc/src/opkind';
+import { OperationContents } from '@taquito/rpc';
 import {
   ActivationSchema,
   DelegationSchema,
@@ -48,7 +50,7 @@ const deleteArrayElementByValue = (array: string[], item: string) => {
 /**
  * @returns A boolean value to indicate whether the operation kind is valid or not
  */
-export const validateOperationKind = (opKind: string) => {
+export const validateOperationKind = (opKind: OpKind) => {
   const opKindList = Object.keys(OperationKindMapping);
   return opKindList.includes(opKind);
 };
@@ -59,8 +61,8 @@ export const validateOperationKind = (opKind: string) => {
  *
  *  @returns array element differences if there are missing required property keys
  */
-export const validateMissingProperty = (operationContent: any) => {
-  const kind: OperationKind = operationContent.kind;
+export const validateMissingProperty = (operationContent: OperationContents) => {
+  const kind = operationContent.kind as OperationKind;
 
   const keys = Object.keys(operationContent);
   const cleanKeys = deleteArrayElementByValue(keys, 'kind');
--- a/packages/taquito-local-forging/src/validator.ts
+++ b/packages/taquito-local-forging/src/validator.ts
@@ -1,4 +1,4 @@
-import { OpKind } from './../../taquito-rpc/src/opkind';
+import { OpKind } from '@taquito/rpc';
 import { OperationContents } from '@taquito/rpc';
 import {
   ActivationSchema,
--- a/packages/taquito-local-forging/test/allTestsCases.ts
+++ b/packages/taquito-local-forging/test/allTestsCases.ts
@@ -30,6 +30,8 @@ import {
 } from './data/contract_with_constant';
 import { codeViewsTopLevel, storageViewsTopLevel } from './data/contract_views_top_level';
 import { opMappingProto13 } from './../src/proto13-jakarta/constants-proto13';
+import {ForgeParams} from "../src/interface"
+import { MichelsonV1Expression, OpKind } from '@taquito/rpc';
 
 function extractOp(
   startIndex: number,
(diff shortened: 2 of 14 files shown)


packages/taquito-local-forging/test/data/generic_contract.ts
```diff
@@ -1,8 +1,10 @@
+import { MichelsonValue } from "../../src/michelson/codec";
+
 export const genericStorage = {
   prim: 'Unit',
 };
 
-export const genericCode = (op: any) => [
+export const genericCode = (op: MichelsonValue) => [
```
roxaneletourneau, line 7: I think the type should be a string instead of MichelsonValue. It will be one of the values in opMapping: https://github.com/ecadlabs/taquito/blob/master/packages/taquito-local-forging/src/constants.ts#L45

packages/taquito-local-forging/test/allTestsCases.ts
```diff
@@ -465,22 +468,22 @@ export const commonCases: TestCase[] = [
...
+            kind: OpKind.ORIGINATION,
             counter: '1',
             source: 'tz1QZ6KY7d3BuZDT1d19dUxoQrtFPN2QJ3hn',
             fee: '10000',
             gas_limit: '10',
             storage_limit: '10',
             balance: '0',
             script: {
-              code: genericCode(op),
+              code: genericCode(op as unknown as MichelsonValue) as MichelsonV1Expression[],
```
roxaneletourneau, line 486: I think that genericCode should accept a string, see comment below. That way, it won't be necessary to caste here.

packages/taquito-rpc/src/taquito-rpc.ts
```diff
@@ -458,7 +458,6 @@ export class RpcClient implements RpcClientInterface {
       'endorsing_reward_per_slot',
       'double_baking_punishment',
       'delay_increment_per_round',
-      'tx_rollup_commitment_bond'
```
roxaneletourneau, line 461: What is the reason for removing the property? it might be a mistake

zainen: definitely unintentional 😮

packages/taquito-rpc/src/rpc-client-modules/rpc-cache.ts
```diff
@@ -88,8 +88,8 @@ export class RpcClientCache implements RpcClientInterface {
   private formatCacheKey(
     rpcUrl: string,
     rpcMethodName: string,
-    rpcMethodParams: any[],
-    rpcMethodData?: any
+    rpcMethodParams: unknown[],
+    rpcMethodData?: string
```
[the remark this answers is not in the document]
zainen: Any Idea what the rpcMethodParams might be like? I am trying to get better at folowing paths to figure out what the type would be. ideally I can replace all `any / unknown` with interfaces or types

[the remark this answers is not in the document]
roxaneletourneau: `rpcMethodParams` is an array of strings and objects. (there is not always an object but see line 523 in `getBakingRights` for an example)
There is also a case where it is an empty array: see `getChainId` at line 801

packages/taquito-utils/src/validators.ts
```diff
@@ -81,7 +84,7 @@ const blockPrefix = [Prefix.B]
  * // This example return 3 which correspond to VALID
  * ```
  */
-export function validateAddress(value: any): ValidationResult {
+export function validateAddress(value: string | string[]): ValidationResult {
```
roxaneletourneau, line 87: I think this should only by string, no array

packages/taquito-utils/src/validators.ts
```diff
@@ -100,7 +103,7 @@ export function validateAddress(value: any): ValidationResult {
  * // This example return 3 which correspond to VALID
  * ```
  */
-export function validateChain(value: any): ValidationResult {
+export function validateChain(value: string | string[]): ValidationResult {
```
roxaneletourneau, line 106: I think this should only by string, no array

packages/taquito-utils/src/validators.ts
roxaneletourneau: The test case `expect(validateAddress([])).toEqual(ValidationResult.NO_PREFIX_MATCHED);` doesn't make really sense and can be removed.

zainen:
@roxaneletourneau I think this one is ready to be looked at one more time. I do need to go through the integration tests one more time since there are failing tests but not sure if they are related

packages/taquito-rpc/src/rpc-client-modules/rpc-cache.ts
```diff
@@ -56,6 +56,9 @@ interface CachedDataInterface {
     response: Promise<any>;
   };
 }
+
+type RpcMethodParam = string | UnparsingMode | BigMapKey | BakingRightsQueryArguments | PackDataParams
```
roxaneletourneau, line 60: EndorsingRightsQueryArguments is missing, see line 549 in getEndorsingRights

packages/taquito-local-forging/src/validator.ts
```diff
@@ -1,3 +1,5 @@
+import { OpKind } from './../../taquito-rpc/src/opkind';
```
roxaneletourneau, line 1: The import should be `import { OpKind } from '@taquito/rpc'` because rpc and local-forging are distinct packages

PR #1600, opened 2022-05-06
feat(sapling viewing key): inMemoryViewingKey class

roxaneletourneau:
A class that holds the viewing key and can be instantiated with a spending key

## Release Note Draft Snippet


packages/taquito-sapling/src/in-memory-viewing-key.ts
```diff
@@ -0,0 +1,76 @@
+import { b58cdecode, b58cencode, Prefix, prefix, validateSpendingKey } from '@taquito/utils';
```
dsawali, line 1: since `validateSpendingKey` is probably something that's quite specific to the Sapling package, could we consider moving that into a util.ts file in the Sapling package instead?

roxaneletourneau: I am not sure. All those validation functions are in the utils package with their prefixes, I thought it was a good place to add this one too.

dsawali: yeah that also makes sense. Sounds good!

PR #1627, opened 2022-05-13
Zs 1612 tx rollup l2 address

zainen:
all tests passing now with addition for tx_rollup_l2_address

not sure if this is all that needs to be done 😮  feels very minimal right now

## Release Note Draft Snippet


packages/taquito-michel-codec/src/michelson-validator.ts
```diff
@@ -417,6 +420,12 @@ export function assertMichelsonInstruction(ex: Expr): ex is MichelsonCode {
         }
         break;
 
+        case "MIN_BLOCK_TIME":
```
roxaneletourneau, line 423: I am not sure MIN_BLOCK_TIME should be added here as it doesn't take an argument.

packages/taquito-michel-codec/src/utils.ts
```diff
@@ -301,6 +302,7 @@ export const tezosPrefix: Record<TezosIDType, TezosIDPrefix> = {
   P256Signature: [64, [54, 240, 44, 52]], // p2sig(98)
   GenericSignature: [64, [4, 130, 43]], // sig(96)
   ChainID: [4, [87, 82, 0]],
+  tx_rollup_l2_address: [20, [1, 128, 120, 31]]
```
roxaneletourneau, line 305: Consider using PascalCase like for the other keys.

I am a bit confused, I am not sure if it is the correct prefix. From tezos code:
```
let rollup_address =
  {
    b58check_prefix = "\001\128\120\031";
    prefix = "txr1";
    hash_size = 20;
    b58check_size = 37;
  }

let l2_address =
  {
    b58check_prefix = "\006\161\166";
    prefix = "tz4";
    hash_size = 20;
    b58check_size = 36;
  }
```
Does tx_rollup_l2_address refer to a tz4 or txr1 address?

roxaneletourneau (changes requested):

packages/taquito-michel-codec/src/utils.ts
zainen: the contracts that i found in regards to this were with the txr1 address so i used that value but am not sure if its the correct one to use or not

roxaneletourneau: Ok we can validate that easily.
We can deploy the following contract:
```
parameter tx_rollup_l2_address;
storage unit;
code { DROP; UNIT; NIL operation; PAIR; }
```
and try to call the default entrypoint with a tz4 and a txr1 address and see what is accepted.

Do you want to try?

zainen: response from the oru-tx channel was that the tz4 are for addresses of account 
txr1 are for addresses of the rollup itself 
tru2 is out of date

roxaneletourneau (approved):

e-asphyx (approved):

PR #1655, opened 2022-05-26
Mk contract security research cases

### LINKED ISSUE #1654: Add to The Taquito Integration Tests the Contract Security tests from InferenceAG / TezosSecurityBaselineChecking - Part 3
includes

integration-tests/contract-security-balance-instruction.spec.ts
integration-tests/contract-security-wrong-field-annotations.spec.ts
integration-tests/contract-security-on-chain-views.spec.ts

These tests were delayed as they required extra research to prepare.

michaelkernaghan:
Close #1654 

These are the remaining contract vulnerability regression tests, to check that vulnerabilities are not exposed by code changes when Tezos changes protocols.


integration-tests/contract-security-wrong-field-annotations.spec.ts
```diff
@@ -0,0 +1,275 @@
...
+  const Tezos = lib;
+  const mondaynet = protocol === Protocols.ProtoALpha ? test: test.skip;
+
+  describe(`Test contracts to verify wrong field annotations are leading to failed transactions using: ${rpc}`, () => {
+    beforeEach(async (done) => {
+      await setup();
+      done();
+    });
+
+    mondaynet('Verify annotation combinations on ithacanet', async () => {
```
roxaneletourneau, line 61: `Verify annotation combinations on ithacanet` should be mondaynet

integration-tests/contract-security-wrong-field-annotations.spec.ts
```diff
@@ -0,0 +1,275 @@
...
+      // Calling Michelson code: CONTRACT %addNoAnnot (pair (nat %valueA) (nat %valueB)) ;
+      try {
+        const op2 = await Tezos.contract
+          .at(contractAddition.address)
+          .then((contract) => {
+            return contract.methodsObject
+              .call({
+                entrypoint: 'addNoAnnot',
+                param: {
+                  0: `1`,
```
roxaneletourneau, line 101: Taquito returns error `'[valueA] Value is not a number: undefined'` because the annotations is incorrect.
Instead of index 0 , it should be valueA
Instead of index 1, it should be valueB

integration-tests/contract-security-wrong-field-annotations.spec.ts
```diff
@@ -0,0 +1,275 @@
...
+      
+      try{
+      const op3 = await Tezos.contract
+        .at(contractAddition.address)
+        .then((contract) => {
+          return contract.methodsObject
+            .call({
+              entrypoint: 'AddWrongAnnot',
+              param: {
+                a: `1`,
```
roxaneletourneau, line 127: Instead of a , the annot should be valueA
Instead of b, the annot should be valueB

integration-tests/contract-security-wrong-field-annotations.spec.ts
```diff
@@ -0,0 +1,275 @@
...
+
+      try{
+      const op = await Tezos.contract
+        .at(contractAddition.address)
+        .then((contract) => {
+          return contract.methodsObject
+            .callWrongAnnot({
+              entrypoint: 'AddNoAnnot',
+              param: {
+                1: `1`,
```
roxaneletourneau, line 238: Instead of 1 , the annot should be a
Instead of 2, the annot should be b

integration-tests/contract-security-wrong-field-annotations.spec.ts
michaelkernaghan: fixed.

integration-tests/contract-security-wrong-field-annotations.spec.ts
michaelkernaghan: I made the change but it still gives the same error,.

integration-tests/contract-security-wrong-field-annotations.spec.ts
michaelkernaghan: I made the change but it still gives the same error,.

integration-tests/contract-security-wrong-field-annotations.spec.ts
michaelkernaghan: I made the change but it still gives the same error,.

Issue #1616, opened 2022-05-09
Rollup support: Contract and Estimate APIs (part 1)


roxaneletourneau:
Add support to estimate and inject the new operation kinds related to rollup to the block chain:

- tx_rollup_origination
- tx_rollup_submit_batch

roxaneletourneau:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Taquito-Development-Team-60ca841c87ca580010057707/issues/ecadlabs/taquito/1616?planning-poker) with ZenHub @dsawali @zainen

PR #1669, opened 2022-06-02
feat(contract, estimeta, batch): support tx_rollup_submit_batch

roxaneletourneau:
Added support on the contract, batch and estimate API allowing users to send a batch to a rollup

re #1616

## Release Note Draft Snippet


packages/taquito/src/contract/interface.ts
```diff
@@ -175,4 +179,24 @@ export interface ContractProvider extends StorageProvider {
...
+
+  /**
+   *
+   * @description Originate a new tx rollup. Will sign and inject an operation using the current context
+   *
+   * @returns An operation handle with the result from the rpc node
+   *
+   * @param TxRollupOriginateParams Originate rollup operation parameter
+   */
+  originateTxRollup(params?: TxRollupOriginateParams): Promise<TxRollupOriginationOperation>;
```
dsawali, line 191: this might be personal preference, but because the other function interface name is `txRollupSubmitBatch`, maybe we could rename this one to be `txRollupOriginate` to have a consistent naming pattern? What do you think?

packages/taquito/src/contract/rpc-contract-provider.ts
```diff
@@ -442,6 +448,70 @@ export class RpcContractProvider
...
 
+  /**
+   *
+   * @description Originate a new tx rollup. Will sign and inject an operation using the current context
+   *
+   * @returns An operation handle with the result from the rpc node
+   *
+   * @param TxRollupOriginateParams Originate rollup operation parameter
+   */
+  async originateTxRollup(params?: TxRollupOriginateParams) {
```
dsawali, line 459: same comment as the interface

packages/taquito/src/estimate/rpc-estimate-provider.ts
```diff
@@ -153,6 +158,7 @@ export class RPCEstimateProvider extends OperationEmitter implements EstimationP
         'storage_size' in result && 'global_address' in result
           ? Number(result.storage_size) || 0
           : 0;
+      totalStorage += 'originated_rollup' in result ? tx_rollup_origination_size : 0;
```
dsawali, line 161: could you give a brief breakdown to what this line is doing?

dsawali (changes requested):

packages/taquito/src/contract/interface.ts
roxaneletourneau: Yes I was thinking the same at some point. I will make the change, thanks

packages/taquito/src/estimate/rpc-estimate-provider.ts
roxaneletourneau: Originating a rollup has a storage cost on the blockchain. We need to take into account the rollup origination size (a constant) to calculate the right storageLimit.

PR #1711, opened 2022-06-23
[ci] Send Slack notification for Mondaynet integration tests result

danielelisi:
Send a slack notification to the `taquito_integration_tests` Slack channel with the result of the scheduled integration test ran against Mondaynet network and Tezedge nodes on Ithacanet

## Test Plan
Triggered the workflow and confirmed we are receiving the Slack notification
Screenshot from 2022-06-23 15-44-27


danielelisi:
The Slack notification is working correctly. The `Mondaynet Integration Test` job isn't passing for an error unrelated to this PR.

PR #1719, opened 2022-06-28
canary PR to check mondaynet

michaelkernaghan:
Mondaynet integration tests will run in cicd if the branch name includes mondaynet. 
This PR is to be used to initiate the cicd mondaynet tests to verify their current status.


michaelkernaghan:
replace with mondaynet-test-new

Issue #1624, opened 2022-05-12
Accept hex prefixed with 0x as well


jchenche:
**Description**
Right now, Taquito only accepts hex bytes without the `0x` prefix. But we want to support both formats, with or without `0x`.

**Steps To Reproduce**
1. Deploy the following contract:
```
parameter %abyte bytes;
storage bytes;
code { CAR ;
       PUSH nat 3;
       PUSH nat 1;
       SLICE ;
       IF_NONE { FAIL } {};
       NIL operation ;
       PAIR }
```

2. Call an entrypoint with the following code (omitting the setups):
Note that contract KT1MwH7SqQgfod9Lmsu9dqFmY3SBASEiEjku is already deployed to Ithaca.
```typescript
Tezos.contract
  .at('KT1MwH7SqQgfod9Lmsu9dqFmY3SBASEiEjku')
  .then((contract) => {
	// console.log(util.inspect(contract.script, false, null, true))
    return contract.methods.default("0x12a47ef2").send()
  })
  .then((op) => {
    console.log(`Waiting for ${op.hash} to be confirmed...`);
    return op.confirmation(3).then(() => op.hash);
  })
  .then((hash) => console.log(`Operation injected: https://ithaca.tzstats.com/${hash}`))
  .catch((error) => console.log(`Error: ${error}`));
```

**Current behaviour**
It will give an "Invalid bytes" error.

**Expected behavior**
Calling an entrypoint should succeed and after calling the entrypoint, if you go and inspect the storage using TzKT, it should be `a47ef2`.

**Extra**
Code responsible for this:
```typescript
  private isValid(val: any): BytesValidationError | null {
    if (typeof val === 'string' && /^[0-9a-fA-F]*$/.test(val) && val.length % 2 === 0) {
      return null;
    } else {
      return new BytesValidationError(val, this, `Invalid bytes: ${val}`);
    }
  }
```
from file `taquito/packages/taquito-michelson-encoder/src/tokens/comparable/bytes.ts`

roxaneletourneau:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Taquito-Development-Team-60ca841c87ca580010057707/issues/ecadlabs/taquito/1624?planning-poker) with ZenHub @dsawali @zainen @hui-an-yang

PR #1735, opened 2022-07-06
added support for 0x prefix hex string in michelson encoder

dsawali:
#1624 

## Release Note Draft Snippet

### DIFF
--- a/packages/taquito-utils/src/taquito-utils.ts
+++ b/packages/taquito-utils/src/taquito-utils.ts
@@ -347,5 +347,5 @@ export function bytes2Char(hex: string): string {
  * @param hex string to strip prefix from
  */
 export function stripHexPrefix(hex: string): string {
-  return hex.slice(2);
+  return hex.startsWith('0x') ? hex.slice(2) : hex;
 }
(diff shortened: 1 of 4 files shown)


hui-an-yang (approved):
lgtm!

packages/taquito-utils/src/taquito-utils.ts
```diff
@@ -339,3 +339,13 @@ export function char2Bytes(str: string) {
...
 }
+
+/**
+ *
+ * @description Strips the first 2 characters of a hex string (0x)
+ *
+ * @param hex string to strip prefix from
+ */
+export function stripHexPrefix(hex: string): string {
+  return hex.slice(2);
```
roxaneletourneau, line 350: Really minor suggestion, the validation `.startsWith('0x')` could be part of the `stripHexPrefix` function and the original string can be returned if it doesn't start with 0x. It will avoid doing the validation `val.startsWith('0x')` every time the function is called.

dsawali: sounds good, I will rework that function

roxaneletourneau (approved):

PR #1757, opened 2022-07-20
Debug tz2 verification

ricklove:
image


jonsutherland:
My bad guys, this is actually an invalid address and the failure is legitimate.  Sorry for the confusion

Issue #1661, opened 2022-05-31
Retrieve contract addresses from batch operation


roxaneletourneau:
**Is your feature request related to a problem and use case? Please describe.**
When batching multiple originations, I would like to retrieve the addresses of the newly originated contracts.

**Describe the solution you'd like**
Add a member on the BatchWalletOperation and the BatchOperation which is an array of contract addresses. Parse the addresses from the operation result and add them to the member.

**Additional context**
Feature requested on Telegram, May 22

roxaneletourneau:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Taquito-Development-Team-60ca841c87ca580010057707/issues/ecadlabs/taquito/1661?planning-poker) with ZenHub @dsawali @zainen

PR #1772, opened 2022-07-26
Add member function that retrieves originated contract addresses

dsawali:
#1661 
- Added member function to retrieve batched originated contract addresses
- Updated integration tests

## Release Note Draft Snippet

### DIFF
--- a/integration-tests/contract-batch.spec.ts
+++ b/integration-tests/contract-batch.spec.ts
@@ -2,7 +2,6 @@ import { CONFIGS } from './config';
 import { ligoSample, ligoSampleMichelson } from './data/ligo-simple-contract';
 import { managerCode } from './data/manager_code';
 import { MANAGER_LAMBDA, OpKind } from '@taquito/taquito';
-import { BatchOperation } from 'taquito/dist/types/operations';
 
 CONFIGS().forEach(({ lib, rpc, setup, knownBaker, knownContract, createAddress }) => {
     const Tezos = lib;
@@ -164,23 +163,18 @@ CONFIGS().forEach(({ lib, rpc, setup, knownBaker, knownContract, createAddress }
             done();
         });
 
-        test('Batch multiple originations and get contract address info from originatedConctractAddresses member variable', async (done) => {
+        test('Batch multiple originations and get contract address info from getOriginatedContractAddresses member function', async (done) => {
           const batch = Tezos.contract
               .batch()
               .withOrigination({
                 balance: '1',
                 code: ligoSample,
                 storage: 0
-              })
-              .withOrigination({
-                balance: '1',
-                code: ligoSampleMichelson,
-                storage: 0
-              })
+              });
               
           const op = await batch.send();
           await op.confirmation();
-          
+
           const addresses = op.getOriginatedContractAddresses();
           expect(op.status).toEqual('applied');
           expect(addresses.length).toEqual(2);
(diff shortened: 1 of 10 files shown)


integration-tests/contract-batch.spec.ts
```diff
@@ -2,6 +2,7 @@ import { CONFIGS } from './config';
 import { ligoSample, ligoSampleMichelson } from './data/ligo-simple-contract';
 import { managerCode } from './data/manager_code';
 import { MANAGER_LAMBDA, OpKind } from '@taquito/taquito';
+import { BatchOperation } from 'taquito/dist/types/operations';
```
roxaneletourneau, line 5: the import path doesn't seem right.

integration-tests/contract-batch.spec.ts
```diff
@@ -162,5 +163,29 @@ CONFIGS().forEach(({ lib, rpc, setup, knownBaker, knownContract, createAddress }
             expect(batchOp.status).toEqual('applied');
             done();
         });
+
+        test('Batch multiple originations and get contract address info from originatedConctractAddresses member variable', async (done) => {
```
roxaneletourneau, line 167: Consider correcting the test title with "member function"

packages/taquito/src/operations/batch-operation.ts
```diff
@@ -33,6 +33,19 @@ export class BatchOperation
...
   }
 
+  public getOriginatedContractAddresses(): string[] {
+    const originationOpResults = this.results.filter(
+      (x) => x.kind === 'origination'
+    ) as OperationContentsAndResultOrigination[];
+
+    let addresses: string[] = [];
+    for (const res of originationOpResults) {
+      addresses = [...addresses, ...res.metadata.operation_result.originated_contracts!];
```
roxaneletourneau, line 43: Are we sure that `originated_contracts` will always be defined? Could it lead to an unexpected error if it is not the case? It might be safer to validate that originated_contracts is defined and then append its values rather than using the non-nul assertion operator.

packages/taquito/src/wallet/batch-operation.ts
```diff
@@ -16,11 +21,26 @@ export class BatchWalletOperation extends WalletOperation {
...
+  public getOriginatedContractAddresses = async (): Promise<string[]> => {
+    const opResult = await this.operationResults();
+
+    const originationOpResults = opResult.filter(
+      (x) => x.kind === 'origination'
+    ) as OperationContentsAndResultOrigination[];
+
+    let addresses: string[] = [];
+    for (const res of originationOpResults) {
+      addresses = [...addresses, ...res.metadata.operation_result.originated_contracts!];
```
roxaneletourneau, line 38: see precedent comment for "!"

roxaneletourneau (changes requested):
Can you add unit tests as well with different scenarios: 1 origination, multiple originations, one where `originated_contracts` is undefined

packages/taquito/src/operations/batch-operation.ts
dsawali: that's a good point, I was under the impression that the property will always be there. I'll amend it

integration-tests/contract-batch.spec.ts
dsawali: good catch, looks like auto import didn't get the correct path

roxaneletourneau (approved):

PR #1774, opened 2022-07-27
Mondaynet PR for checking CICD ; not to be merged.

michaelkernaghan:
For tracking diffs between current protocol and what is on Monday net new each week.

replaces mondaynet-test which has build issues


michaelkernaghan:
there is some corruption in this branch that is worth finding. Closing this PR and starting another

PR #1777, opened 2022-08-02
refactor(sapling): small refactors

roxaneletourneau:
Moved some files in specific folders to make the package cleaner

## Release Note Draft Snippet


dsawali:
could we combine `helpers.ts` and `utils.ts` in this case?

zainen (approved):
LGTM i'll do better at remembering my semicolons 😅 
and the preference is to have `'` rather than `"` from what im getting. I will do better at remembering that since I'm more used to having it as `"`

roxaneletourneau:
> could we combine `helpers.ts` and `utils.ts` in this case?

the utils file only has functions used by the sapling state, so I put the file inside the sapling-sate folder.
The helpers file is used by the sapling-tx-viewer, so I created a sapling-tx-viewer where I put the file.

roxaneletourneau:
> LGTM i'll do better at remembering my semicolons sweat_smile and the preference is to have `'` rather than `"` from what im getting. I will do better at remembering that since I'm more used to having it as `"`

This is an autoformatting that is applied when I run "npm run commit", not sure why it is not doing the same formatting on your side @zainen

dsawali:
@roxaneletourneau Ah okay that's fair. I don't really have a strong preference, it's just that they have a similar usage semantically. I think it's okay to keep it separate for now. If there are overlaps in the future we could look into combining them and putting them higher in the scope

dsawali (approved):

PR #1778, opened 2022-08-02
remove `gas buffer` from empty implicit account test

### LINKED ISSUE #1771: Emptying an implicit account does not cost extra-gas anymore
We have an integration test where we added a gas buffer to be able to drain an implicit account. Validate if the test is passing without the workaround(buffer) and remove the workaround if so.
The test is: contract-empty-implicit-account-into-new-implicit-account.spec.ts

**Additional context**
https://gitlab.com/tezos/tezos/-/merge_requests/5566

michaelkernaghan:
CLOSES #1771 

should be no change for jakartanet

For kathmandunet, the gas buffer should no longer be required.


michaelkernaghan:
this PR has been replaced by a newer and simpler one as it is in too much of a mess.

PR #1793, opened 2022-08-09
feat(taquito): add getSaplingDiffByContract method to the TzReadProvider

roxaneletourneau:
BREAKING CHANGE:
Classes implementing the TzReadProvider interface will need to implement
 the getSaplingDiffByContract method

## Release Note Draft Snippet

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,6 +37,7 @@ jobs:
 
   integration-tests:
     runs-on: ubuntu-latest
+    continue-on-error: true
     strategy:
       matrix:
         protocol:
--- a/packages/taquito/src/read-provider/interface.ts
+++ b/packages/taquito/src/read-provider/interface.ts
@@ -114,8 +114,8 @@ export interface TzReadProvider {
 
   /**
    * @description Access the sapling state of a smart contract.
-   * @param contractAddress Address of the smart contract
-   * @param block from which we want to retrieve the sapling state
+   * @param contractAddress The address of the smart contract
+   * @param block The block you want to retrieve the sapling state from
    */
   getSaplingDiffByContract(
     contractAddress: string,
--- a/packages/taquito/src/read-provider/interface.ts
+++ b/packages/taquito/src/read-provider/interface.ts
@@ -113,6 +113,16 @@ export interface TzReadProvider {
   ): Promise<SaplingDiffResponse>;
 
   /**
+   * @description Access the sapling state of a smart contract.
+   * @param contractAddress Address of the smart contract
+   * @param block from which we want to retrieve the sapling state
+   */
+  getSaplingDiffByContract(
+    contractAddress: string,
+    block: BlockIdentifier
+  ): Promise<SaplingDiffResponse>;
+
+  /**
    * @description Return the list of entrypoints of the contract
    * @param contract address of the contract we want to get the entrypoints of
    */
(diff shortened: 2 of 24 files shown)


packages/taquito/src/read-provider/interface.ts
```diff
@@ -112,6 +112,16 @@ export interface TzReadProvider {
     block: BlockIdentifier
   ): Promise<SaplingDiffResponse>;
 
+  /**
+   * @description Access the sapling state of a smart contract.
+   * @param contractAddress Address of the smart contract
```
zainen, line 117: nit "contractAddress ~Address~ of the smart contract"

zainen: not sure if we need two addresses there

packages/taquito/src/read-provider/interface.ts
```diff
@@ -112,6 +112,16 @@ export interface TzReadProvider {
...
+   * @param block from which we want to retrieve the sapling state
```
zainen, line 118: nit "the block you want to retrieve the sapling state from"

packages/taquito/src/read-provider/rpc-read-adapter.ts
```diff
@@ -159,6 +159,18 @@ export class RpcReadAdapter implements TzReadProvider {
     return this.context.rpc.getSaplingDiffById(saplingStateQuery.id, { block: String(block) });
   }
 
+  /**
+   * @description Access the sapling state of a smart contract.
+   * @param contractAddress Address of the smart contract
+   * @param block from which we want to retrieve the sapling state
```
zainen, line 165: same if applicable

zainen (approved):
super minor method description suggestions 
LGTM 😄

.github/workflows/main.yml
```diff
@@ -37,6 +37,7 @@ jobs:
 
   integration-tests:
     runs-on: ubuntu-latest
+    continue-on-error: true
```
dsawali, line 40: was there a reason why we had it not continue when we encounter a failure before? I assume it was to make sure it doesn't take too much time when it does fail. On that note, why did we decide to change?

packages/taquito/src/read-provider/interface.ts
dsawali: @zainen since `contractAddress` is the variable name I think it makes sense

.github/workflows/main.yml
roxaneletourneau: @dsawali I just cherry-picked the commit from Dan to fix the integration tests run that were canceling by themselves.

packages/taquito/src/read-provider/interface.ts
roxaneletourneau: yes "contractAddress" is the name of the param and "Address of the smart contract" is the description

roxaneletourneau: It looks like this with typedoc:
image

zainen: maybe contractAddress the address of the smart contract?

zainen: the address address just made me do a double take :O

roxaneletourneau: ok i will add "the"

Issue #1572, opened 2022-04-26
SaplingTransactionBuilder


Innkst:
SaplingTransactionBuilder - This class allows preparing a sapling transaction (shield, unshield, internal transaction) which is represented by a hexadecimal string. A user can prepare a sapling transaction and then send it to the smart contract using the <code class="notranslate">ContractAbstraction` from the <code class="notranslate">@taquito/taquito` package.

PR #1794, opened 2022-08-09
feat(sapling): prepare sapling transactions

roxaneletourneau:
The SaplingToolkit class allows to prepare shielded transactions, unshielded transactions and
sapling transactions based on a spendingkey

re #1572

## Release Note Draft Snippet


packages/taquito-sapling/src/sapling-tx-builder/sapling-transactions-builder.ts
```diff
@@ -0,0 +1,352 @@
...
+    this.#contractAddress = saplingContractDetails.contractAddress;
+    this.#memoSize = saplingContractDetails.memoSize;
+    this.#inMemorySpendingKey = inMemorySpendingKey;
+    this.#saplingState = new SaplingState(32);
+    this.#saplingId = saplingContractDetails.saplingId;
+    this.#saplingWrapper = saplingWrapper;
+    this.#readProvider = readProvider;
+  }
+
+  async createShieldTx(
```
zainen, line 54: could the methods in these files use descriptions?

packages/taquito-sapling/src/taquito-sapling.ts
```diff
@@ -3,6 +3,323 @@
...
+        throw new InvalidMemo(memo, 'The memo is too long.');
+      }
+
+      formatedParams.push({ to: param.to, amount: amountMutez, memo });
+    });
+
+    return { formatedParams, totalAmount };
+  }
+
+  private async getRoot() {
```
zainen, line 233: is this guaranteed to return a root? or is it possible to be undefined? should we handle the potential undefined is so?

packages/taquito-sapling/src/taquito-sapling.ts
```diff
@@ -3,6 +3,323 @@
...
+        break;
+      }
+      case 'tz2': {
+        pad = Buffer.from('01', 'hex');
+        break;
+      }
+      case 'tz3': {
+        pad = Buffer.from('02', 'hex');
+        break;
+      }
```
zainen, line 261: since this is used in a couple places could this be turned into a helper function and included in all applicable places?

packages/taquito-sapling/src/types.ts
```diff
@@ -23,11 +36,129 @@ export interface Ciphertext {
...
 }
 
-export interface SaplingTransaction {
-  spendDescriptions: SpendDescription[];
-  outputDescriptions: OutputDescription[];
+export interface SaplingTransactionInput {
+  cv: Buffer;
+  nf: Buffer;
+  rk: Buffer;
+  proof: Buffer;
```
zainen, line 43: is there documentation for these to know what the abbreviation is referring to?

packages/taquito-sapling/src/types.ts
```diff
@@ -23,11 +36,129 @@ export interface Ciphertext {
...
+
+export type SaplingContractId =
+  | {
+      contractAddress?: never;
+      saplingId: string;
+    }
+  | {
+      contractAddress: string;
+      saplingId?: never;
+    };
```
zainen, line 143: not sure if i understand how this works as a type 
is it the same as `{contractAddress: string} | {saplingId: string}`?

packages/taquito-sapling/src/types.ts
```diff
@@ -23,11 +36,129 @@ export interface Ciphertext {
...
+  sumSelectedInputs: BigNumber;
+}
+
+export interface SaplingTransactionParamsTxBuilder {
+  root: string;
+  saplingTransactionParams: {
+    to: string;
+    amount: string;
+    memo: string;
+  }[];
```
zainen, line 156: maybe pull this type into another type? this same object is passed in as the param type and could use an interface to save on space and clutter

packages/taquito-sapling/test/sapling-tx-builder/sapling-module-wrapper.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+    });
+
+    expect(signedSpendDesc.signature.length).toEqual(64); // the returned value is changing
+    expect(signedSpendDesc.cv).toEqual(spendDescription.cv);
+    expect(signedSpendDesc.proof).toEqual(spendDescription.proof);
+    expect(signedSpendDesc.rk).toEqual(spendDescription.rk);
+    expect(signedSpendDesc.nf).toEqual(spendDescription.nf);
+
+    done();
+  }); */
```
zainen, line 115: still need?

zainen (changes requested):
just some questions and nits 
great work! there is so much to unpack there 😮

packages/taquito-sapling/src/constants.ts
```diff
@@ -1,2 +1,4 @@
 export const KDF_KEY = 'KDFSaplingForTezosV1';
 export const OCK_KEY = 'OCK_keystringderivation_TEZOS';
+export const defaultMemo = '';
```
dsawali, line 3: kind of a nitpick, but for constants maybe we should keep the same format like `DEFAULT_MEMO`

packages/taquito-sapling/src/constants.ts
```diff
@@ -1,2 +1,4 @@
...
+export const defaultBoundData = Buffer.from('', 'hex');
```
dsawali, line 4: same here with `DEFAULT_BOUND_DATA`

packages/taquito-sapling/src/error.ts
```diff
@@ -30,3 +30,38 @@ export class TreeConstructionFailure extends Error {
...
+
+/**
+ *  @category Error
+ *  @description Error indicating that there is not enough balance to prepare the sapling transaction
+ */
+export class InsufficientBalance extends Error {
+  public name = 'InsufficientBalance';
+  constructor(realBalance: string, amountToSpend: string) {
+    super(
+      `The transaction would spend ${amountToSpend} mutez while the balance is only ${realBalance} mutez.`
```
dsawali, line 53: I think the message wording can be a bit ambiguous in a sense that it _might_ still spend mutez. Maybe we could consider changing it to something like `Unable to spend ${amountToSpend} mutez while balance is only ${realBalance} mutez`

packages/taquito-sapling/src/sapling-tx-builder/sapling-module-wrapper.ts
```diff
@@ -0,0 +1,146 @@
+import * as sapling from '@airgap/sapling-wasm';
```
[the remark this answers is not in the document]
dsawali: @zainen it's the sapling-wasm library that help us do some sapling-specific stuff. It's a wrapper of a Rust implementation of Sapling based methods

packages/taquito-sapling/src/sapling-tx-builder/sapling-module-wrapper.ts
```diff
@@ -0,0 +1,146 @@
...
+  SaplingTransactionInput,
+} from '../types';
+import * as path from 'path';
+import * as fs from 'fs';
+import axios, { AxiosResponse } from 'axios';
+
+const SAPLING_PARAMS_DIR = path.resolve(__dirname, 'sapling-params');
+const ZCASH_DOWNLOAD_URL = 'https://download.z.cash/downloads';
+const SPEND_PARAMS_FILE_NAME = 'sapling-spend.params';
+const OUTPUT_PARAMS_FILE_NAME = 'sapling-output.params';
```
dsawali, line 16: are these variables something that are constant? Shall we consider moving them to `constants`?

packages/taquito-sapling/src/sapling-tx-builder/sapling-transactions-builder.ts
dsawali, line 54: I think `createShieldedTx` might be a more descriptive name for this method, what do you think?

dsawali (changes requested):

packages/taquito-sapling/src/sapling-tx-builder/sapling-module-wrapper.ts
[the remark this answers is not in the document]
zainen: sorry wasnt clear I meant the whole file for sapling-module-wrapper.ts not just the one file just picked the first line

packages/taquito-sapling/src/sapling-keys/in-memory-spending-key.ts
```diff
@@ -49,14 +49,14 @@ export class InMemorySpendingKey {
 
   /**
    *
-   * @param mnemonic list of words
+   * @param mnemonic strings of words
```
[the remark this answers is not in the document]
roxaneletourneau: yes thanks

packages/taquito-sapling/src/sapling-keys/in-memory-spending-key.ts
```diff
@@ -49,14 +49,14 @@ export class InMemorySpendingKey {
...
    * @param derivationPath tezos current standard 'm/'
    * @returns InMemorySpendingKey class instantiated
    */
 
-  static async fromMnemonic(mnemonic: string[], derivationPath = 'm/') {
+  static async fromMnemonic(mnemonic: string, derivationPath = 'm/') {
     // no password passed here. password provided only changes from sask -> MMXj
-    const fullSeed = await bip39.mnemonicToSeed(mnemonic.join(' '));
+    const fullSeed = await bip39.mnemonicToSeed(mnemonic);
```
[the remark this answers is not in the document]
roxaneletourneau: I replaced the array for a string for consistency with the `InMemorySigner` where we also use a string. 
Also, `bip39.generateMnemonic()` returns a string, so I found it more convenient to accept a string in the method.
Yes, if there is any errors in spacing it could cause a problem

packages/taquito-sapling/src/sapling-tx-builder/sapling-module-wrapper.ts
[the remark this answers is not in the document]
roxaneletourneau: It created a separate class to wrap the "sapling" instance just to be able to write my unit tests. Some of the methods from the sapling-wasm library require a sapling context which is a number that is always different, and the returned values are different from one run to another because of this context. I could not properly test without wrapping and mocking the class.

packages/taquito-sapling/src/taquito-sapling.ts
roxaneletourneau: I don't think the root can be undefined.. even when there is no transactions yet in the pool, there is a root

packages/taquito-sapling/src/types.ts
roxaneletourneau: There is, but would it be better if I use the whole name instead of the abbreviation?

packages/taquito-sapling/src/types.ts
roxaneletourneau: Yes it is the same

packages/taquito-sapling/src/types.ts
roxaneletourneau: I am not sure if I understand, do you refer to using `ParametersSaplingTransaction` instead? The properties type are not exactly the same.

packages/taquito-sapling/test/sapling-tx-builder/sapling-module-wrapper.spec.ts
roxaneletourneau: This test failed because of the issue when using the @airgap/sapling package with node. I commented out until this it is fixed.

packages/taquito-sapling/src/constants.ts
roxaneletourneau: Will address

packages/taquito-sapling/src/error.ts
roxaneletourneau: Thanks, I will make the change.

packages/taquito-sapling/src/sapling-tx-builder/sapling-transactions-builder.ts
roxaneletourneau: good catch

roxaneletourneau: I don't intend this file to be used by the users directly, so I didn't put descriptions. The separation there was mostly to help me write the unit tests.

packages/taquito-sapling/src/types.ts
zainen: i feel like for people working that know what it is it would be fine to keep as is but if people are new it would have to be fairly clear what they are supposed to be so if there is documentation for it I'm sure it would be fine

packages/taquito-sapling/src/types.ts
zainen: would it be more clear just to have it without the other option even with the `never`?

packages/taquito-sapling/src/sapling-tx-builder/sapling-transactions-builder.ts
```diff
@@ -0,0 +1,352 @@
...
+    this.#saplingWrapper = saplingWrapper;
+    this.#readProvider = readProvider;
+  }
+
+  async createShieldedTx(
+    saplingTransactionParams: {
+      to: string;
+      amount: string;
+      memo: string;
+    }[],
```
zainen, line 59: these were the ones i was referring to since they are objects as types for multiple places. Would it be easier to have the object be an interface like
```
export interface saplingTransactionParams {
      to: string;
      amount: string;
      memo: string;
    }
```
and then type the param `saplingTransactionParams: SaplingTransactionParams[]`

the name might make more sense if called something different than the key name since it requires the `[]`

zainen: really just a nit that would even though it would force the dev to look for the type it would remove clutter of the written in type

packages/taquito-sapling/src/types.ts
roxaneletourneau: Documentation is a bit spread, I renamed the variables to be more descriptive

packages/taquito-sapling/src/types.ts
roxaneletourneau: Might not be totally equivalent fianlly. The change caused a build issue:
```
src/sapling-tx-viewer/sapling-transaction-viewer.ts(120,33): error TS2339: Property 'saplingId' does not exist on type 'SaplingContractId'.
  Property 'saplingId' does not exist on type '{ contractAddress: string; }'.
src/sapling-tx-viewer/sapling-transaction-viewer.ts(122,39): error TS2339: Property 'saplingId' does not exist on type 'SaplingContractId'.
  Property 'saplingId' does not exist on type '{ contractAddress: string; }'.
src/sapling-tx-viewer/sapling-transaction-viewer.ts(125,40): error TS2339: Property 'contractAddress' does not exist on type 'SaplingContractId'.
  Property 'contractAddress' does not exist on type '{ saplingId: string; }'.
src/sapling-tx-viewer/sapling-transaction-viewer.ts(127,33): error TS2339: Property 'contractAddress' does not exist on type 'SaplingContractId'.
  Property 'contractAddress' does not exist on type '{ saplingId: string; }'.
```

packages/taquito-sapling/src/sapling-tx-builder/sapling-transactions-builder.ts
roxaneletourneau: ok got it, I made the suggested change

PR #1829, opened 2022-08-16
1762 map typecheck error bug

zainen:
Addresses reported bug that contract threw an error trying to access storage. 

removes the need for encodeObject method called when typechecking. 

checks if valid structure if needed in encodeObject or returns `void` if no error is thrown. 

## Release Note Draft Snippet

### DIFF
--- a/packages/taquito-michelson-encoder/src/tokens/comparable/string.ts
+++ b/packages/taquito-michelson-encoder/src/tokens/comparable/string.ts
@@ -43,7 +43,7 @@ export class StringToken extends ComparableToken {
     return { string: val };
   }
 
-  public TypecheckValue(_val: unknown) {
+  public TypecheckValue(_val?: unknown) {
     return
   }
 
--- a/packages/taquito-michelson-encoder/src/tokens/comparable/string.ts
+++ b/packages/taquito-michelson-encoder/src/tokens/comparable/string.ts
@@ -43,6 +43,10 @@ export class StringToken extends ComparableToken {
     return { string: val };
   }
 
+  public TypecheckValue(_val: unknown) {
+    return
+  }
+
   public ToKey({ string }: any) {
     return string;
   }
--- a/packages/taquito-michelson-encoder/src/tokens/comparable/string.ts
+++ b/packages/taquito-michelson-encoder/src/tokens/comparable/string.ts
@@ -43,7 +43,7 @@ export class StringToken extends ComparableToken {
     return { string: val };
   }
 
-  public TypecheckValue(_val?: unknown) {
+  public TypecheckValue(_val?: string) {
     return
   }
 
--- a/packages/taquito-michelson-encoder/src/tokens/comparable/timestamp.ts
+++ b/packages/taquito-michelson-encoder/src/tokens/comparable/timestamp.ts
@@ -34,7 +34,7 @@ export class TimestampToken extends ComparableToken {
     return { string: val };
   }
 
-  public TypecheckValue(_val: unknown) {
+  public TypecheckValue(_val?: unknown) {
     return
   }
 
--- a/packages/taquito-michelson-encoder/src/tokens/comparable/timestamp.ts
+++ b/packages/taquito-michelson-encoder/src/tokens/comparable/timestamp.ts
@@ -34,6 +34,10 @@ export class TimestampToken extends ComparableToken {
     return { string: val };
   }
 
+  public TypecheckValue(_val: unknown) {
+    return
+  }
+
   /**
    * @deprecated ExtractSchema has been deprecated in favor of generateSchema
    *
--- a/packages/taquito-michelson-encoder/src/tokens/comparable/timestamp.ts
+++ b/packages/taquito-michelson-encoder/src/tokens/comparable/timestamp.ts
@@ -34,7 +34,7 @@ export class TimestampToken extends ComparableToken {
     return { string: val };
   }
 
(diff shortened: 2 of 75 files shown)


packages/taquito-michelson-encoder/src/tokens/comparable/string.ts
```diff
@@ -43,6 +43,10 @@ export class StringToken extends ComparableToken {
     return { string: val };
   }
 
+  public TypecheckValue(_val: unknown) {
+    return
```
roxaneletourneau, line 47: should we validate that val is indeed a string?

packages/taquito-michelson-encoder/src/tokens/comparable/timestamp.ts
```diff
@@ -34,6 +34,10 @@ export class TimestampToken extends ComparableToken {
     return { string: val };
   }
 
+  public TypecheckValue(_val: unknown) {
+    return
```
roxaneletourneau, line 38: We should look at how timestamp values are represented by the RPC, and add an appropriate validation for val.

packages/taquito-michelson-encoder/src/tokens/map.ts
```diff
@@ -107,6 +107,13 @@ export class MapToken extends Token {
       });
   }
 
+  public TypecheckValue(val: unknown) {
+    const err = this.isValid(val);
```
roxaneletourneau, line 111: As this case is quite complex, I think we should have proper testing:
- a test where we retrieve the storage of a contract having a map inside a map

packages/taquito-michelson-encoder/src/tokens/pair.ts
```diff
@@ -158,6 +158,10 @@ export class PairToken extends ComparableToken {
     };
   }
 
+  public TypecheckValue(_val: unknown) {
+    return
```
roxaneletourneau, line 162: we should call TypecheckValue on the left and right tokens

packages/taquito-michelson-encoder/src/tokens/sapling-state.ts
```diff
@@ -66,6 +66,18 @@ export class SaplingStateToken extends Token {
     }
   }
 
+  public TypecheckValue(val: unknown) {
+    if (this.isValid(val)) {
```
roxaneletourneau, line 70: val here will be a string representing the sapling state id

packages/taquito-michelson-encoder/src/tokens/sapling-transaction-deprecated.ts
```diff
@@ -54,6 +54,10 @@ export class SaplingTransactionDeprecatedToken extends Token {
     return { bytes: String(val).toString() };
   }
 
+  public TypecheckValue(_val: unknown) {
+    return
```
roxaneletourneau, line 58: val will be an hexadecimal string

packages/taquito-michelson-encoder/src/tokens/sapling-transaction.ts
```diff
@@ -54,6 +54,10 @@ export class SaplingTransactionToken extends Token {
     return { bytes: String(val).toString() };
   }
 
+  public TypecheckValue(_val: unknown) {
+    return
```
roxaneletourneau, line 58: val will be an hexadecimal string

packages/taquito-michelson-encoder/src/tokens/ticket.ts
```diff
@@ -41,9 +41,14 @@ export class TicketToken extends Token {
     if (semantic && semantic[TicketToken.prim]) {
       return semantic[TicketToken.prim](args, this.val);
     }
+
     throw new EncodeTicketError();
   }
 
+  public TypecheckValue(_arg: unknown) {
+    return
```
roxaneletourneau, line 49: Consider adding appropriate validation for arg

packages/taquito-michelson-encoder/src/tokens/token.ts
```diff
@@ -103,6 +103,13 @@ export abstract class Token {
 
   public abstract EncodeObject(args: any, semantics?: SemanticEncoding): any;
 
+  /**
+   *
+   * @description checks the value of what is passed validating the same way as EncodeObject and throws error if necessary
+   * @param arg same as EncodeObject
```
roxaneletourneau, line 109: `arg same as EncodeObject` ->
Not exactly, there are discrepancies between EncodeObject, and this is why we need a separate TypecheckValue method to handle those case

packages/taquito/test/contract/rpc-contract-provider.spec.ts
```diff
@@ -1856,4 +1856,191 @@ describe('RpcContractProvider test', () => {
...
+            }
+          ]
+        }
+      })
+      const rpcContract = await rpcContractProvider.at('KT19mzgsjrR2Er4rm4vuDqAcMfBF5DBMs2uq');
+      const storage = await rpcContract.storage() as any;
+      expect(rpcContract).toBeDefined();
+
+      const keyList = storage.keyMap;
+      expect(keyList.size).toEqual(3);
```
roxaneletourneau, line 2042: It would be interesting to see how the ticket is represented in the storage as well.

This test is great, I think we should add more tests like that for map having different key and value:
- map having a map as key or value
- map having a bigmap as key or value
- map having [any other more complex type] as key or value

PR #1838, opened 2022-08-20
New 1762 map typcheck error

zainen:
have not recreated a failing test yet though 

## Release Note Draft Snippet


packages/taquito/test/contract/rpc-contract-provider.spec.ts
```diff
@@ -1856,4 +1856,233 @@ describe('RpcContractProvider test', () => {
       done();
     });
   });
+
+  describe('getSaplingDiffByID', () => {
```
roxaneletourneau, line 1860: This test and the following seem to be duplicated.

packages/taquito/test/contract/rpc-contract-provider.spec.ts
```diff
@@ -1856,4 +1856,233 @@ describe('RpcContractProvider test', () => {
...
+      done();
+    });
+  });
+
+  describe('Testing Bug #1762 mapTypecheckError', () => {
+    it('should have a defined response and storage of 3 should not throw an error', async (done) => {
+      mockRpcClient.getEntrypoints.mockResolvedValue({
+        entrypoints: {},
+      })
+      mockRpcClient.getContract.mockResolvedValue({
```
roxaneletourneau, line 1907: This mocked value takes considerable space in the test. Consider adding it to a separate file and importing it from there.

zainen: 👍

packages/taquito/test/contract/rpc-contract-provider.spec.ts
```diff
@@ -1856,4 +1857,20 @@ describe('RpcContractProvider test', () => {
       done();
     });
   });
+
+  describe('Testing Bug #1762 mapTypecheckError', () => {
```
dsawali, line 1861: Try and use the name of the member method/variable as the describe block name. In this case, it should be something like `describe('storage', ...`

describe blocks should be quite high level and general sounding

packages/taquito/test/contract/rpc-contract-provider.spec.ts
```diff
@@ -1856,4 +1857,20 @@ describe('RpcContractProvider test', () => {
...
+    it('should have a defined response and storage of 3 should not throw an error', async (done) => {
```
dsawali, line 1862: you can put the issue # in the it block like so:
```
it('should return storage without errors (#1762)', ...
```

or even a comment above the `it` block linking to the issue itself

dsawali (changes requested):
Could we also get a negative case unit test in for what would make a `.storage` call fail?

roxaneletourneau (approved):
Lgtm! :)

dsawali (approved):

Issue #1769, opened 2022-07-25
Replace consumed_gas with consumed_milligas


roxaneletourneau:
The field consumed_gas, deprecated in Jakarta, is removed in the K protocol.

**Describe the solution you'd like**
Replace the occurrence and logic around consumed_gas with consumed_milligas in all the Taquito codebases.

Some occurrence of `consumed_gas`:
- https://github.com/ecadlabs/taquito/blob/master/packages/taquito/src/estimate/rpc-estimate-provider.ts#L153
- https://github.com/ecadlabs/taquito/blob/master/packages/taquito/src/operations/batch-operation.ts#L63 (keep the `consumedGas` method but divide the result by 1000, add a new `consumedMilliGas` method. Do the same for the other operations.)
- https://github.com/ecadlabs/taquito/blob/master/packages/taquito/src/wallet/receipt.ts#L39

**Additional context**
https://gitlab.com/tezos/tezos/-/merge_requests/5536

PR #1854, opened 2022-08-25
fix(wallet receipt): replace consumed_gas with consumed_milligas

roxaneletourneau:
In kathmandu, the property consumed_gas is removed from the RPC responses in favor of
consumed_milligas. Added a totalMilliGas property to the receipt and totalGas is now the total
milliGas divided by 1000 and roundedup.

re #1769

## Release Note Draft Snippet

### DIFF
--- a/packages/taquito/src/wallet/receipt.ts
+++ b/packages/taquito/src/wallet/receipt.ts
@@ -1,8 +1,7 @@
 import { OperationContentsAndResult } from '@taquito/rpc';
 import BigNumber from 'bignumber.js';
+import { COST_PER_BYTE } from '../constants';
 import { flattenOperationResult } from '../operations/operation-errors';
-
-export const COST_PER_BYTE = 250;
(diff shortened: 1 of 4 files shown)


packages/taquito/src/wallet/receipt.ts
```diff
@@ -2,9 +2,12 @@ import { OperationContentsAndResult } from '@taquito/rpc';
 import BigNumber from 'bignumber.js';
 import { flattenOperationResult } from '../operations/operation-errors';
 
+export const COST_PER_BYTE = 250;
```
zainen, line 5: this is a protocol constant right? should we collect values like this in a constants file? so its easier to find and maintain.

although it wont be changing without a fair amount of forewarning

roxaneletourneau: Thanks, moved to the constants.ts file

Issue #1767, opened 2022-07-25
Support new operation Increase_paid_storage on the Contract API


roxaneletourneau:
**Is your feature request related to a problem and use case? Please describe.**
Increase_paid_storage is a new operation that enables a payer to increase the paid storage of a smart contract by some bytes amount. 

```
{ /* Increase_paid_storage */
         "kind": "increase_paid_storage",
         "source": $Signature.Public_key_hash,
         "fee": $014-PtKathma.mutez,
         "counter": $positive_bignum,
         "gas_limit": $positive_bignum,
         "storage_limit": $positive_bignum,
         "amount": $bignum,
         "destination": $014-PtKathma.contract_id.originated }
```

**Describe the solution you'd like**
- Add support for the operation in the `RpcContractProvider`, `RPCEstimateProvider` and `OperationBatch` classes (as well as to the associated interfaces)
- Add unit tests
- Add integration tests
- Add documentation on the website explaining how this new operation can resolve the issue where `several operations on the same contract could fail when they were added at the same level because the storage limit given was lower than the real paid_storage_size_diff`

**Additional context**
https://gitlab.com/tezos/tezos/-/merge_requests/5605

roxaneletourneau:
> This scenario on kathmanadunet concerns three addresses: the originater tz1VS2LW6jJvANsgZL8kBYgprGrUYcD7dJy9, the user tz1KmCaBGCZzwGumrZVQUV6DxdGavhqWKsVx, and the contract KT1TW1CC4ZN4WT2Bnx3PqkEohrRoSiqHXJaa. The whole scenario can be read by observing the operations run on the contract.
> Level 4792: the originater originates the contract. It has two entry-points, to push (add) or pop (remove) integers in the storage (a list of integers).
> 0.0905 storage fees: 105 bytes were added to the storage. Size is 105 bytes, 105 paid bytes.
> Level 4859: the user adds 1234567890 to the storage.
> 0.0015 storage fees: 6 bytes were added to the storage. Size is 111 bytes, 111 paid bytes.
> Level 4930: the user removes an integer from the storage.
> 0 storage fees: no byte was added to the storage. Size is 105 bytes, 111 paid bytes.
> Level 4943: the user adds 11235813 to the storage.
> 0 storage fees: no byte was added to the storage. Size is 110 bytes, 111 paid bytes.
[...]

PR #1870, opened 2022-08-31
1767 storage

dsawali:
#1767 

## Release Note Draft Snippet


packages/taquito/test/batch/rpc-batch-provider.spec.ts
```diff
@@ -566,6 +566,45 @@ describe('OperationBatch test', () => {
     });
   });
 
+  describe('withIncreasePaidStorage batch operation', () => {
+    it('should produce an operation batch which contains an increasePaidStorage operation', async (done) => {
```
roxaneletourneau, line 570: could you add the following test case as well:
- should produce a batch operation which contains an increasePaidStorage operation where fee, gas limit and storage limit are specified by the user
- should produce a batch operation which contains a reveal and an increasePaidStorage operation

packages/taquito/test/contract/rpc-contract-provider.spec.ts
```diff
@@ -1558,6 +1560,41 @@ describe('RpcContractProvider test', () => {
     });
   });
 
+  describe('increasePaidStorage', () => {
+    it('should produce an increasePaidStorage operation', async (done) => {
```
roxaneletourneau, line 1564: could you add a test :
should produce a reveal and an increasePaidStorage operation

packages/taquito/test/operations/increase-paid-storage-operation.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    fakeContext.rpc.getBlock.mockResolvedValue({
+      operations: [[{ hash: 'ooBghN2ok5EpgEuMqYWqvfwNLBiK9eNFoPai91iwqk2nRCyUKgE' }], [], [], []],
+      header: {
+        level: 1,
+      },
+    });
+  });
+
+  it('should return status of IncreasePaidStorage operation', () => {
```
roxaneletourneau, line 73: Can you add coverage for the other members: consumedMilligas, ...

integration-tests/contract-increase-paid-storage-operation.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+        })
+        .withIncreasePaidStorage({
+          amount: 1,
+          destination: 'KT1UiLW7MQCrgaG8pubSJsnpFZzxB2PMs92W'
+        })
+        .send();
+      await op.confirmation();
+      expect(op.status).toEqual('applied');
+      done();
+    })
```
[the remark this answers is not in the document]
dsawali: ah I thought this syntax was deprecated, will update

packages/taquito/test/contract/rpc-contract-provider.spec.ts
dsawali: looks like the current test already does both, I will update the test name then

roxaneletourneau: we should have a test that include a reveal if needed and on that doesn't

roxaneletourneau: I forgot to mention about this case too: validate that the user-specified fees will be taken into account when specified

roxaneletourneau (approved):
Thanks for adding coverage tests, lgtm :)

PR #1871, opened 2022-08-31
1743 support emit instruc

zainen:
Adds the `EMIT` instruction to the retStack function

adds two passing test and one invalid contract test 

current unknowns if done correctly are:

how is the type persisted to be added with the Event

when the `CONS` instruction is added to the stack it puts the `operation` into a list but loses the annotation.

is there a place missed to add the instruction `ty` as well as the annotated `tag` ? 

## Release Note Draft Snippet


packages/taquito-michel-codec/test/contracts_014.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+  ]
+};
+
+describe('PtKathmandu', () => {
+  for (const [group, list] of Object.entries(contracts)) {
+    describe(group, () => {
+      for (const contract of list) {
+        it(contract, () => {
+          const options: ContractOptions = {
+            protocol: Protocol.PtJakarta,
```
roxaneletourneau, line 35: Should be protocol Kathmandu

zainen: 👍

packages/taquito-michel-codec/src/michelson-typecheck.ts
```diff
@@ -2004,9 +2004,20 @@ function functionTypeInternal(
         ensurePushableType(s[0]);
         return [annotateVar({ prim: 'option', args: [instruction.args[1]] }), ...stack.slice(2)];
       }
+
       case 'MIN_BLOCK_TIME':
         return [annotateVar({ prim: 'nat' }), ...stack];
 
+      case 'EMIT': {
+        const ia = instructionAnn({ f: 1, t: 1}, { specialVar: true });
```
roxaneletourneau: consider removing `{ specialVar: true }` as it doesn't apply to EMIT

roxaneletourneau (commented):
Looks like `EMIT` should be in `primitives` in the binary.ts file

roxaneletourneau (approved):

PR #1896, opened 2022-09-12
Update for the Taquito test dapp

claudebarde:
- Added more convenient dropdowns instead of default select button
- Fixed issue with icons in the left-side list
- Added code to clean the local storage on mount and avoid invalid Beacon states
- Started work on adding Sapling tests (WIP)


apps/taquito-test-dapp/src/App.svelte
```diff
@@ -15,9 +16,30 @@
   // https://better-call.dev/kathmandunet/KT1BQuSVXWz23iGeXQCrAGR6GcVcqKeE1F7T/operations
 
   let browser = "";
+  let availableNetworks = [
+    { value: "ghostnet", label: "Ghostnet", group: "current testnets" },
+    { value: "kathmandunet", label: "Kathmandunet", group: "current testnets" },
+    { value: "mainnet", label: "Mainnet", group: "mainnet" },
+    { value: "dailynet", label: "Dailynet", group: "other testnets" },
+    { value: "hangzhounet", label: "Hangzhounet", group: "other testnets" },
```
roxaneletourneau, line 24: Should we remove hangzhounet? I don't think this testnet is still used.

claudebarde: You're right, I removed it :)

roxaneletourneau (approved):

PR #1913, opened 2022-09-14
Mk 1865 broken link in wallet doc

### LINKED ISSUE #1865: broken link in the Wallet doc
**Description**
There is a broken link on a docs page

**Steps To Reproduce**
Steps to reproduce the behaviour:
1. Go to Taquito online documentation
2. Go to https://tezostaquito.io/docs/wallet_api/#try-the-beacon-wallet
3. Scroll down to 'You can find a list of all the default handlers [in the beacon-sdk Github repo](https://github.com/airgap-it/beacon-sdk/blob/master/src/events.ts).'
4. The link to the GitHub repo gets a 404

**Expected behaviour**
Should load the intended repo

To Fix:
the correct link should be
https://github.com/airgap-it/beacon-sdk

michaelkernaghan:
fix the broken link in the wallet doc

CLOSE #1865


docs/wallet_API.md
```diff
@@ -81,7 +81,7 @@ const options = {
 const wallet = new BeaconWallet(options);
 ```
 
-The necessary bare minimum to instantiate the wallet is an object with a `name` property that contains the name of your dapp. However, the Beacon wallet allows you to customize your dapp responses to different events. In the example above, instead of getting the default Beacon pop-up after the user connects the wallet, it will display the available data in the console. You can use whatever solution you prefer for feedback. You can find a list of all the default handlers [in the beacon-sdk Github repo](https://github.com/airgap-it/beacon-sdk/blob/master/src/events.ts).
+The necessary bare minimum to instantiate the wallet is an object with a `name` property that contains the name of your dapp. However, the Beacon wallet allows you to customize your dapp responses to different events. In the example above, instead of getting the default Beacon pop-up after the user connects the wallet, it will display the available data in the console. You can use whatever solution you prefer for feedback. You can find a list of all the default handlers [in the beacon-sdk Github repo](https://github.com/airgap-it/beacon-sdk).
```
roxaneletourneau, line 84: Should we link the same page as before: https://github.com/airgap-it/beacon-sdk/blob/master/packages/beacon-dapp/src/events.ts?

michaelkernaghan: Replaced the link with the one you have supplied, which shows the desired information. Thanks.

Issue #1872, opened 2022-09-01
Update reference to testnet faucet


roxaneletourneau:
**Description**
The implementation of the testnet faucet has changed: https://teztnets.xyz/. It is not offering a faucet key in JSON format to download anymore. 

We need to 
- [ ] update our documentation accordingly. Using the InMemorySigner with a secret_key instead of a faucet. We can show how to retrieve the pkh from the secret_key and link to https://faucet.kathmandunet.teztnets.xyz/ to fund the account. We can have another documentation page telling how to create a secret_key. Add warning that this is only for testnet.
- [x] Adapt the integration test config file which offers the ability to run the tests with a faucet.
- [x] Update the examples in the "example" folder which rely on a faucet key.

roxaneletourneau:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Taquito-Development-Team-60ca841c87ca580010057707/issues/ecadlabs/taquito/1872?planning-poker) with Zenhub @dsawali @hui-an-yang @zainen

PR #1928, opened 2022-09-20
Update the examples from the example folder to remove usage of faucet…

roxaneletourneau:
… key

Updated the examples to use ghostnet at the same time to reduce maintenance.

#1872 

## Release Note Draft Snippet


example/deploy-integration-tests-contracts.ts
```diff
@@ -1,19 +1,18 @@
...
 import { knownBigMapContract } from './data/knownBigMapContract';
 import { singleSaplingStateContractJProtocol } from '../integration-tests/data/single_sapling_state_contract_jakarta_michelson';
-import Faucet from './faucet-interface';
 import { char2Bytes } from '@taquito/utils';
 import { fa2ForTokenMetadataView } from '../integration-tests/data/fa2-for-token-metadata-view';
 
-const { email, password, mnemonic, activation_code } =
-  require('./faucet-default-values.json') as Faucet;
-
 const provider = 'https://kathmandunet.ecadinfra.com/';
```
michaelkernaghan, line 9: Should this also change to ghost net?

roxaneletourneau: I left Kathmandu because I thought we use this file to deploy the known contracts used in the integration tests when there is a new protocol. So when adding support for L, we will replace Kathmandunet by Lnet. But I think we can completely delete the file because there is another script in the integration tests folder. Is it ok for you if a delete the file?

michaelkernaghan: Need to check the other script includes the contracts originated in this script

if any are missing they can be transferred over

michaelkernaghan: I will make a new issue for consolidating these two scripts into one

roxaneletourneau: Ok then I will merge this PR and we can follow up with the issue you created.

PR #1970, opened 2022-09-27
test(config txrollup): added origination of txrollup to the script us…

### LINKED ISSUE #1971: Adapt integration tests config so they can be run against a sandbox.
- [x] Investigate the issue where Taquito misses blocks when running against a sandbox
- [x] Allow running the test suite using a secret key
- [x] Add an environment variable to easy set the polling interval
- [x] Add an environment variable to disable the RPC caching
- [x] Modify the script "originate-know-contracts.ts" so the addresses of the known contract are outputted to a file (known-contracts-ProtoALph.ts). The config.ts file will read from "known-contracts-ProtoALph.ts" instead of having hardcoded contract addresses. This change allows us to deploy and use the "known contracts" when using a sandbox.
- [x] Create a dockerFile that can be used to run the tests (originate the "known contracts" + run the tests)

roxaneletourneau:
…ed to deploy known contracts

This change is needed to run the integrations against alpha using a sandbox. The script originate-know-contracts.ts deploy the contracts(or txRollup) that are required in different tests.

Closes #1971

### DIFF
--- a/integration-tests/config.ts
+++ b/integration-tests/config.ts
@@ -4,9 +4,9 @@ import { HttpBackend } from '@taquito/http-utils';
 import { b58cencode, Prefix, prefix } from '@taquito/utils';
 import { importKey, InMemorySigner } from '@taquito/signer';
 import { RpcClient, RpcClientCache } from '@taquito/rpc';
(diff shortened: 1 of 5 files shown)


zainen (approved):
LGTM :D

PR #1977, opened 2022-09-28
fixed michelson encoder for timestamp to handle numbers being passed

### LINKED ISSUE #1888: Local Forging fails for timestamps
**Description**
The default forger (Local Forging) fails to properly encode timestamp values.  A contract method with a single timestamp parameter fails to encode properly.

**Steps To Reproduce**
- Create a contract with a "set_something(timestamp)" method.
- Confirm Taquito sees it as a timestamp `toTransferParams()`, and `getSignature()`
`{
  ..snipped..
  parameter: {
    entrypoint: 'set_something',
    value: { string: 2022-09-06T16:52:13.000Z }
  }
}`
`[ 'timestamp' ]`
- Call the method passing Date.now()
- Taquito throws an exception:
`The contract call failed and the following error was returned: Failed to encode michelson value '[object Object]' UnexpectedMichelsonValueError: Failed to encode michelson value '[object Object]'
    at valueEncoder ([removed]/node_modules/@taquito/local-forging/dist/taquito-local-forging.umd.js:449:13)
    at parametersEncoder ([removed]/node_modules/@taquito/local-forging/dist/taquito-local-forging.umd.js:926:26)
    at [removed]/node_modules/@taquito/local-forging/dist/taquito-local-forging.umd.js:1118:29
    at Array.reduce (<anonymous>)
    at [removed]/node_modules/@taquito/local-forging/dist/taquito-local-forging.umd.js:1106:19
    at Object.encoders.<computed> [as transaction] ([removed]/node_modules/@taquito/local-forging/dist/taquito-local-forging.umd.js:1223:95)
    at [removed]/node_modules/@taquito/local-forging/dist/taquito-local-forging.umd.js:1090:75
    at [removed]/node_modules/@taquito/local-forging/dist/taquito-local-forging.umd.js:1114:79
    at Array.reduce (<anonymous>)
    at [removed]/node_modules/@taquito/local-forging/dist/taquito-local-forging.umd.js:1114:36 {
  value: { string: 1662575448078 }
}`
- Works using non-local forger:
`const rpcForger = tezos.getFactory(RpcForger)();
tezos.setForgerProvider(rpcForger);`

**Expected behavior**
Shouldn't throw an exception on Timestamps...

**Desktop (please complete the following information):**
- node v16.17.0
- taquito 13.0.0

**Additional context**
This came up on a (rarely used) script that was working properly on 11.2, but failing after upgrading to 13.0 (confirmed to also be failing on 12.0, likely when the change to local-forging by default was made?

dsawali:
Closes #1888 

## Release Note Draft Snippet

### DIFF
--- a/packages/taquito-michelson-encoder/src/tokens/comparable/timestamp.ts
+++ b/packages/taquito-michelson-encoder/src/tokens/comparable/timestamp.ts
@@ -25,7 +25,7 @@ export class TimestampToken extends ComparableToken {
   public Encode(args: any[]): any {
     const val = args.pop();
     if (typeof val === 'number') {
-      return { string: String(val) };
+      return { int: String(val) };
     } else {
       return { string: val };
     }
@@ -37,7 +37,7 @@ export class TimestampToken extends ComparableToken {
     }
 
     if (typeof val === 'number') {
-      return { string: String(val) };
+      return { int: String(val) };
     } else {
       return { string: val };
     }
--- a/packages/taquito-michelson-encoder/src/tokens/comparable/timestamp.ts
+++ b/packages/taquito-michelson-encoder/src/tokens/comparable/timestamp.ts
@@ -24,14 +24,23 @@ export class TimestampToken extends ComparableToken {
 
   public Encode(args: any[]): any {
     const val = args.pop();
-    return { string: val };
+    if (typeof val === 'number') {
+      return { string: String(val) };
+    } else {
+      return { string: val };
+    }
   }
 
   public EncodeObject(val: any, semantic?: SemanticEncoding): any {
     if (semantic && semantic[TimestampToken.prim]) {
       return semantic[TimestampToken.prim](val);
     }
-    return { string: val };
+
+    if (typeof val === 'number') {
+      return { string: String(val) };
+    } else {
+      return { string: val };
+    }
   }
 
   /**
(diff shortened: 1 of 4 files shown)


packages/taquito-michelson-encoder/src/tokens/comparable/timestamp.ts
```diff
@@ -24,14 +24,23 @@ export class TimestampToken extends ComparableToken {
 
   public Encode(args: any[]): any {
     const val = args.pop();
-    return { string: val };
+    if (typeof val === 'number') {
+      return { string: String(val) };
```
roxaneletourneau, line 28: Should it be { string: String(val) } or { int: String(val) }? Does both work?
Can you add an integration test with a timestamp in the storage to make sure the node accepts it? (maybe 2 tests, one where timestamps is a string and another is a number)

dsawali: wouldn't it be { string: String(val) } in this case since we are parsing from Number to String? I can confirm that String: String(val) works since I tried modifying some of our integration tests to make sure it worked.

Ah sure, I actually had that scenario on the integration test but decided not to put it in since I thought it would be an edge case, but I'll add it in the revision.

dsawali: Or perhaps I'm misunderstanding and we need to retain the original type that was passed instead?

dsawali: update: will be updating it to `int` instead as suggested in this thread: https://tezos-dev.slack.com/archives/C69RKF3PG/p1664818848427549

roxaneletourneau (approved):
lgtm!

PR #2021, opened 2022-10-12
1717 add a parameter to spcify amount and limits to with contract call

hui-an-yang:

## Release Note Draft Snippet


integration-tests/batch-api.spec.ts
```diff
@@ -12,7 +12,7 @@ CONFIGS().forEach(({ lib, rpc, setup, knownBaker, createAddress }) => {
       done()
     })
     it('Verify simple batch transfers with origination', async (done) => {
-      const batch = await Tezos.batch()
+      const batch = await Tezos.contract.batch()
```
roxaneletourneau, line 15: We already have a file to test the contract.batch: https://github.com/ecadlabs/taquito/blob/master/integration-tests/contract-batch.spec.ts
This one should not be modified. We will delete it when we remove the deprecated method.

packages/taquito/src/batch/rpc-batch-provider.ts
```diff
@@ -94,14 +95,16 @@ export class OperationBatch extends OperationEmitter {
 
   /**
    *
-   * @description Add a transaction operation to the batch
+   * @description Add a coantract call to the batch
```
roxaneletourneau, line 98: typo in contract

roxaneletourneau (changes requested):
Looks good :) some minor comments

integration-tests/batch-api.spec.ts
hui-an-yang: Gotcha, it's been rolled back

packages/taquito/src/batch/rpc-batch-provider.ts
hui-an-yang: Thanks, pushed new commit to address this

roxaneletourneau (approved):

PR #2022, opened 2022-10-13
1228 tz1 mnemonic

zainen:
Initial work for tz1 from mnemonic. made helper functions. currently only works for tz1 and will be restructuring for the ability for tz2/3 and bip25519 but quickly thrown together

is this an appropriate implementation way that allows for more flexibility? 

further refactor of the tz1 could make a `generateKey` function taht would handle all the logic in the case ed25519 block and other like it. 

just a quick look to make sure its on the right track

## Release Note Draft Snippet

### DIFF
--- a/packages/taquito-signer/test/taquito-signer.spec.ts
+++ b/packages/taquito-signer/test/taquito-signer.spec.ts
@@ -1,3 +1,7 @@
+import {
+  InvalidDerivationPathError,
+  InvalidMnemonicError,
+} from '../src/errors';
 import { InMemorySigner } from '../src/taquito-signer';
 
 describe('inmemory-signer', () => {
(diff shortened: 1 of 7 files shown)


packages/taquito-signer/test/taquito-signer.spec.ts
```diff
@@ -183,4 +184,49 @@ describe('inmemory-signer', () => {
     );
     done();
   });
+
+  it('Intantiate tz1 from mnemonic from in memory signer', async (done) => {
```
dsawali, line 188: typo 'instantiate'

also for consistency, let's still use the 'should..' form as the unit test name prefix

packages/taquito-signer/src/mnemonicHelpers.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+export const createNode = (seed: Uint8Array, key: Uint8Array): KeyNode => {
+  const derived = hmac(SHA512, key, seed);
+
+  return {
+    privateKey: derived.slice(0, 32),
+    chainCode: derived.slice(32)
+  };
+}
+
```
dsawali, line 19: I'm noticing some missing semi-colons and extra white spaces like these ones, could probably fix it by running the linter through it

dsawali (changes requested):

roxaneletourneau (commented):
Could you add typedoc comments for the new methods and Error class too?

PR #2024, opened 2022-10-14
fix(confirmationPollingTimeoutSecond): fix for contract/wallet API

### LINKED ISSUE #2006: confirmationPollingTimeoutSecond not working
**Description**
There is an issue with the `confirmationPollingTimeoutSecond` on the TezosToolkit. During the operation confirmation (both on the contract and wallet  API), the timer for the timeout seems to restart to 0 on every emission of a new block.

**Steps To Reproduce**

Execute the following code:
```
Tezos.setProvider({config: {confirmationPollingTimeoutSecond:15}})
const op = await Tezos.contract.originate({
        balance: "1",
        code: `parameter string;
        storage string;
        code {CAR;
              PUSH string "Hello ";
              CONCAT;
              NIL operation; PAIR};
        `,
        init: `"test"`,
        gasLimit:800000
})
await op.confirmation()
```
**Expected behavior**
We set a high `gasLimit`, so the operation will not be included in a block. We set a timeout of 15 seconds and expect that an error `'Confirmation polling timed out'` will be thrown after 15 seconds. The error is not thrown.

**Desktop (please complete the following information):**
 - Version 14.0.0

roxaneletourneau:
The timeout was not taken into consideration on both the contract and wallet API when confirming the operations. In cases where an operation was not found, our integration tests would never end instead of throwing a timeout error.

fix #2006

## Release Note Draft Snippet

### DIFF
--- a/integration-tests/contract-confirmation-timeout.spec.ts
+++ b/integration-tests/contract-confirmation-timeout.spec.ts
@@ -1,7 +1,5 @@
 import BigNumber from "bignumber.js";
-import { CONFIGS } from "./config";
-
-const sleep = (ms: number) => new Promise((r) => setTimeout(r, ms));
+import { CONFIGS, sleep } from "./config";
 
 CONFIGS().forEach(({ lib, rpc, setup }) => {
   const Tezos = lib;
--- a/integration-tests/contract-confirmation-timeout.spec.ts
+++ b/integration-tests/contract-confirmation-timeout.spec.ts
@@ -1,0 +1,40 @@
+import BigNumber from "bignumber.js";
+import { CONFIGS } from "./config";
+
+const sleep = (ms: number) => new Promise((r) => setTimeout(r, ms));
+
+CONFIGS().forEach(({ lib, rpc, setup }) => {
+  const Tezos = lib;
+  let timeBetweenBlocks: BigNumber;
+  describe(`Test confirmationPollingTimeoutSecond with contract API using: ${rpc}`, () => {
+
+    beforeEach(async (done) => {
+      await setup();
+      timeBetweenBlocks = (await Tezos.rpc.getConstants()).delay_increment_per_round ?? new BigNumber(15);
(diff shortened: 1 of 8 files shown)


packages/taquito/src/wallet/operation.ts
```diff
@@ -64,7 +66,8 @@ export class WalletOperation {
       return typeof x !== 'undefined';
     }),
     first(),
-    shareReplay({ bufferSize: 1, refCount: true })
+    publishReplay(1),
```
dsawali, line 69: I think publishReplay will be removed in version 8 of RxJS. https://rxjs.dev/api/operators/publishReplay#publishreplay

integration-tests/contract-confirmation-timeout.spec.ts
```diff
@@ -0,0 +1,40 @@
+import BigNumber from "bignumber.js";
+import { CONFIGS } from "./config";
+
+const sleep = (ms: number) => new Promise((r) => setTimeout(r, ms));
```
dsawali, line 4: since we used this more than a couple times now, perhaps we should create a utils.ts file in `integration-tests`?

packages/taquito/src/wallet/operation.ts
dsawali: It's prob not that big of a deal, but just something to keep in mind if we some day want to upgrade to RxJS 8

integration-tests/contract-confirmation-timeout.spec.ts
roxaneletourneau: Make sense, I will create the utils file, thanks.

dsawali (approved):

integration-tests/contract-confirmation-timeout.spec.ts
roxaneletourneau: Moved to the config.ts file instead

PR #2059, opened 2022-10-24
feat(wallet connect 2): @taquito/wallet-connect-2 package + support in the Taquito test dapp

### LINKED ISSUE #2145: Implement a wallet provider for wallet connect 2

Allow dapps to connect to wallets through wallet connect 2. Allow to send operations and sign payload.

- [x] Create a new package for wallet connect 2. The `WalletConnect2` class implements the `WalletProvider` interface and acts as a wrapper around the wc sign client. WalletConnect2 is intended to be used in dapps in a very similar way to BeaconWallet.
- [x] Allow to configure `WalletConnect2` with an existing pairing
- [x] Allow to configure `WalletConnect2` with a new pairing/session using the QR code modal
- [x] Allow to configure `WalletConnect2` with an existing session
- [x] wc2 allows to connect to multiple accounts; add a method `setActiveAccount` that the dapp needs to call to specify which account is used; the method needs to be called every time the active account is changed in the dapp
- [x] Namespace validate; reject the session if the namespace returned by the wallet is invalid
- [x] Permission handling; reject operation if the session doesn't have the correct permission
- [x] Add a listener for `session_delete`. It informs when a session is deleted on the wallet side. If the topic of the deleted session corresponds to the topic of the `session` member in the `WalletConnect2` class, set `session` to undefined.
- [x] Add listeners for session_update, to update the session stored in the WalletConnect2 class accordingly.
- [x] What happens if the chain/network is changed in the wallet? Can we give permission for multiple networks at once? (Yes) How to know what is the active network?
- [x] What happens if the account is changed in the wallet? Which event listener needs to be used?
- [ ] Add unit tests
- [x] Add a parameter named `registryUrl` to the `requestPermissions` method that allows specifying wallets displayed in the Modal.
- [ ] Provide a default list of all Tezos wallets compatible with wc2 that can be passed as the `registryUrl` param.

### LINKED ISSUE #2146: Support wallet connect 2 in the Taquito test dapp
Allow users to connect through beacon or wallet connect on the test dapp.

- [x] Add a button to connect to a wallet using wc2 and the `WalletConnect2` class from Taquito
- [x] wc2 allows to connect to multiple accounts; show all connected accounts; the user should be able to switch between them in the dapp
- [x] Allow connecting using an existing pairing
- [ ] Should we add an option to connect through an existing session?
- [ ] Test the behavior on session expiry (NOTE: a session expires after seven days. It is not possible to lower the expiry time. [See this Q&A](https://github.com/orgs/WalletConnect/discussions/1600#discussioncomment-4076993))
- [ ] Use Ecad Waku node by default in the dapp and add an option to use wc cloud service with a project id.
- [x] When we are on the test page of the dapp and reload it, we are brought back to the home page, the connection to the wallet is lost, and we need to connect again. See if we can stay on the same page without connecting the wallet again. The last active session should be used.
- [ ] Handle session request expiration; if the connection was never approved on the wallet side
- [x] Handle wallet deleted session. If the session is deleted in the wallet, the dapp should listen for the event and reset the dapp state, clean up from user session (for example, go back to the connect wallet page)
- [x] Handle wallet updated session; integrate the updated session state into the dapp state. Accounts might have changed; call `walletConnect2.getAccounts` and update the state of the app (selected account, available accounts, balance, )
- [ ] Add a ping button?

roxaneletourneau:
Closes #2145 
Closes #2146 

## Release Note Draft Snippet


packages/taquito-wallet-connect-2/src/taquito-wallet-connect-2.ts
```diff
@@ -0,0 +1,575 @@
...
+    }
+  }
+
+  private getTezosRequiredNamespace(): {
+    chains: string[];
+    methods: string[];
+    events: string[];
+  } {
+    if (
+      Object.prototype.hasOwnProperty.call(this.getSession().requiredNamespaces, TEZOS_PLACEHOLDER)
```
dsawali, line 492: is there a reason why we needed to do the prototype inheritance `.calll()` instead of referencing `this.session` directly?

roxaneletourneau: Made the change

BearCooder:
Hi @dsawali do you know who is reviewing this PR?

dsawali:
Hello @BearCooder, this PR is actually a work in progress. Myself and other Taquito team members are going to be reviewing it, but as it stands right now this is more of a PoC. There needs to be a lot more polish and updating for it to be ready for merge.

hui-an-yang:
wc2 package has been revamped and included in pr #3045

PR #2088, opened 2022-10-31
Configure NPM Workspaces 

danielelisi:
Ref #2127

This PR enables native integration of NPM Workspaces within Lerna. More context around the problem in the linked Issue.

Flipping `useWorkspaces` to `true` in `lerna.json` configuration causes all package dependencies to me downloaded to the repo's root `node_modules` folder and merges all sub-package `package-lock.json` files to a centralized `package-lock.json` also at the project's root. This changes requires a number of tweaks in other configuration files and tools we use in the repo:

 - Delete all `packages/**/package-lock.json`  files as they will not be used by NPM anymore.
 - Remove the `typeRoots` parameter from all `packages/**/tsconfig.prod.json` since now all dependencies are resolved by default from the root `node_modules`.
 - Use `npm -w <package_name>` to run NPM scripts in sub-packages instead of manually `cd`ing into folders.

The remaining changes are tech debts that failed to link local packages or caused CI to fail building Taquito Website and Taquito Test Dapp:
 - All Taquito dependencies in each sub-package are replaced with `*`. This allows Lerna to detect which dependency needs to be locally linked. For example, `@taquito/rpc` depends on `@taquito/http-utils`; Lerna will automatically symlink `@taquito/http-utils` to the repo local `packages/taquito-rpc` folder.
 - Replaced `docusaurus2-dotenv` with `docusaurus-plugin-dotenv` as the former isn't unmaintained and caused Webpack build errors when building the website.
 - Upgraded a number of dependencies in `taquito-test-dapp`, `website`, `taquito` to solve CICD build failures.
 - Fix in `packages/taquito-http-utils/src/taquito-http-utils.ts` was required for a lint failure.
 - Config changes in `packages/taquito/webpack.config.js` to support Webpack v4 to v5 upgrade in `package/taquito`.
 - Upgraded Github Actions to solve deprecation warnings in CICD pipelines
 - Use Node LTS versions 16` and `18` when building the repo in CICD. It's generally preferred to ensure that a project can build on LTS releases.


roxaneletourneau (commented):
The changes look good to me, but since there is this caching issue with Netlifly that makes the preview build fails, we should probably wait before merging

roxaneletourneau (approved):

Issue #2091, opened 2022-11-01
Integration test `contract-multisig-set-and-remove-delegate.spec.ts` hardcodes node RPC endpoint


arvidj:
**Description**

The integration test `contract-multisig-set-and-remove-delegate.spec.ts` hardcodes the node RPC endpoint to `http://mondaynet.ecadinfra.com:8732`. Consequently, this test can only run against mondaynet which is a problem if you want to run the test suite in other environments.

**Steps To Reproduce**

2. Run `TEZOS_RPC_MONDAYNET=http://127.0.0.1:18731 npm run "test:mondaynet-secret-key" "contract-security-non-existent-KT-address.spec.ts"`
3. Observe that it unsuccessfully attempts to contact `http://mondaynet.ecadinfra.com:8732`.

**Expected behavior**

This test should respect environment variable `TEZOS_RPC_MONDAYNET`.

PR #2092, opened 2022-11-01
contract-security-non-existent-KT-address: use rpc (#2091)

arvidj:
To test:

1. Start local, sandboxed Octez node with RPC endpoint `http://127.0.0.1:18731`.
2. Run `TEZOS_RPC_MONDAYNET=http://127.0.0.1:18731 npm run "test:mondaynet-secret-key" "contract-security-non-existent-KT-address.spec.ts"`
3. The test should work with this fix, and will fail without it as described in #2091


roxaneletourneau:
Thank you very much for the fix @arvidj !

PR #2095, opened 2022-11-01
2071 lima rpc changes

zainen:
this PR adds:

- the change for tokens_per_roll -> minimal_stake
- addition for the type of ticket_updates

tokens_per_roll remains on the `ConstantsResponseCommon` but is as minimal_stake for `ConstantsResponseProto015`

below is they type added for ticket_updates
TO ADD


packages/taquito-rpc/src/taquito-rpc.ts
```diff
@@ -467,7 +467,7 @@ export class RpcClient implements RpcClientInterface {
       'hard_gas_limit_per_operation',
       'hard_gas_limit_per_block',
       'proof_of_work_threshold',
-      'tokens_per_roll',
```
roxaneletourneau, line 470: I don't think `tokens_per_roll` should be removed here, it is causing a breaking change as the type is changed from BigNumber to string.

packages/taquito-rpc/src/types.ts
```diff
@@ -1499,6 +1500,23 @@ export interface ConstantsResponseCommon {
...
+  };
+  updates: {
+    account: string;
+    amount: BigNumber;
+  }
+}
+
+export interface ConstantsResponseProto015 extends ConstantsResponseProto014 {
+  minimal_stake: BigNumber;
+  ticket_updates: TicketUpdates[];
```
roxaneletourneau, line 1517: `ticket_updates` property is not in the correct place.

roxaneletourneau (changes requested):

packages/taquito-rpc/src/types.ts
roxaneletourneau: See https://gitlab.com/tezos/tezos/-/blob/master/docs/api/lima-openapi.json?plain=1
image

packages/taquito-rpc/src/types.ts
```diff
@@ -1499,6 +1500,23 @@ export interface ConstantsResponseCommon {
 
 export type Ratio = { numerator: number; denominator: number };
 
+export interface TicketUpdates {
+  ticket_token: {
+    ticketer: string;
+    content_type: MichelsonV1Expression;
+    content: MichelsonV1Expression;
+  };
+  updates: {
```
roxaneletourneau, line 1509: updates should be an array of object

packages/taquito-rpc/src/types.ts
```diff
@@ -1499,6 +1500,23 @@ export interface ConstantsResponseCommon {
...
+    account: string;
+    amount: BigNumber;
```
roxaneletourneau, line 1511: If amount is set as a BigNumber, it needs to be cast to that type in the corresponding rpc method. It is probably simpler to set it as a string.

packages/taquito-rpc/src/types.ts
```diff
@@ -1273,11 +1273,24 @@ export interface TezosGenericOperationError {
...
+    amount: string;
+  }[]
+}
+
 export interface OperationResultTransaction {
   status: OperationResultStatusEnum;
   storage?: MichelsonV1Expression;
   big_map_diff?: ContractBigMapDiff;
   balance_updates?: OperationBalanceUpdates;
+  ticket_updates?: TicketUpdates[]
```
roxaneletourneau, line 1293: Related to the issue discussed during the Proto-ecosystem (nov 15), this property is named `ticket_updates` or `ticket_receipt`:

- contents -> metadata -> operation_result -> ticket_updates
- contents -> metadata -> internal_operation_result -> [ ] ->  ticket_receipt

Both cases are present in this block: https://limanet.ecadinfra.com/chains/main/blocks/10150/

We need to decide how to handle this. From what I understand, `ticket_receipt` might be replaced by `ticket_updates` in the next protocol... Should we go ahead and override  `ticket_receipt` for `ticket_updates`?

And make sure we have unit tests for both cases.

roxaneletourneau (changes requested):

packages/taquito-rpc/test/taquito-rpc.spec.ts
```diff
@@ -622,6 +623,212 @@ describe('RpcClient test', () => {
...
+        preserved_cycles: 3,
+        blocks_per_cycle: 4096,
+        blocks_per_commitment: 32,
+        nonce_revelation_threshold: 256,
+        blocks_per_stake_snapshot: 256,
+        cycles_per_voting_period: 1,
+        hard_gas_limit_per_operation: new BigNumber(1040000),
+        hard_gas_limit_per_block: new BigNumber(5200000),
+        proof_of_work_threshold: new BigNumber(-1),
+        minimal_stake: "6000000000",
```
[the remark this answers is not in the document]
zainen: i had an error on this one consistently complaining about big number and said expected a string. wasnt sure why this one is the main different one.

packages/taquito-rpc/src/types.ts
zainen: added for the mean time ticket_receipt as the same type. so either could be accessed

we can remove the one that is removed when its done :D

packages/taquito-rpc/test/data/rpc-responses.ts
```diff
@@ -5175,3 +5175,225 @@ export const txRollupInbox = {
...
+                          "bytes": "00002a3c4ef8b90e40d0a0f3d793e78da4d40ff7ab05"
+                        },
+                        {
+                          "bytes": "016a41f23d91041e57a4cf22d0ebf27bba7bc67b2c00"
+                        }
+                      ]
+                    }
+                  ]
+                },
+                "ticket_receipt": [
```
roxaneletourneau, line 5370: `ticket_receipt` is not inside the `operation_result` property but the `internal_operation_result`. The mock data should represent exactly was is returned by the RPC. See:
image

packages/taquito-rpc/test/taquito-rpc.spec.ts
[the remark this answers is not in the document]
roxaneletourneau: Should be `minimal_stake: new BigNumber("6000000000")`

packages/taquito-rpc/src/types.ts
```diff
@@ -1273,11 +1273,26 @@ export interface TezosGenericOperationError {
...
+}
+export type TicketReceipt  = TicketUpdates
+
 export interface OperationResultTransaction {
   status: OperationResultStatusEnum;
   storage?: MichelsonV1Expression;
   big_map_diff?: ContractBigMapDiff;
   balance_updates?: OperationBalanceUpdates;
+  ticket_updates?: TicketUpdates[];
+  ticket_receipt?: TicketReceipt[];
```
roxaneletourneau, line 1295: I am not sure that showing the 2 properties is the best solution for the users, as we know that `ticket_receipt` will be removed in the protocol M. Let have a discussion about that.

roxaneletourneau (changes requested):

integration-tests/tx-rollup-l2-address.spec.ts
```diff
@@ -59,7 +59,7 @@ CONFIGS().forEach(({lib, setup, protocol}) => {
       const contract = await op.contract();
       const symbolReturn = await contract.storage()
       expect(JSON.stringify(symbolReturn)).toEqual(JSON.stringify(UnitValue))
-      const methodCall = await contract.methods.default('txr1di1E5BgY6QSPx61eE62P7KW93ckGwH2e7', 'tz4VHgLiRx5ZZjwU2QaybHc11EMJk3NcyvVc', '1').send();
+      const methodCall = await contract.methods.default(txRollupAddress, 'tz4VHgLiRx5ZZjwU2QaybHc11EMJk3NcyvVc', '1').send();
```
dsawali, line 62: small nitpick, but the variable name `methodCall` isn't very descriptive. Consider renaming it to something that reflects what the contract is doing. Like `ticketDeposit` or `deposit` or something similar

packages/taquito-rpc/test/taquito-rpc.spec.ts
```diff
@@ -3215,6 +3424,33 @@ describe('RpcClient test', () => {
...
+      const content = response.operations[0][0].contents[0] as OperationContentsAndResultTransaction
+      const ticketUpdates = content.metadata.operation_result.ticket_updates && content.metadata.operation_result.ticket_updates[0]
+
+      expect(ticketUpdates).toBeDefined();
+      expect(ticketUpdates?.ticket_token.ticketer).toEqual('KT1JGcC8DuWHcShu6XvtfgKVnV2zcYsZ4TVH')
+      expect(ticketUpdates?.updates[0].account).toEqual('KT1JoRgUcR6NApwMLnBZ2pehCzp8tR4HtkHj')
+      expect(ticketUpdates?.updates[0].amount).toEqual('-2')
+      done()
+    });
+    // may be removed
```
dsawali, line 3440: could you explain a bit about the background of this comment?

integration-tests/contract-with-txr1-address.spec.ts
```diff
@@ -78,7 +76,7 @@ CONFIGS().forEach(({ lib, setup }) => {
       expect(newMapStorage.addressMap.get('txr1bZx7qro4xNsxLhmpXfmzQojHkg8XqrUvL')).toBeUndefined();
       done();
     })
-    it('should pass with KT1', async (done) => {
+    test('should pass with KT1', async (done) => {
```
dsawali, line 79: we should use `it` for consistency

dsawali (changes requested):

integration-tests/tx-rollup-l2-address.spec.ts
```diff
@@ -82,13 +82,13 @@ CONFIGS().forEach(({lib, setup, protocol}) => {
...
-      await methodCall.confirmation();
-      expect(methodCall.includedInBlock).toBeLessThan(Number.POSITIVE_INFINITY)
+      const ticketDeposit = await contract.methods.default(pkh).send();
+      await ticketDeposit.confirmation();
+      expect(ticketDeposit.includedInBlock).toBeLessThan(Number.POSITIVE_INFINITY)
       done();
     })
   })
-  describe(`mondaynet contract originations and method calls to test the type of tx_rollup_l2_address`, () => {
+  describe(`it contract originations and method calls to test the type of tx_rollup_l2_address`, () => {
```
roxaneletourneau, line 91: Is there a difference between these tests and the precedent, it seems to be duplication.

integration-tests/rpc-get-protocol-constants.spec.ts
```diff
@@ -617,7 +617,7 @@ CONFIGS().forEach(({ lib, protocol, rpc }) => {
         hard_gas_limit_per_operation: new BigNumber(1040000),
         hard_gas_limit_per_block: new BigNumber(5200000),
         proof_of_work_threshold: new BigNumber(-1),
-        minimal_stake: "6000000000",
+        minimal_stake: '6000000000',
```
roxaneletourneau, line 620: Failure in the integration test, it is a bigNumber

roxaneletourneau (changes requested):

integration-tests/rpc-get-protocol-constants.spec.ts
zainen: sorry must have been reverted when rebased

roxaneletourneau (approved):

PR #2105, opened 2022-11-07
Deprecate Tezbridge

egarson:
This PR is to deprecate Tezbridge from the codebase, as per https://github.com/ecadlabs/taquito/tree/2080-deprecate-taquito-tezbridge.

It only removes code and modifies documentation. One small merge conflict with website/sidebars.js was resolved.

I believe this is slated for Lima, parking here for now.


roxaneletourneau (approved):
Lgtm :)

PR #2107, opened 2022-11-07
1630 ballot

### LINKED ISSUE #1630: Add support for the Ballot operation
Ballot is supported on the local-forging package, but not on the high-level APIS.
The ballot operation allows delegates to vote for a proposal using yay, nay or pass during the exploration or promotion phases.

We need to add support for the operation on the contract, estimate, batch.

dsawali:
closes #1630 

## Release Note Draft Snippet


dsawali:
@nicolasochem thanks, we'll look into that! We also have a proposals operation PR coming in. Perhaps we can test both at the same time.

egarson:
Why do we have a yarn.lock? I thought we had coalesced on npm?

dsawali:
> Why do we have a yarn.lock? I thought we had coalesced on npm?

@egarson It was part of legacy where we used to use yarn (before even I joined). We probably need to trim that out eventually, but it's staying in for now just in case it gets mad and breaks

packages/taquito-rpc/src/types.ts
```diff
@@ -182,7 +182,7 @@ export interface InlinedPreEndorsement {
   signature?: string;
 }
 
-export type OperationContentsBallotEnum = 'nay' | 'yay' | 'pass';
+export type BallotEnum = 'nay' | 'yay' | 'pass';
```
zainen, line 185: nit should we rename since its not an actual enum?

packages/taquito/src/operations/ballot-operation.ts
```diff
@@ -0,0 +1,43 @@
...
+    context: Context
+  ) {
+    super(hash, raw, results, context);
+  }
+
+  get operationResults() {
+    const ballotOp =
+      Array.isArray(this.results) &&
+      (this.results.find((op) => op.kind === 'ballot') as OperationContentsAndResultBallot);
+    const result = ballotOp;
```
zainen, line 27: nit could these two lines be combined?

packages/taquito/src/operations/operation-emitter.ts
```diff
@@ -81,17 +83,33 @@ export abstract class OperationEmitter {
...
+      if (op.kind === 'ballot') {
+        try {
+          currentVotingPeriodPromise = this.rpc.getCurrentPeriod();
+        } catch (e) {
+          throw new RPCResponseError(
+            `Failed to get the current voting period index: ${JSON.stringify(e)}`
+          );
+        }
+      }
+    });
```
zainen, line 99: what is the return of this.rpc.getCurrentPeriod() when there is no period? 
could this be simplified to `let currentVotingPeriod = op.find.... -> return this.rpc.getCurrentPeriod()? 

just was odd seeing the promise.resolve(undefined)

packages/taquito/src/operations/operation-emitter.ts
```diff
@@ -81,17 +83,33 @@ export abstract class OperationEmitter {
...
         break;
       }
     }
 
-    const [hash, protocol, headCounter] = await Promise.all([
+    const [hash, protocol, headCounter, currentVotingPeriod] = await Promise.all([
       blockHashPromise,
       blockProtoPromise,
       counterPromise,
+      currentVotingPeriodPromise,
```
zainen, line 112: my own learning

is this a more efficient way of awaiting multiple values? that you sort them all into a promise all and grab the values needed?

zainen (commented):
just a few questions about some parts

is there a doc on how to create a proposal? 

im assuming we wont have integration tests for this since it involves having the create a proposal first

packages/taquito-rpc/src/types.ts
dsawali: it's still an enum just without an initializer

packages/taquito/src/operations/ballot-operation.ts
dsawali: I separated those 2 for readability to make the pattern a bit clearer

packages/taquito/src/operations/operation-emitter.ts
dsawali: a node will always have a voting period, but the undefined is a way to catch cases where, say, the node doesn't respond correctly.

packages/taquito/src/operations/operation-emitter.ts
dsawali: kind of, it's just an older way to handle asynchronous data before async/await. In this case I stuck to the pattern that's already available.

packages/taquito-rpc/src/types.ts
dsawali: sure, do you have a name in mind?

dsawali:
> just a few questions about some parts
> 
> is there a doc on how to create a proposal?
> 
> im assuming we wont have integration tests for this since it involves having the create a proposal first

@zainen there's this document on amendment and voting: https://tezos.gitlab.io/lima/voting.html?highlight=amendment%20voting

but if you're talking about to create a proposal in Taquito, it will be part of the next PR.

And yes, the integration test will be written after this PR and another one gets merged

packages/taquito-rpc/src/types.ts
zainen: the three that come to mind are BallotVote, VotingOptions, BallotVotingOptions just a little more descriptive

zainen (approved):
LGTM :D

packages/taquito-rpc/src/types.ts
dsawali: I like BallotVote as well, I'll update it

PR #2125, opened 2022-11-16
2087 ballot test

### LINKED ISSUE #2087: Integration test for the ballot operation
Create an integration test for the ballot operation using flextesa to ensure regression/breaking changes can be caught in the future regarding this operation.

dsawali:
closes #2087 

- Added test for Ballot op and Proposals op
- Updated main.yml Github workflow with Flextesa bootstrap account secret key
- Fixed `contract-multisig-set-and-remove-delegate.spec.ts` regression due to different Flextesa bootstrap account being used

## Release Note Draft Snippet


integration-tests/sandbox-ballot-operation.spec.ts
```diff
@@ -0,0 +1,49 @@
+import { CONFIGS, sleep } from './config';
+
+CONFIGS().forEach(async ({ lib, protocol, setup, rpc }) => {
+  const Tezos = lib;
+  const flextesanet = (rpc === 'http://0.0.0.0:20000') ? it : it.skip;
+
+  Tezos.setRpcProvider(rpc);
```
roxaneletourneau, line 7: I don't think this line is required as it is done when calling setup bellow.

integration-tests/sandbox-ballot-operation.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+      try {
+        const period = await Tezos.rpc.getCurrentPeriod();
+        await sleep((period.remaining + 5)  * 1000);
+
+        const proposalsOp = await Tezos.contract.proposals({
+          proposals: ['ProtoALphaALphaALphaALphaALphaALphaALphaALphaDdp3zK']
+        });
+
+        await proposalsOp.confirmation();
+      } catch(e) {
```
roxaneletourneau, line 22: For both tests, after the confirmation, can you verify that `op.status `is `applied `and `hash` is defined?

roxaneletourneau (changes requested):

integration-tests/sandbox-ballot-operation.spec.ts
dsawali: for this operation there is no `status` in `operationResults`

integration-tests/contract-multisig-set-and-remove-delegate.spec.ts
```diff
@@ -1,13 +1,22 @@
 import { CONFIGS } from "./config";
-import { MANAGER_LAMBDA } from "@taquito/taquito";
+import { MANAGER_LAMBDA, TezosToolkit  } from "@taquito/taquito";
 import { genericMultisig } from "./data/multisig";
 
 CONFIGS().forEach(({ lib, rpc, setup, createAddress }) => {
-  const Tezos = lib;
- 
+  const Funder = lib;
```
zainen, line 6: maybe add a not for why its called Funder? difference from the norm of other tests

zainen (commented):
lgtm nice work!
just one question about naming for a little more clarity for why to use the toolkit from configs rather than creating a new one. 
pretty self explanatory but may be good for anyone new coming to look at it

roxaneletourneau (approved):

integration-tests/contract-multisig-set-and-remove-delegate.spec.ts
dsawali: it's called Funder since it isn't actually signing anything. There are a few flextesa accounts that is provided by the sandbox. In this case, we only use the `flextesa_bootstrap` account to 'fund' the signer account we're using. 

It's purpose is to only fund the account to the signer TezosToolkit instance, which is still named `Tezos`

PR #2151, opened 2022-11-24
feat(rpc): Adds method for fetching block events

RomarQ:
Currently, taquito only supports subscribing to new events and does not provide a way for fetching past events. This PR adds a new method for fetching the events contained in a given block.


roxaneletourneau (changes requested):
Hi @RomarQ,
Thank you for your contribution. 
We describe the RPC package as a one-to-one mapping to the corresponding Tezos RPC API endpoints. However, the `getBlockEvents` method doesn't follow this structure that we want to keep. 
What we suggest instead of having a `getBlockEvents` method is to provide a "utils" function that can be applied to the `getBlock` response.
Please let us know what you think!

RomarQ:
Hi @roxaneletourneau
The one-to-one mapping on the RPC package makes sense. 
For now, I will just close the PR. Hopefully, a new RPC will be added to octez for explicitly querying the block events `chains/<chain>/blocks/<block>/events`.

PR #2161, opened 2022-11-28
add try catch to show error as well as mnemonic

zainen:
add catch for failure to see the mnemonic when method fails

### DIFF
--- a/integration-tests/fromMnemonic-send-tez.spec.ts
+++ b/integration-tests/fromMnemonic-send-tez.spec.ts
@@ -16,67 +16,79 @@ CONFIGS().forEach(({ lib, rpc, setup }) => {
       done();
     });
     it('Should create an InMemorySigner (ed25519) with the fromMnemonic method and transfer tez to an account', async (done) => {
-      const Tezos = new TezosToolkit(rpc);
-      // with all default values
-      const signer = InMemorySigner.fromMnemonic({ mnemonic });
-      Tezos.setSignerProvider(signer);
+      try {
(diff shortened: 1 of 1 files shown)


integration-tests/fromMnemonic-send-tez.spec.ts
```diff
@@ -16,67 +16,79 @@ CONFIGS().forEach(({ lib, rpc, setup }) => {
...
-      expect(status).toEqual('applied');
+        const status = await returnOp.status();
+        expect(status).toEqual('applied');
 
-      const balance = await Tezos.tz.getBalance(tezosPKH);
-      expect(Number(balance)).toBeGreaterThan(0);
+        const balance = await Tezos.tz.getBalance(tezosPKH);
+        expect(Number(balance)).toBeGreaterThan(0);
+      } catch (err) {
+        console.log({ err, mnemonic });
```
roxaneletourneau, line 40: Should the error be re-thrown after the console.log to make sure that the test fails in the pipeline and that we have visibility?

dsawali: I agree, I think we should also throw an error alongside the console.log.

If anything, it's probably better for us to throw an error with the error message instead of outputting using console.log

zainen: right for some reason i thought the error would still be thrown but thrown while logging. 

switch to throw the error plus the mnemonic

roxaneletourneau (approved):

Issue #2075, opened 2022-10-25
Michel Codec - Support LAMBDA_REC + fix TICKET instructions


roxaneletourneau:
**Is your feature request related to a problem and use case? Please describe.**
1 - Add support to properly validate the code of a contract having the `LAMBDA_REC` instruction in the michel-codec package.

2- TICKET instruction now returns option ticket 'a instead of ticket 'a

**Describe the solution you'd like**
1 - Add a case for the `LAMBDA_REC` instruction in the `retStack` constant: https://github.com/ecadlabs/taquito/blob/master/packages/taquito-michel-codec/src/michelson-typecheck.ts#L889

>  LAMBDA_REC 'a 'b code: Push a lambda with itself on top of the
>  code, recursively, with the given parameter type 'a and return a value of type   'b onto the stack (if it terminates before gas exhaustion).
> 
> ```
> :: 'A ->  (lambda 'a 'b) : 'A
>    iff code::'a: (lambda 'a 'b):[] -> 'b:[]
> 
> > LAMBDA_REC 'a 'b code / S  => {LAMBDA_REC 'a 'b code; code} : S
> ```

2- Add a condition for the `TICKET` instruction in the `retStack` constant to ensure that an optional ticket is returned for the Lima protocol.

Add relevant unit tests. Consider creating a contracts_015 folder containing all tests for the L protocol (as is the case for the other protocols already).

Contract examples:
- https://gitlab.com/tezos/tezos/-/blob/master/src/proto_015_PtLimaPt/lib_protocol/test/integration/michelson/contracts/fail_rec.tz
- https://gitlab.com/tezos/tezos/-/blob/master/src/proto_015_PtLimaPt/lib_protocol/test/regression/contracts/rec_id_unit.tz
- https://gitlab.com/tezos/tezos/-/blob/master/src/proto_015_PtLimaPt/lib_protocol/test/integration/michelson/contracts/omega.tz
- https://gitlab.com/tezos/tezos/-/blob/master/src/proto_015_PtLimaPt/lib_protocol/test/integration/michelson/contracts/rec_fact_apply.tz
- https://gitlab.com/tezos/tezos/-/blob/master/src/proto_alpha/lib_protocol/test/integration/michelson/contracts/rec_fact_store.tz
- https://gitlab.com/tezos/tezos/-/blob/master/src/proto_015_PtLimaPt/lib_protocol/test/integration/michelson/contracts/rec_fact.tz
[...]

PR #2175, opened 2022-11-30
2075 lambda rec michel codec

zainen:
add LAMBDA_REC to typecheck file running tests

adjust TICKET to be protocol sensitive
adding `option ticket 'a` for lima

lost of lines changed form formatting a file adjusting spaceing

closes - #2075


packages/taquito-michel-codec/src/michelson-typecheck.ts
```diff
@@ -1847,10 +1851,22 @@ function functionTypeInternal(
       case 'TICKET': {
         const s = args(0, null, ['nat'])[0];
         ensureComparableType(s);
-        return [
-          annotate({ prim: 'ticket', args: [s] }, instructionAnn({ t: 1, v: 1 })),
-          ...stack.slice(2),
-        ];
+        if (ctx?.protocol === Protocol.PtLimaPtL) {
```
roxaneletourneau, line 1854: Could you revert the if-else statement as follow:
```
if (ProtoInferiorTo(proto, Protocol.PtLimaPtL) {
         return [
            annotate({ prim: 'ticket', args: [s] }, instructionAnn({ t: 1, v: 1 })),
            ...stack.slice(2),
          ];
        } else {
return [
            annotateVar({
              prim: 'option',
              args: [
                annotate({ prim: 'ticket', args: [s] }, instructionAnn({ t: 1, v: 1 }))
              ]
            }),
            ...stack.slice(2)
          ];
        }

That way, we don't have to update for upcoming protocol, and everything will work with alpha (mondaynet)

File: packages/taquito-michel-codec/src/michelson-types.ts [hunk 2]
```diff
@@ -225,6 +226,7 @@ export interface MichelsonTypeContract<T extends MichelsonType> extends TypeX<'c
 export interface MichelsonTypeOr<T extends [MichelsonType, MichelsonType]> extends TypeX<'or', T> {}
 export interface MichelsonTypeLambda<Arg extends MichelsonType, Ret extends MichelsonType>
   extends TypeX<'lambda', [Arg, Ret]> {}
+
```
roxaneletourneau, line 229: There is a `data` section in this file, I think we need `Lambda_rec` in `MichelsonDataID` at least (maybe in other places too).

roxaneletourneau: @e-asphyx can you validate?

packages/taquito-michel-codec/src/michelson-typecheck.ts
zainen: right! i saw that function but forgot to use it! thanks

packages/taquito-michel-codec/src/michelson-types.ts
```diff
@@ -321,6 +324,7 @@ type DataList<T extends MichelsonData[]> = T & Node;
 export type MichelsonDataPair<T extends MichelsonData[]> = DataX<'Pair', T> | DataList<T>;
 export type MichelsonMapElt = PrimX<'Elt', [MichelsonData, MichelsonData]>;
 export type MichelsonMapEltList = List<MichelsonMapElt>;
+export type MichelsonLambdaRec = DataX<'Lambda_rec', [MichelsonLambdaRec]>
```
zainen, line 327: wondering if this makes sense since it wouldnt be MichelsonData right? it would be itself recursively

roxaneletourneau: I don't think `MichelsonLambdaRec` should be inside `DataX`, but a list of Michelson instructions (there is probably an existing type for that).

packages/taquito-michel-codec/src/binary.ts
```diff
@@ -186,6 +186,7 @@ const primitives: PrimID[] = [
   'constant',
   'SUB_MUTEZ',
   'EMIT',
+  'LAMBDA_REC',
```
roxaneletourneau, line 189: Lambda_rec is missing

roxaneletourneau: Fixed it in my PR, because it was needed for pack/unpack

roxaneletourneau (approved):

PR #2178, opened 2022-12-01
fix: force lerna to npm clean install

RichAyotte:
## Summary
Lerna runs `npm install` by default in a non-ci environment. This is problematic for local environments as `npm install` could alter the `package-lock.json` file. This change forces lerna to run `npm clean-install` in a local environment.

refs: https://github.com/lerna/lerna/tree/main/commands/bootstrap#--no-ci

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   "scripts": {
     "build": "nx run-many --target=build --exclude=@taquito/website,taquito-test-dapp-vite",
     "build-test-dapp": "lerna run build --scope=taquito-test-dapp-vite",
-    "bootstrap": "lerna bootstrap --force-local",
+    "bootstrap": "lerna bootstrap --force-local --ci",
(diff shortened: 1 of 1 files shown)


egarson:
Thanks, we are looking into this.

RichAyotte:
There's probably a better way. Running npm ci multiple times is a bit of a waste. Maybe mom ci shouldn't automatically call lerna?

danielelisi:
Hi @RichAyotte thanks for the PR and the suggestion!

I checked Lerna's docs, and they recommend deprecating `lerna bootstrap`  and letting npm workspaces handle the module installation and local package linking. 

https://lerna.js.org/docs/features/legacy-package-management

Since we're removing references to `lerna bootstrap` command from the repo I'll close this PR

RichAyotte:
Nice, that's even better.

RichAyotte:
Fixed in https://github.com/ecadlabs/taquito/pull/2189

PR #2183, opened 2022-12-02
2068 support drain delegate operation in contract api

hui-an-yang:

## Release Note Draft Snippet


packages/taquito/src/contract/rpc-contract-provider.ts
```diff
@@ -60,7 +62,7 @@ import { TxRollupOriginationOperation } from '../operations/tx-rollup-originatio
 import { TxRollupBatchOperation } from '../operations/tx-rollup-batch-operation';
 import { TransferTicketOperation } from '../operations/transfer-ticket-operation';
 import { IncreasePaidStorageOperation } from '../operations/increase-paid-storage-operation';
-import { BallotOperation } from '../operations';
+import { BallotOperation, DrainDelegateOperation } from '../operations';
```
dsawali, line 65: I think we should separate the imports like the others. I think I've made a mistake on the BallotOperation import path. It should be its own line for clarity.

integration-tests/sandbox-drain-delegate-operation.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+    })
+    flextesanet('Should be able to inject drain_delegate operation', async (done) => {
+      const drainOp = await Destination.contract.drainDelegate({
+        consensus_key: destinationPkh,
+        delegate: delegatePkh,
+        destination: destinationPkh,
+      });
+      await drainOp.confirmation();
+      
+      expect((await Delegate.rpc.getBalance(delegatePkh)).toNumber()).toEqual(0);
```
dsawali, line 51: could you add more assertions to check whether the op has been injected in the block?

i.e. `includedInBlock`, `status` or something similar?

integration-tests/sandbox-drain-delegate-operation.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+      expect((await Destination.rpc.getBalance(destinationPkh)).toNumber()).toBeGreaterThan(0);
```
dsawali, line 52: if balance of the previous baker was `x` and the balance of the destination was `y`, we should check whether the destinationPkh balance is now `x+y`

integration-tests/sandbox-drain-delegate-operation.spec.ts
```diff
@@ -0,0 +1,56 @@
+import { TezosToolkit } from "@taquito/taquito";
+import { CONFIGS, sleep } from "./config";
+
+CONFIGS().forEach(({ lib, rpc, setup, createAddress }) => {
+  const Tezos = lib;
+  const flextesanet = (rpc === 'http://localhost:20001') ? it : it.skip;
```
dsawali, line 6: we could maybe specify which protocol it is by renaming this to `flextesaLimanet`

dsawali (changes requested):

integration-tests/sandbox-env-lima.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+export RUN_LIMANET_WITH_SECRET_KEY=true
```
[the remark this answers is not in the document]
hui-an-yang: Didn't know it's the common understanding, will change it back

packages/taquito/src/contract/rpc-contract-provider.ts
hui-an-yang: Will update both, was wondering do we export them in `operations/index.ts` only for end users not for code base?

packages/taquito/src/contract/interface.ts
```diff
@@ -221,6 +223,16 @@ export interface ContractProvider extends StorageProvider {
    */
   txRollupSubmitBatch(params: TxRollupBatchParams): Promise<TxRollupBatchOperation>;
 
+  /**
+   *
+   * @description Submit drain delegate for an ongoing proposal
```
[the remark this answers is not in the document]
hui-an-yang: sorry, it'll be updated

integration-tests/sandbox-drain-delegate-operation.spec.ts
hui-an-yang: thanks, updated to flextesaLimaNet

integration-tests/sandbox-drain-delegate-operation.spec.ts
hui-an-yang: will add includedInBlock to beDefined

dsawali:
I think we should leverage Dan's PR to use multiple bootstrap accounts. That way we remove the risk of failures in other tests due to the drain delegate op. PR : https://github.com/ecadlabs/taquito/pull/2219

packages/taquito/src/contract/interface.ts
```diff
@@ -221,6 +223,16 @@ export interface ContractProvider extends StorageProvider {
    */
   txRollupSubmitBatch(params: TxRollupBatchParams): Promise<TxRollupBatchOperation>;
 
+  /**
+   *
+   * @description Submit drain delegate operation
```
zainen, line 228: Submits a drain delegate operation i think is a bit more proper

integration-tests/sandbox-drain-delegate-operation.spec.ts
```diff
@@ -0,0 +1,60 @@
+import { TezosToolkit } from "@taquito/taquito";
+import { CONFIGS, sleep } from "./config";
+
+CONFIGS().forEach(({ lib, rpc, setup, createAddress }) => {
+  const Tezos = lib;
+  const flextesaLimaNet = (rpc === 'http://localhost:20000') ? it : it.skip;
```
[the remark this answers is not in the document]
hui-an-yang: Think you have a point, can see how `flextesanet` will be less coding debt when we adapt into Lima

packages/taquito/src/contract/interface.ts
hui-an-yang: thanks, will push a new commit

docs/consensus_key.md
```diff
@@ -20,3 +20,20 @@ await op.confirmation();
 ```
 - `pk` is the public key you want the consensus key to point to
 
+
+## Drain Delegate
+This is an operation that must be signed by the active consensus key `consensus_pkh` of the baker `baker_pkh`. This operation immediately transfers all the spendable balance of the `baker_pkh`’s implicit account into the `destination_pkh` implicit account. It has no effect on the frozen balance. This operation is included in pass 2 (anonymous operations). So drain operations don’t compete with regular manager operations for gas and block size quota; the 1M restriction (one-operation-per-manager-per-block) applies to drain operations as well, meaning that a drain for a baker and a transfer operation from the same baker are in conflict. As an incentive for bakers to include drain operations, a fixed fraction of the drained baker’s spendable balance is transferred as fees to the baker that includes the operation, i.e. the maximum between 1tz or 1% of the spendable balance.
```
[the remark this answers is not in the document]
hui-an-yang: Thank you, with this doc line 6 Davis have left reference to general consensus_key release note, is it still better to add reference for drain_delegate op part?

packages/taquito/src/operations/drain-delegate-operation.ts
```diff
@@ -0,0 +1,45 @@
...
+    private readonly params: RPCDrainDelegateOperation,
+    raw: ForgedBytes,
+    results: OperationContentsAndResult[],
+    context: Context
+  ) {
+    super(hash, raw, results, context);
+  }
+
+  get operationResults() {
+    const DrainDelegateOp =
```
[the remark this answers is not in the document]
hui-an-yang: pushed a commit to address this

roxaneletourneau (approved):
Lgtm  :)

dsawali (approved):

PR #2204, opened 2022-12-03
Delete redundant package-lock.json files from Taquito packages

danielelisi:
After the introduction of NPM Workspaces in #2088 NPM now only manages a centralized `package-lock.json` at the root of the repo. All the other `package-lock.json` file in sub packages (`integration-tests`, `website`, `packages/taquito*`) are redundant and are not taken into consideration when installing package dependencies.

I meant to remove these in a previous PR #2088 since I mention this change in the description; however I must have made a bad PR merge at some point which reintroduced them by mistake.


egarson:
Good catch.

PR #2221, opened 2022-12-08
Run integration test against testnets for external PRs

danielelisi:
External PRs from users outside our `ecadlabs` GH org don't have access to GH Secrets. This is an intended behaviour to avoid exposing sensitive keys and tokens to third parties.

We block some resource-intensive RPC endpoints from our public-facing nodes. Since some integration tests need to contact these RPC endpoints we let GH runners join our private network and contact internal RPC Tezos nodes directly.  GH Runners use GH Secrets to join our Tailscale private network.

This PR refactors the GH workflows and skips the `baking_rights` and `endorsing_rights` tests when an external user submits a PR to Taquito. Additionally, we set the keygen `keyURL` to point to the publicly accessible keygen instance so that we can request keys without having to login to our private network. 

I've also added a condition to skip the `deploy-edge-package` workflow on external PRs. Similarly to what I mentioned above, Github Secrets aren't available on external PRs and this job would always fail as it requires a secret to publish the packages on Verdaccio. Also, this is a security consideration as we don't want to allow external users to arbitrarily publish packages to an NPM registry.

## Test Plan

Opened a PR as an external contributor and confirmed that:
 - the Tailscale job is skipped
 - Integration test the public RPC endpoint

Example workflow run from an external PR https://github.com/ecadlabs/taquito/pull/2223/checks

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -53,7 +53,9 @@ jobs:
       with:
         node-version: 16
     # TODO: remove after debugging
-    - run: echo "${{ github.event }}"
+    - env:
+        GITHUB_CONTEXT: ${{ toJson(github) }}
+      run: echo "$GITHUB_CONTEXT"
     - run: npm ci
     - run: npm run build
     - if: ${{ !github.event.pull_request.head.repo.fork }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -67,7 +67,8 @@ jobs:
     - run: npm -w integration-tests run test:${{ matrix.testnet }} -- --maxWorkers=8
       env:
         # Ternary operator workaround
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -42,22 +42,32 @@ jobs:
...
+          - testnet: kathmandunet
+            testnet_uppercase: KATHMANDUNET
+          - testnet: limanet
+            testnet_uppercase: LIMANET
     steps:
     - uses: actions/checkout@v3
     - uses: actions/setup-node@v3
       with:
         node-version: 16
+    # TODO: remove after debugging
```
roxaneletourneau, line 55: Should this be deleted?

danielelisi: ah thanks for catching it! removed the debug statements

roxaneletourneau (approved):

Issue #1762, opened 2022-07-22
`MapTypecheckError` on `storage()` for contract `KT1UwCd8rNCATj2mwC5UgEo4uqPRVqfjRPa6` (ghostnet)


psmolinsky:
**Description**
Retrieving storage of contract `KT1UwCd8rNCATj2mwC5UgEo4uqPRVqfjRPa6` on ghostnet fails with `MapTypecheckError`.

**Steps To Reproduce**

1. Get the contract abstraction.
  ```typescript
  const toolkit = new TezosToolkit('ghostnet-node-url');
  const contract = await toolkit.contract.at('KT1UwCd8rNCATj2mwC5UgEo4uqPRVqfjRPa6');
  ```
2. Retrieve storage of the contract.
  ```typescript
  await contract.storage();
  ```
3. It fails.
    ```
    MapTypecheckError: value not compliant with underlying michelson type
        at MichelsonMap.assertTypecheckValue (...\node_modules\@taquito\michelson-encoder\src\michelson-map.ts:110:13)    
        at MichelsonMap.set (...\node_modules\@taquito\michelson-encoder\src\michelson-map.ts:166:10)
        at ...\node_modules\@taquito\michelson-encoder\src\tokens\map.ts:50:11
        at Array.forEach (<anonymous>)
        at MapToken.Execute (...\node_modules\@taquito\michelson-encoder\src\tokens\map.ts:49:9)
        at Schema.Execute (...\node_modules\@taquito\michelson-encoder\src\schema\storage.ts:143:31)
        at RpcContractProvider.<anonymous> (...\node_modules\@taquito\taquito\src\contract\rpc-contract-provider.ts:85:27)
        at Generator.next (<anonymous>)
        at fulfilled (...\node_modules\@taquito\taquito\dist\taquito.umd.js:41:62)
    ```

roxaneletourneau:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Taquito-Development-Team-60ca841c87ca580010057707/issues/ecadlabs/taquito/1762?planning-poker) with ZenHub @dsawali @hui-an-yang @zainen

psmolinsky:
[...]

PR #2228, opened 2022-12-14
fix(michelson-encoder): fix for MapTypecheckError

roxaneletourneau:
Quick fix for the method Schema.Typecheck that fails to parse map when its value is of type big_map. Note that it is not the first bug happening with the Typecheck method: using "EncodeObject" to do the validation is not always appropirate. I think this part of the code should be rework in the futur.

re #1762

## Release Note Draft Snippet


zainen:
this was merged in with my pr from the rebase on top of this

PR #2238, opened 2022-12-19
1993 fix duplicate errors

### LINKED ISSUE #1993: Consolidate duplicated Errors across packages
_Follow up of https://github.com/ecadlabs/taquito/issues/160_ as described in https://hackmd.io/@tca0T8LZT3au8LG6hSX20g/r1zTRTlfj_

Task 2: Consolidate duplicated Errors across packages (InvalidViewParameterError, InvalidAddressError, InvalidBlockHashError, InvalidHexStringError, InvalidMessageError, ...)

zainen:
moves duplicates to taquito/core

two changed class names for clairty since joining the classes would have obscured the reason error and the name of the error

new classes 
InvalidContractViewParameterError (previously InvalidViewParameterError) -> may want to change the other as its in tzip16 and would be less disruptive than changing whats in `@taquito`
InvalidHashLengthError (previously InvalidMessageError)

moved other duplicates with possible minor change of adding either a message or param to the constructor based on needs

will merge into branch [1992-create-high-level-categories-of-errors](https://github.com/ecadlabs/taquito/tree/1992-create-high-level-categories-of-errors) if not merged first 

closes #1993


zainen:
missed updating a test. will update

packages/taquito-contracts-library/src/errors.ts
```diff
@@ -1,12 +1,4 @@
-/**
- *  @category Error
- *  @description Error that indicates an invalid address being used or passed
- */
-export class InvalidAddressError extends Error {
-  constructor(message: string) {
-    super(message);
-  }
-}
+export * from '@taquito/core';
```
hui-an-yang, line 1: this file path seem renamed to `packages/taquito-contracts-library/src/taquito-core.ts` ?

packages/taquito-ledger-signer/src/taquito-ledger-signer.ts
```diff
@@ -43,19 +44,6 @@ export class InvalidDerivationTypeError extends Error {
   }
 }
```
dsawali, line 45: shouldn't `InvalidDerivationTypeError` be moved to error.ts as well?

packages/taquito-local-forging/src/taquito-local-forging.ts
```diff
@@ -3,13 +3,14 @@
...
 
+import { InvalidBlockHashError } from './error';
 import { ForgeParams, Forger } from './interface';
 import { CODEC } from './constants';
 import { decoders } from './decoder';
 import { encoders } from './encoder';
 import { Uint8ArrayConsumer } from './uint8array-consumer';
 import { validateBlock, ValidationResult, InvalidOperationKindError } from '@taquito/utils';
-import { InvalidBlockHashError, InvalidOperationSchemaError } from './error';
+import { InvalidOperationSchemaError } from './error';
```
dsawali, line 13: this and `InvalidBlockHashError` in line 6 seems like they're from the same file. Are there any reason they are on different import statements?

packages/taquito-local-forging/src/michelson/codec.ts
```diff
@@ -135,8 +135,8 @@ export const intEncoder: Encoder<IntValue> = ({ int }) => {
     binary.length <= 6
       ? 6
       : (binary.length - 6) % 7
-      ? binary.length + 7 - ((binary.length - 6) % 7)
-      : binary.length;
+        ? binary.length + 7 - ((binary.length - 6) % 7)
```
dsawali, line 138: is this from the lint, or is the whitespace unintended?

packages/taquito-local-forging/src/uint8array-consumer.ts
```diff
@@ -8,11 +9,11 @@ export class Uint8ArrayConsumer {
       );
       return new Uint8ArrayConsumer(arr);
     } else {
-      throw new InvalidHexStringError(lowHex);
+      throw new InvalidHexStringError(`The hex string '${lowHex}' is invalid`);
     }
   }
 
-  constructor(private readonly arr: Uint8Array, private offset: number = 0) {}
+  constructor(private readonly arr: Uint8Array, private offset: number = 0) { }
```
dsawali, line 16: I think the previous was correct (no whitespace in between curly brackets)

packages/taquito-signer/src/derivation-tools/ed25519.ts
```diff
@@ -4,7 +4,8 @@ import { SHA512 } from '@stablelib/sha512';
 import { generateKeyPairFromSeed } from '@stablelib/ed25519';
 import { ExtendedPrivateKey, Hard } from './index';
 import { parseHex } from './utils';
-import { InvalidDerivationPathError, InvalidSeedLengthError } from '../errors';
+import { InvalidSeedLengthError } from '../errors';
+import { InvalidDerivationPathError } from '../errors';
```
dsawali, line 8: not sure why they're on different lines, they should be from the same file it seems

packages/taquito-signer/src/derivation-tools/utils.ts
```diff
@@ -1,4 +1,5 @@
-import { InvalidHexStringError } from "@taquito/utils";
+import { InvalidHexStringError } from '../errors';
```
dsawali, line 1: I see this happening in a few files where a common error is being imported from `@taquito/core` and is re-exported into the file that needs it. I think for those cases we should instead import directly from `@taquito/core`.

The only case where we'd need to import and re-export is when the error used to live in the package, but is then moved into `@taquito/core`. We should be leveraging direct imports from `@taquito/core` wherever possible

packages/taquito/src/batch/rpc-batch-provider.ts
```diff
@@ -326,7 +326,7 @@ export class OperationBatch extends OperationEmitter {
 }
 
 export class RPCBatchProvider {
-  constructor(private context: Context, private estimator: EstimationProvider) {}
+  constructor(private context: Context, private estimator: EstimationProvider) { }
```
dsawali, line 329: same comment here about the whitespace

packages/taquito/src/wallet/wallet.ts
```diff
@@ -37,7 +37,7 @@ export type WalletParamsWithKind =
 export class WalletOperationBatch {
   private operations: WalletParamsWithKind[] = [];
 
-  constructor(private walletProvider: WalletProvider, private context: Context) {}
+  constructor(private walletProvider: WalletProvider, private context: Context) { }
```
dsawali, line 40: same comment here on the whitespace

packages/taquito/src/wallet/wallet.ts
```diff
@@ -156,7 +156,7 @@ export class WalletOperationBatch {
 }
 
 export class Wallet {
-  constructor(private context: Context) {}
+  constructor(private context: Context) { }
```
dsawali, line 159: whitespace

packages/taquito/test/estimate/rpc-estimate-provider.spec.ts
```diff
@@ -26,7 +26,9 @@ import {
 } from '../contract/helper';
 import { OpKind } from '@taquito/rpc';
 import { TransferTicketParams } from '../../src/operations/types';
-import { InvalidAddressError, InvalidContractAddressError } from '@taquito/utils';
+import { InvalidContractAddressError } from '@taquito/utils';
```
dsawali, line 29: I think this can be moved/extended from `@taquito/core` as it's basically a parameter error or a `InvalidXError` but perhaps that's for the other ticket

dsawali (changes requested):

packages/taquito-contracts-library/test/taquito-contracts-library.spec.ts
```diff
@@ -106,7 +107,7 @@ describe('ContractsLibrary tests', () => {
       })
     ).toThrow(
       expect.objectContaining({
-        message: expect.stringContaining('Address is invalid: KTinvalid'),
+        message: expect.stringContaining("The address 'KTinvalid' is invalid"),
```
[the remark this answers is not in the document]
zainen: 👍

packages/taquito-ledger-signer/src/error.ts
```diff
@@ -1,3 +1,5 @@
+export { InvalidDerivationPathError } from '@taquito/core';
```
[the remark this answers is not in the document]
zainen: i have done these to keep them exportable from the same file to make it exportable from the index of this package or the file that contains it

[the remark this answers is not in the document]
zainen: to stop breaking changes of moving it to taquito core

packages/taquito-contracts-library/test/taquito-contracts-library.spec.ts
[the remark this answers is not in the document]
zainen: split the InvalidAddress with InvalidContractAddress but didnt change this one. changed now

packages/taquito-ledger-signer/src/taquito-ledger-signer.ts
zainen: was already moved an imported just missed clearing the declaration

zainen: nvm see what i missed

packages/taquito-local-forging/src/taquito-local-forging.ts
zainen: thanks for the catch put imported from taquito/core then switched it to error

packages/taquito-local-forging/src/michelson/codec.ts
zainen: this is from formatting for the aligning of different ternary operators
should it be reverted? i think it looks a little more clear for the different ternary cases

packages/taquito-local-forging/src/uint8array-consumer.ts
zainen: yea i need to change my formatter since it is default for that 

only spaces we want between curly braces is with variables inbetween right? for the braces that are opened and closed on the same line

packages/taquito-signer/src/derivation-tools/utils.ts
zainen: good point i'll make sure to change them import directly from taquito/core
just wanted to be consistent and picked the import path that was longer 😓

packages/taquito/test/estimate/rpc-estimate-provider.spec.ts
zainen: i have moved it to taquito core!

zainen: will push soon

packages/taquito-local-forging/src/taquito-local-forging.ts
```diff
@@ -47,7 +47,7 @@ export function getCodec(codec: CODEC, _proto: ProtocolsHash) {
 }
 
 export class LocalForger implements Forger {
-  constructor(public readonly protocolHash = PROTOCOL_CURRENT) {}
+  constructor(public readonly protocolHash = PROTOCOL_CURRENT) { }
```
dsawali, line 50: should remove the whitespace between curly brackets

packages/taquito-ledger-signer/src/error.ts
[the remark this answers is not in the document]
dsawali: wouldn't it make more sense to keep InvalidDerivationPathError in this package instead of moving it up to`core`? Or was there a reason why it might need to be put up there?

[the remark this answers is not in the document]
dsawali: if it's not like a shared across packages thing, like say, `InvalidAddressError` or something like that. Might want to keep it contained in the package it's referenced at instead. But I might be missing context

packages/taquito-core/src/error/common.ts
```diff
@@ -0,0 +1,45 @@
+/**
+ *  @category Error
+ *  @description Error that indicates an invalid address being passed or used (both contract and implicit)
```
hui-an-yang, line 3: Thinking with InvalidContractAddressError being distinguished that InvalidAddressError will just be implicit?

packages/taquito-core/src/error/common.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+export class InvalidAddressError extends Error {
```
hui-an-yang, line 5: Unsure I understand the context correctly, I thought we plan to extend common errors by higher level errors created by task1?

packages/taquito-local-forging/package.json
```diff
@@ -73,6 +73,7 @@
   },
   "devDependencies": {
     "@taquito/rpc": "^15.0.1",
+    "@taquito/core": "^15.0.0",
```
hui-an-yang, line 76: Saw your comments in a channel the other day agree that we should update taquito/core to ^15.0.1 to align with other packages.

packages/taquito/test/estimate/rpc-estimate-provider.spec.ts
```diff
@@ -26,7 +26,9 @@ import {
 } from '../contract/helper';
 import { OpKind } from '@taquito/rpc';
 import { TransferTicketParams } from '../../src/operations/types';
-import { InvalidAddressError, InvalidContractAddressError } from '@taquito/utils';
+import { InvalidContractAddressError } from '@taquito/core';
```
hui-an-yang, line 29: if line 29 and 30 are both importing from @taquito/core, will we consider to make it one line?

packages/taquito-utils/src/taquito-utils.ts
```diff
@@ -131,7 +132,7 @@ export function b58decodeL2Address(payload: string) {
  */
 export function encodePubKey(value: string) {
   if (value.substring(0, 2) === '00') {
-    const pref: { [key: string]: Uint8Array } = {
+    const pref: { [key: string]: Uint8Array; } = {
```
hui-an-yang, line 135: Am curious are these semi-colons in curly brackets from linter? Would like to know why it only pops up now?

hui-an-yang (commented):
Would like to know what is the rule of thumb to put error classes in common or not? Feeling like many classes in `packages/taquito-utils/src/errors.ts` are used in other packages can be moved to `packages/taquito-core/src/error/common.ts` to remove circular dependency too.  I understand to avoid breaking changes the best we can, so would like to know under what conditions we put it in `common.ts` file?

packages/taquito-core/src/error/common.ts
zainen: this is just moving the current error to common. we can change what it extends after this ticket

packages/taquito-utils/src/taquito-utils.ts
zainen: from my formatter 
types want to be ended with `;`  even for values of keys in objects

packages/taquito-ledger-signer/src/error.ts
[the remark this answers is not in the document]
zainen: its used in both taquito signer and taquito ledger signer

packages/taquito-contracts-library/src/errors.ts
hui-an-yang, line 1: Would like to understand is there a reason behind exporting the entire @taquito/core package here instead of specific classes used in this package like the rest of the code change?

zainen: this is to just stop breaking changes from happening for moving errors from thats that they would otherwise have been coming from

zainen: in case anyone is importing from the error file directly

packages/taquito-core/src/error/common.ts
hui-an-yang: I see, so taquito/core will also be one of the package that we'll work on ticket 1994. Since common.ts will be used across packages do we want to prioritize as first 5 that we'll tackle?

packages/taquito-local-forging/src/error.ts
```diff
@@ -1,15 +1,6 @@
...
- *  @category Error
- *  @description Error that indicates an invalid block hash being passed or used
- */
-export class InvalidBlockHashError extends Error {
-  public name = 'InvalidBlockHashError';
-  constructor(public message: string) {
-    super(message);
-  }
-}
+export { InvalidAddressError, InvalidHexStringError } from '@taquito/core';
```
hui-an-yang, line 3: Didn't see InvalidAddressError export here previously, would like to understand why we'd like to export here now?

packages/taquito-utils/src/errors.ts
```diff
@@ -1,3 +1,5 @@
+export { InvalidHexStringError, InvalidContractAddressError } from '@taquito/core';
```
hui-an-yang, line 1: Do we also want to export InvalidAddressError, InvalidBlockHashError here to avoid breaking changes?

packages/taquito-local-forging/src/error.ts
zainen: thanks for the catch

packages/taquito-core/src/error/common.ts
zainen: yes because any modifications of locations or additions of higher level functions will add to taquito/core

dsawali (approved):

dsawali:
> Would like to know what is the rule of thumb to put error classes in common or not? Feeling like many classes in `packages/taquito-utils/src/errors.ts` are used in other packages can be moved to `packages/taquito-core/src/error/common.ts` to remove circular dependency too. I understand to avoid breaking changes the best we can, so would like to know under what conditions we put it in `common.ts` file?

I agree, we should probably centralize them in core now

hui-an-yang (approved):
except for `export * from '@taquito/core';` in `packages/taquito-contracts-library/src/errors.ts` to be error class specific the rest lgtm :)

PR #2240, opened 2022-12-20
Mondaynet 2022 12 19

zainen:
adjust constants for mondaynet protocol constants changes 

change tx-rollup tests to be kathmandu and limanet running

remove transferTicket integration test as it relies on mondaynet block period and toru is now disabled on mondaynet


zainen:
changes were merged in #2246

PR #2246, opened 2022-12-21
Remove kathmandunet integration tests

zainen:
removes kathmandu integration tests and configs 
does not touch svelt configs


danielelisi (approved):
Changes to Github workflows LGTM :+1:

packages/taquito-michel-codec/src/micheline-emitter.ts
```diff
@@ -12,7 +12,7 @@ export interface FormatOptions {
 }
 
 class Formatter {
-  constructor(private opt?: FormatOptions, private lev: number = 0) {}
```
dsawali, line 15: was this done by the linter? I think the previous one is already correct

zainen: i think its my settings where the `{}` want a `{ }` same as any `{ key: 'string'}` instead of `{key: 'string'}`

dsawali (approved):

PR #2263, opened 2023-01-03
fix bytes in signing data example

### LINKED ISSUE #2185: Uncompilable code in Signer example on tezostaquito.io
**Description**

The code sample at https://tezostaquito.io/docs/signing/#generating-a-signature-with-beacon-sdk has a couple of problems with it:
```
import { char2Bytes } from '@taquito/utils';

const bytes = char2Bytes(formattedInput);
const payloadBytes = '05' + '0100' + char2Bytes(bytes.length) + bytes;
```
* Incorrect padding (according to https://tezos-dev.slack.com/archives/GQFF7PHCY/p1669672178242869)
* char2Bytes will not compile with a numeric argument

zainen:
update bytes length to have padding and divide hex string length to bytes length

closes #2185

### DIFF
--- a/docs/signing.md
+++ b/docs/signing.md
@@ -55,7 +55,7 @@ import { char2Bytes } from '@taquito/utils';
 
     const bytes = char2Bytes(formattedInput);
     const bytesLength = (bytes.length / 2).toString(16)
-    const addPadding = `0000${bytesLength}`
+    const addPadding = `00000000${bytesLength}`
     const paddedBytesLength = addPadding.slice(addPadding.length - 4)
     const payloadBytes = '0x' + '05' + '0100' + char2Bytes(paddedBytesLength) + bytes;
 
--- a/docs/signing.md
+++ b/docs/signing.md
@@ -56,7 +56,7 @@ import { char2Bytes } from '@taquito/utils';
     const bytes = char2Bytes(formattedInput);
     const bytesLength = (bytes.length / 2).toString(16)
     const addPadding = `00000000${bytesLength}`
-    const paddedBytesLength = addPadding.slice(addPadding.length - 4)
+    const paddedBytesLength = addPadding.slice(addPadding.length - 8)
     const payloadBytes = '0x' + '05' + '0100' + char2Bytes(paddedBytesLength) + bytes;
 
 ```
--- a/docs/signing.md
+++ b/docs/signing.md
@@ -63,7 +63,7 @@ The bytes representation of the string must be prefixed with 5 pieces of informa
 - "0x" add zero-padding
 - "05" indicates that this is a Micheline expression
 - "01" indicates that a string was converted to bytes
-- the number of characters in the bytes (hexadecimal string divided by 2) string encoded on 4 bytes as string 
+- the number of characters in the bytes (hexadecimal string divided by 2) encoded on 4 bytes
 - bytes of formatted input to be signed
 
 Once you have your bytes, you can send them to the wallet to have them signed:
--- a/docs/signing.md
+++ b/docs/signing.md
@@ -53,8 +53,11 @@ After formatting the string properly, you can convert it into bytes, for example
 ```js
 import { char2Bytes } from '@taquito/utils';
 
-const bytes = char2Bytes(formattedInput);
-const payloadBytes = '0x' + '05' + '0100' + char2Bytes((bytes.length / 2).toString()) + bytes;
+    const bytes = char2Bytes(formattedInput);
+    const bytesLength = (bytes.length / 2).toString(16)
+    const addPadding = `0000${bytesLength}`
+    const paddedBytesLength = addPadding.slice(addPadding.length - 4)
+    const payloadBytes = '0x' + '05' + '0100' + char2Bytes(paddedBytesLength) + bytes;
 
 ```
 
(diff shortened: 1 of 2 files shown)


docs/signing.md
```diff
@@ -54,14 +54,17 @@ After formatting the string properly, you can convert it into bytes, for example
...
 const bytes = char2Bytes(formattedInput);
-const payloadBytes = '05' + '0100' + char2Bytes(bytes.length) + bytes;
+const payloadBytes = '0x' + '05' + '01' + `00${char2Bytes((bytes.length / 2).toString())}` + bytes;
+
 ```
 
-The bytes representation of the string must be prefixed with 3 pieces of information:
+The bytes representation of the string must be prefixed with 5 pieces of information:
 
+- "0x" add zero-padding
```
dsawali, line 63: I think `0x` is optional in our case

docs/signing.md
```diff
@@ -54,14 +54,17 @@ After formatting the string properly, you can convert it into bytes, for example
...
 - "05" indicates that this is a Micheline expression
 - "01" indicates that a string was converted to bytes
-- the number of characters in the byte string encoded on 4 bytes
+- the number of characters in the byte string encoded on 4 bytes as string
+- bytes of input
```
dsawali, line 67: could you elaborate what this line means?

docs/signing.md
```diff
@@ -54,14 +54,17 @@ After formatting the string properly, you can convert it into bytes, for example
 import { char2Bytes } from '@taquito/utils';
 
 const bytes = char2Bytes(formattedInput);
-const payloadBytes = '05' + '0100' + char2Bytes(bytes.length) + bytes;
+const payloadBytes = '0x' + '05' + '01' + `00${char2Bytes((bytes.length / 2).toString())}` + bytes;
```
dsawali, line 57: could you also make sure that the examples compile properly? (post a screenshot or something)

dsawali (changes requested):

docs/signing.md
zainen: i added a test that runs the exact same code here 
it doesnt check the string is what it should be but it checks that the hex dec is valid

docs/signing.md
zainen: Hi @egarson just wanted to check with you if this is Klassare meant. if not i'll ping them for a review!

egarson (commented):
Could you improve in the assertion? Existence is a bit weak as verification of correct functionality.

docs/signing.md
```diff
@@ -54,14 +54,18 @@ After formatting the string properly, you can convert it into bytes, for example
...
 const bytes = char2Bytes(formattedInput);
-const payloadBytes = '05' + '0100' + char2Bytes(bytes.length) + bytes;
+const bytesLength = (bytes.length / 2).toString(16);
+const addPadding = `00000000${bytesLength}`;
+const paddedBytesLength = addPadding.slice(addPadding.length - 8);
+const payloadBytes = '05' + '01' + paddedBytesLength + bytes;
 ```
 
-The bytes representation of the string must be prefixed with 3 pieces of information:
+The bytes representation of the string must be prefixed with 4 pieces of information:
```
dsawali, line 63: refer to slack thread regarding change

docs/signing.md
```diff
@@ -54,14 +54,18 @@ After formatting the string properly, you can convert it into bytes, for example
...
 
 - "05" indicates that this is a Micheline expression
 - "01" indicates that a string was converted to bytes
```
dsawali, line 66: refer to slack thread

packages/taquito-utils/test/format.spec.ts
```diff
@@ -21,4 +22,23 @@ describe('Format', () => {
...
+    ].join(' ');
+
+    const bytes = char2Bytes(formattedInput);
+    const bytesLength = (bytes.length / 2).toString(16);
+    const addPadding = `00000000${bytesLength}`;
+    const paddedBytesLength = addPadding.slice(addPadding.length - 8);
+    const payloadBytes = '05' + '01' + paddedBytesLength + bytes;
+    const regex = new RegExp('^(0x|0X)?[a-fA-F0-9]+$');
+    const check = regex.test(payloadBytes);
+    expect(check).toEqual(true);
```
dsawali, line 41: I'm not sure this actually tests the functionality properly. The test seems to just go through the steps in the util method. A better assertion would be to re-convert the bytes to string successfully with the same value retained.

dsawali (changes requested):

packages/taquito-utils/test/format.spec.ts
```diff
@@ -21,4 +22,24 @@ describe('Format', () => {
   it('Should convert tz to mutez', () => {
     expect(format('tz', 'mutez', 1)).toEqual(new BigNumber(1000000));
   });
+
+  it('Should be valid bytes tezostaquito example', () => {
```
dsawali, line 26: could you update the description to be more specific. This tests a specific part in the doc on signing bytes through a wallet. We should make that reflect the test name. It also lives in a completely separate part of the codebase.

You could give a link in the form of a comment to the section in the doc. 

It's also prob a good idea to make a new describe block since this doesn't seem like it's the same set of cases.

dsawali (approved):

PR #2277, opened 2023-01-09
2020 prepare provider

### LINKED ISSUE #2020: Add a PrepareProvider on the TezosToolkit
Follow up of #432 

The Taquito contract API abstracts multiple steps when sending an operation to the blockchain. Preparing an operation in the expected RPC format is not evident. (Need to include the right counter, some endpoints will expect the branch, some will expect the protocol hash, parameters in the case of a contract entry point call need to be in Michelson, ...). This problem brings the following ones: 
    1. Hard to retrieve information of one of the particular underlying steps (i.e., simulation). The user needs to use the lower-level abstractions and figure out how to write the operation object. 
    2. Hard to obtain the operation hash before its injection

Existing Solution to run a simulation of a transfer operation:
1. fetch the source `tezos.signer.publicKeyHash()`
2. fetch the counter from the RPC and increase its value `tezos.rpc.getContract(pkh)`
3. fetch the branch from the RPC `(await tezos.rpc.getBlockHeader()).hash`
4. run an estimation of the operation `Tezos.estimate.transfer()` (optional if setting my own fees)
5. properly construct the operation object:
```
{
  branch,
  contents: [{
    kind: OpKind.TRANSACTION,
		source,
    destination: 'tz1...',
    amount: '1000000', //mutez
    fee: String(estimate.suggestedFeeMutez),
    gas_limit: String(estimate.gasLimit),
    storage_limit: String(estimate.storageLimit),
    source,
    counter: String(parseInt(counter || '0', 10) + 1),
  }],
}
```
6. forge the operation `Tezos.rpc.forge({...})`
7. sign the forged bytes `Tezos.signer.sign(forgedBytes.opbytes, new Uint8Array([3]))`
8. fetch the protocol from the RPC `(await this.rpc.getBlockMetadata()).next_protocol`
9. call the preapplyOperations RpcClient method:
```
await tezos.rpc.preapplyOperations([{
  branch,
  contents: [{
    kind: OpKind.TRANSACTION,
		source,
    destination: 'tz1...',
    amount: '1000000', //mutez
    fee: String(estimate.suggestedFeeMutez),
    gas_limit: String(estimate.gasLimit),
    storage_limit: String(estimate.storageLimit),
    source,
    counter: String(parseInt(counter || '0', 10) + 1),
  }],
signature,
protocol
}]);
```

- As a user, when sending an operation (i.e., transaction, origination, delegation,...), I want to retrieve information from one of the particular underlying steps without getting to the operation injection. For example, I want to retrieve the output of the simulation step easily. I want the preparation step to be abstracted, so I don't need to set the different properties myself (counter, branch, parameters with the correct format,...)
- Similar to the preceding point, as a user, I want to easily retrieve the output of the simulation step when calling a smart contract method. I want the preparation step to be abstracted, so I don't need to set the different properties myself (counter, branch, parameters with the correct format,...)
- As a user, I want to retrieve the operation's hash before injection easily. I want the preparation step to be abstracted, so I don't need to set the different properties myself (counter, branch, parameters with the correct format,...)

We propose to provide a new class named PrepareProvider which contains the following methods: transfer, originate, reveal, registerDelegate, setDelegate, contractCall, batch. Each method returns an instance of a PreparedOperation. The PrepareProvider class will be instantiated on the TezosToolkit and accessible via a member (named prepare). We will also provide utility functions that allow the conversion of the preparedOperation into the different RPC expected formats. Examples of helper functions will be: toDryRun, toForge. 

Example of use - Prepare and simulate (dry-run) a transfer operation:
```
const preparedTransfer = await TezosToolkit.prepare.transfer(...);
TezosToolkit.rpc.preapplyOperations(toDryRun(preparedTransfer));
```

Example of use - Prepare and forge a transfer operation:
```
const preparedTransfer = await TezosToolkit.prepare.transfer(...);
TezosToolkit.forger.forge(toForge(preparedTransfer));
```

Example of use - Prepare and simulate (dry-run) a contract method call:
```
const contractAbs = await TezosToolkit.contract.at('address');
const op = contractAbs.methods.increment(1)
const preparedContractCall = await TezosToolkit.prepare.contractCall(op);
TezosToolkit.rpc.preapplyOperations(toDryRun(preparedContractCall))
```

- [ ] Implement a `PrepareProvider` class and a specific `preparedOperation` class for each operation type as well as for a contract call

Follow-up issues: #2256, #2257

**Additional context**
https://www.notion.so/ecad/Design-refactor-contract-API-8cbcd324414d475bbe2d667e86fc188c

dsawali:
closes #2020 

- Added a PrepareProvider class to modularize Tezos operation preparation step for our users.
- Added methods for operations like `transaction`, `delegation`, `origination`, etc as member methods accessible from the `TezosToolkit` class.
- Added `batch` member method to prepare for batched operations
- Added `contractCall` member method to prepare contract calls on `ContractAbstraction`
- Added unit tests
- Added integration tests
- Updated implementation from last PR to now utilize the estimate provider

Background information:
- https://www.notion.so/ecad/Design-refactor-contract-API-8cbcd324414d475bbe2d667e86fc188c

## Release Note Draft Snippet

### DIFF
--- a/integration-tests/prepare-operation.spec.ts
+++ b/integration-tests/prepare-operation.spec.ts
@@ -1,0 +1,42 @@
+import { 
+  OpKind, 
+  RPCOriginationOperation, 
+  DEFAULT_FEE, 
+  DEFAULT_GAS_LIMIT, 
+  DEFAULT_STORAGE_LIMIT, 
+  RPCBallotOperation,
+  PreparedOperation
+} from '@taquito/taquito';
+import { CONFIGS } from './config';
+
+
+
+CONFIGS().forEach(({ lib, setup }) => {
+  const Tezos = lib;
+
+  describe(`Test Preparation of operations using the PrepareProvider`, () => {
+    beforeEach(async (done) => {
+      await setup();
+      done();
+    });
+
+    it('should be able to prepare a ballot operation', async (done) => {
+      const prepared = await Tezos.prepare.ballot({ operation: 
+        {
+          kind: OpKind.BALLOT,
+          proposal: 'PtKathmankSpLLDALzWw7CGD2j2MtyveTwboEYokqUCP4a1LxMg',
+          ballot: 'yay', 
+        } as RPCBallotOperation 
+      });
+
+      expect(prepared).toBeDefined();
+      expect(prepared.counter).toBeDefined();
+      expect(prepared.opOb).toBeDefined();
+      expect(prepared.opOb.branch).toBeDefined();
+      expect(prepared.opOb.contents).toBeDefined();
+      expect(prepared.opOb.protocol).toBeDefined();
+
+      done();
+    });
+  });
+}) 
--- a/integration-tests/prepare-operation.spec.ts
+++ b/integration-tests/prepare-operation.spec.ts
@@ -1,6 +1,6 @@
+import { OperationContentsBallot, OperationContentsTransaction } from '@taquito/rpc';
 import { 
   OpKind, 
-  RPCBallotOperation,
 } from '@taquito/taquito';
 import { CONFIGS } from './config';
 
@@ -37,13 +37,10 @@ CONFIGS().forEach(({ lib, setup }) => {
       done();
     });
 
-    it('should be able to prepare a ballot operation', async (done) => {
-      const prepared = await Tezos.prepare.ballot({ operation: 
-        {
-          kind: OpKind.BALLOT,
-          proposal: 'PtKathmankSpLLDALzWw7CGD2j2MtyveTwboEYokqUCP4a1LxMg',
-          ballot: 'yay', 
-        } as RPCBallotOperation 
+    it('should be able to prepare a transaction operation', async (done) => {
+      const prepared = await Tezos.prepare.transaction({
+        to: 'tz1QZ6KY7d3BuZDT1d19dUxoQrtFPN2QJ3hn',
+        amount: 5
       });
 
       expect(prepared).toBeDefined();
@@ -51,8 +48,75 @@ CONFIGS().forEach(({ lib, setup }) => {
       expect(prepared.opOb).toBeDefined();
       expect(prepared.opOb.branch).toBeDefined();
       expect(prepared.opOb.contents).toBeDefined();
-      expect(prepared.opOb.protocol).toBeDefined();
 
+      const content = prepared.opOb.contents[0] as OperationContentsTransaction;
+
+      expect(content.kind).toEqual('transaction');
+      expect(content.destination).toEqual('tz1QZ6KY7d3BuZDT1d19dUxoQrtFPN2QJ3hn')
+      expect(content.amount).toEqual('5000000');
+
+      done();
+    });
+
+    it('should be able to prepared a batch operation', async (done) => {
+      const prepared = await Tezos.prepare.batch([
+        {
+          kind: OpKind.TRANSACTION,
+          to: 'tz1QZ6KY7d3BuZDT1d19dUxoQrtFPN2QJ3hn',
+          amount: 2,
+        },
+        {
+          kind: OpKind.TRANSACTION,
+          to: 'tz1QZ6KY7d3BuZDT1d19dUxoQrtFPN2QJ3hn',
+          amount: 2,
+        },
+      ]);
+
+      expect(prepared).toBeDefined();
+      expect(prepared.counter).toBeDefined();
+      expect(prepared.opOb).toBeDefined();
+      expect(prepared.opOb.branch).toBeDefined();
+      expect(prepared.opOb.contents).toBeDefined();
+      expect(prepared.opOb.contents.length).toEqual(2);
+      expect(prepared.opOb.contents[0].kind).toEqual('transaction');
+      expect(prepared.opOb.contents[1].kind).toEqual('transaction');
+
+      done();
+    });
+
+    it('should be able to prepare a txRollupOriginate operation', async (done) => {
+      const prepared = await Tezos.prepare.txRollupOrigination({});
+
+      expect(prepared).toBeDefined();
+      expect(prepared.counter).toBeDefined();
+      expect(prepared.opOb).toBeDefined();
+      expect(prepared.opOb.branch).toBeDefined();
+      expect(prepared.opOb.contents).toBeDefined();
+      expect(prepared.opOb.contents[0].kind).toEqual('tx_rollup_origination');
+
+      done();
+    });
+
+    it('should be able to prepare a ballot operation', async (done) => {
+      const prepared = await Tezos.prepare.ballot({
+        proposal: 'PtKathmankSpLLDALzWw7CGD2j2MtyveTwboEYokqUCP4a1LxMg',
+        ballot: 'yay'
+      });
+
+      expect(prepared).toBeDefined();
+      expect(prepared.counter).toBeDefined();
+      expect(prepared.opOb).toBeDefined();
+      expect(prepared.opOb.branch).toBeDefined();
+      expect(prepared.opOb.contents).toBeDefined();
+
+      const content = prepared.opOb.contents[0] as OperationContentsBallot
+
+      expect(prepared.opOb.contents[0].kind).toEqual('ballot');
+      expect(content.proposal).toBeDefined();
+      expect(content.proposal).toEqual('PtKathmankSpLLDALzWw7CGD2j2MtyveTwboEYokqUCP4a1LxMg');
+      expect(content.ballot).toBeDefined();
+      expect(content.ballot).toEqual('yay');
+      expect(prepared.opOb.protocol).toEqual('PtLimaPtLMwfNinJi9rCfDPWea8dFgTZ1MeJ9f1m2SRic6ayiwW');
       done();
     });
 
@@ -68,7 +132,7 @@ CONFIGS().forEach(({ lib, setup }) => {
       expect(prepared.opOb.protocol).toBeDefined();
 
       done();
-    })
+    });
   });
 
 }) 
--- a/integration-tests/prepare-operation.spec.ts
+++ b/integration-tests/prepare-operation.spec.ts
@@ -1,22 +1,39 @@
 import { 
   OpKind, 
-  RPCOriginationOperation, 
-  DEFAULT_FEE, 
-  DEFAULT_GAS_LIMIT, 
-  DEFAULT_STORAGE_LIMIT, 
   RPCBallotOperation,
-  PreparedOperation
 } from '@taquito/taquito';
 import { CONFIGS } from './config';
 
-
-
 CONFIGS().forEach(({ lib, setup }) => {
   const Tezos = lib;
+  let contractAddress: string;
 
   describe(`Test Preparation of operations using the PrepareProvider`, () => {
-    beforeEach(async (done) => {
+    beforeAll(async (done) => {
       await setup();
+      
+      try {
+        const op = await Tezos.contract.originate({
+          code: `{ parameter (or (or (int %decrement) (int %increment)) (unit %reset)) ;
+            storage int ;
+            code { UNPAIR ;
+                   IF_LEFT { IF_LEFT { SWAP ; SUB } { ADD } } { DROP 2 ; PUSH int 0 } ;
+                   NIL operation ;
+                   PAIR } }`,
+          storage: 1
+        });
+        await op.confirmation();
+
+        contractAddress = op.contractAddress!;
+
+      } catch(e: any) {
+        console.log('Unable to originate contract: ', JSON.stringify(e));
+      }
+
+      done();
+    })
+    
+    beforeEach(async (done) => {
       done();
     });
 
@@ -38,5 +55,20 @@ CONFIGS().forEach(({ lib, setup }) => {
 
       done();
     });
+
+    it('should be able to prepare a contractCall', async (done) => {
+      const contractAbs = await Tezos.contract.at(contractAddress);
+      const method = await contractAbs.methods.increment(1);
+      const prepared = await Tezos.prepare.contractCall(method);
+
+      expect(prepared).toBeDefined();
+      expect(prepared.counter).toBeDefined();
+      expect(prepared.opOb.branch).toBeDefined();
+      expect(prepared.opOb.contents).toBeDefined();
+      expect(prepared.opOb.protocol).toBeDefined();
+
+      done();
+    })
   });
+
 }) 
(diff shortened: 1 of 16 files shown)


integration-tests/prepare-operation.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+      expect(prepared.opOb.branch).toBeDefined();
+      expect(prepared.opOb.contents).toBeDefined();
+      expect(prepared.opOb.contents.length).toEqual(2);
+      expect(prepared.opOb.contents[0].kind).toEqual('transaction');
+      expect(prepared.opOb.contents[1].kind).toEqual('transaction');
+
+      done();
+    });
+
+    it('should be able to prepare a txRollupOriginate operation', async (done) => {
```
zainen, line 87: is this still possible? thought txRollups were disabled

packages/taquito/src/prepare/interface.ts
```diff
@@ -0,0 +1,198 @@
...
+  /**
+   *
+   * @description Method to prepare a delegation operation
+   *
+   * @param params delegation operation parameters
+   * @param source string or undefined source pkh
+   *
+   * @returns a PreparedOperation object
+   */
+  // delegation({ operation, source }: PrepareOperationParams): Promise<PreparedOperation>;
```
zainen, line 68: this needed?

packages/taquito/src/prepare/interface.ts
```diff
@@ -0,0 +1,198 @@
...
+/**
+ * @description PrepareProvider is a utility class to output the prepared format of an operation
+ */
+export interface PreparationProvider {
+  /**
+   *
+   * @description Method to prepare a reveal operation
+   *
+   * @param params reveal operation parameters
+   * @param source string or undefined source pkh
```
zainen, line 31: all of these params are a little confusing
is it possible to link the interface thats applicable?

packages/taquito/src/prepare/interface.ts
```diff
@@ -0,0 +1,198 @@
...
+   * @param params txRollupSubmitBatch operation parameters
+   * @param source string or undefined source pkh
+   *
+   * @returns a PreparedOperation object
+   */
+  txRollupSubmitBatch(params: TxRollupBatchParams): Promise<PreparedOperation>;
+
+  /**
+   *
+   * @description Method to prepare a update_consensus_key operation
```
zainen, line 107: an

packages/taquito/src/prepare/interface.ts
```diff
@@ -0,0 +1,198 @@
...
+   * @param params reveal operation parameters
+   * @param source string or undefined source pkh
+   *
+   * @returns a PreparedOperation object
+   */
+  reveal(params: RevealParams): Promise<PreparedOperation>;
+
+  /**
+   *
+   * @description Method to prepare a origination operation
```
zainen, line 39: an origination

packages/taquito/src/prepare/interface.ts
```diff
@@ -0,0 +1,198 @@
...
+   * @param params TransferTicketx operation parameters
+   * @param source string or undefined source pkh
+   *
+   * @returns a PreparedOperation object
+   */
+  transferTicket(params: TransferTicketParams): Promise<PreparedOperation>;
+
+  /**
+   *
+   * @description Method to prepare a increase_paid_storage operation
```
zainen, line 129: an

packages/taquito/src/prepare/prepare-provider.ts
```diff
@@ -0,0 +1,863 @@
...
+
+  private getFee(op: RPCOpWithFee, pkh: string, headCounter: number) {
+    if (!this.#counters[pkh] || this.#counters[pkh] < headCounter) {
+      this.#counters[pkh] = headCounter;
+    }
+    const opCounter = ++this.#counters[pkh];
+
+    return {
+      counter: `${opCounter}`,
+      fee: typeof op.fee === 'undefined' ? '0' : `${op.fee}`,
```
zainen, line 106: could any of these be something other than undefined? the type is `fee?: string` correct?
if so we could replace all of the `typeof op.fee <ternary op>` to eg: `fee: op.fee ?? '0'` and seems a little more readable to me 
what do you think? 
its a typescript nuance

packages/taquito/src/prepare/prepare-provider.ts
```diff
@@ -0,0 +1,863 @@
...
+    } else {
+      return [op];
+    }
+  }
+
+  private constructOpContents(
+    ops: RPCOperation[],
+    headCounter: number,
+    pkh: string,
+    source?: string | undefined,
```
zainen, line 150: is this `undefined` needed? `source?:` already covers that

packages/taquito/src/prepare/prepare-provider.ts
```diff
@@ -0,0 +1,863 @@
...
+      switch (op.kind) {
+        case OpKind.ACTIVATION:
+        case OpKind.DRAIN_DELEGATE:
+          return {
+            ...op,
+          };
+        case OpKind.ORIGINATION:
+          return {
+            ...op,
+            balance: typeof op.balance !== 'undefined' ? `${op.balance}` : '0',
```
zainen, line 163: should we keep consistency with order for ternaries? here its `!==` other was `===` (if keeping the ternaries)

packages/taquito/src/prepare/prepare-provider.ts
```diff
@@ -0,0 +1,863 @@
...
+        counter: headCounter,
+      };
+    } else {
+      throw new RevealOperationError(`The publicKeyHash '${pkh}' has already been revealed.`);
+    }
+  }
+
+  /**
+   *
+   * @description Method to prepare a origination operation
```
zainen, line 276: an

packages/taquito/src/prepare/prepare-provider.ts
```diff
@@ -0,0 +1,863 @@
...
+        contents,
+        protocol,
+      },
+      counter: headCounter,
+    };
+  }
+
+  /**
+   *
+   * @description Method to prepare a update_consensus_key operation
```
zainen, line 531: an

packages/taquito/src/prepare/prepare-provider.ts
```diff
@@ -0,0 +1,863 @@
...
+        contents,
+        protocol,
+      },
+      counter: headCounter,
+    };
+  }
+
+  /**
+   *
+   * @description Method to prepare a increase_paid_storage operation
```
zainen, line 574: an

packages/taquito/test/contract/rpc-contract-provider.spec.ts
```diff
@@ -2167,6 +2166,7 @@ describe('RpcContractProvider test', () => {
         const opBatch = new OperationBatch(rpcContractProvider['context'], mockEstimate);
 
         expect(rpcContractProvider.batch()).toBeInstanceOf(OperationBatch);
+        console.log(opBatch);
```
zainen, line 2169: log

packages/taquito/test/operations/origination-operation.spec.ts
```diff
@@ -118,6 +118,8 @@ describe('Origination operation', () => {
         fakeContext,
         fakeContractProvider
       );
+
+      console.log(op['params']);
```
zainen, line 122: log

zainen (commented):
Awesome work! 
just some comments 
will look into jest.spyOn to understand whats happening in the test better!

packages/taquito/src/prepare/interface.ts
dsawali: Not sure what's confusing about the params, maybe you could elaborate?

integration-tests/prepare-operation.spec.ts
dsawali: I think some of the operations might still be available, I put them in just in case

integration-tests/prepare-operation.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+      const prepared = await Tezos.prepare.transaction({
+        to: 'tz1QZ6KY7d3BuZDT1d19dUxoQrtFPN2QJ3hn',
+        amount: 5
+      });
+
+      expect(prepared).toBeDefined();
+      expect(prepared.counter).toBeDefined();
+      expect(prepared.opOb).toBeDefined();
+      expect(prepared.opOb.branch).toBeDefined();
+      expect(prepared.opOb.contents).toBeDefined();
```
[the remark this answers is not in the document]
dsawali: unfortunately not, the values of each property will change depending on the counter, block hash, and fee estimates. I have more assertions below that to make sure the `PreparedOperation` object is correct.

packages/taquito/src/prepare/interface.ts
zainen: should param be put here twice? or should it just be referencing what each param is or the interface that it is

dsawali: yes, that's an annotation. you use 1 `@param` annotation per parameter generally

dsawali: it's a pretty common practice, you can see existing examples in `rpc-contract-provider.ts`

integration-tests/prepare-operation.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+      expect(content.proposal).toEqual('PtKathmankSpLLDALzWw7CGD2j2MtyveTwboEYokqUCP4a1LxMg');
+      expect(content.ballot).toBeDefined();
+      expect(content.ballot).toEqual('yay');
+      expect(prepared.opOb.protocol).toEqual('PtLimaPtLMwfNinJi9rCfDPWea8dFgTZ1MeJ9f1m2SRic6ayiwW');
+      done();
+    });
+
+    it('should be able to prepare a contractCall', async (done) => {
+      const contractAbs = await Tezos.contract.at(contractAddress);
+      const method = await contractAbs.methods.increment(1);
```
egarson, line 125: I would grab the current value (setup) and then assert after the call that the storage was indeed incremented by 1, being pedagogical.

packages/taquito/src/prepare/interface.ts
egarson: +1

packages/taquito/src/prepare/interface.ts
egarson: +1

packages/taquito/src/taquito.ts
```diff
@@ -348,6 +350,13 @@ export class TezosToolkit {
     return this._context.contract;
   }
 
+  /**
+   * @description Provide access to tezos operation preparation utilities
+   */
+  get prepare(): PreparationProvider {
+    return this._context.prepare;
```
egarson, line 357: I would use lambda syntax for all 1-line methods but that's just a stylistic preference. In general I like to see as much code as possible, with spaces only between method declarations, most of the time (and: assumes small, focused methods),

egarson (approved):
Looks like excellent work in general, super clean and simple.
Like I said above, a lot of extraneous line breaks, but again, just a stylistic preference.

integration-tests/prepare-operation.spec.ts
dsawali: Ah, about that. This provider actually _doesn't_ inject the operation at all. Rather to prepare the object itself to be forged and injected.

dsawali: that assertion won't be possible yet. But follow up tickets will then utilize this provider class to be used in the contract-provider and others. I believe that's where the test you're proposing will live at.

integration-tests/prepare-operation.spec.ts
hui-an-yang: Do we want to change it to limanet instead in case we change over to the next protocol will break. Also, will be easier to remove them when change over to mumbai?

integration-tests/prepare-operation.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+      expect(prepared.opOb.branch).toBeDefined();
+      expect(prepared.opOb.contents).toBeDefined();
+
+      const content = prepared.opOb.contents[0] as OperationContentsBallot
+
+      expect(prepared.opOb.contents[0].kind).toEqual('ballot');
+      expect(content.proposal).toBeDefined();
+      expect(content.proposal).toEqual('PtKathmankSpLLDALzWw7CGD2j2MtyveTwboEYokqUCP4a1LxMg');
+      expect(content.ballot).toBeDefined();
+      expect(content.ballot).toEqual('yay');
```
hui-an-yang, line 118: Something small, if we are expecting proposal and ballot to be 'PtKathmankSpLLDALzWw7CGD2j2MtyveTwboEYokqUCP4a1LxMg', and 'yay' maybe checking if it's `.toBeDefined()` is not needed.

dsawali: good point, I'll remove

packages/taquito/test/prepare/prepare-provider.spec.ts
```diff
@@ -0,0 +1,1000 @@
...
+          protocol: 'test_protocol',
+        },
+        counter: 0,
+      });
+    });
+  });
+
+  describe('drainDelegate', () => {
+    it('should return a prepared drain_delegate operation', async () => {
+      jest.spyOn(context.estimate, 'reveal').mockResolvedValue(estimate);
```
hui-an-yang, line 251: I am unsure if my understanding is correct, would like to double-check. Is estimation step only necessary for operations with `gas_limit`, `storage_limit`, `fee` properties in contents object? This is the reason when I worked with drain_delegate, I didn't update it in `EstimationProvider`. If my assumption is correct, maybe we don't need line 251, but I can be wrong and didn't implement drain_delegate correctly too.

hui-an-yang: or line 251 supposed to be `.mockResolveValue(undefined)`?

packages/taquito/test/prepare/prepare-provider.spec.ts
```diff
@@ -0,0 +1,1000 @@
...
+          ],
+          protocol: 'test_protocol',
+        },
+        counter: 0,
+      });
+    });
+  });
+  describe('updateConsensusKey', () => {
+    it('should return a prepared udpateConsensusKey operation', async () => {
+      jest.spyOn(context.estimate, 'reveal').mockResolvedValue(estimate);
```
hui-an-yang, line 430: Not sure update_consensus key op need to reveal either, since it's the delegate initiating this op it has been revealed.

packages/taquito/test/prepare/prepare-provider.spec.ts
dsawali: Yeah you are correct, it was a miss on my part. Thanks for catching that

packages/taquito/test/prepare/prepare-provider.spec.ts
dsawali: update_consensus_key should have estimates. So I think this is still relevant. But you are correct with drain_delegate

PR #2291, opened 2023-01-13
1994 improve errors http utils

hui-an-yang:

## Release Note Draft Snippet


packages/taquito-core/src/interface/http_status_code.ts
```diff
@@ -0,0 +1,378 @@
+/**
```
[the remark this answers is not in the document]
hui-an-yang: yes no changed values, but the file name from status_code to http_status_code to be specific since it's been moved out of http-utils package

packages/taquito-core/src/error/common.ts
```diff
@@ -2,8 +2,9 @@ import {
   ParameterValidationError,
   PermissionDeniedError,
   UnsupportedAction,
-  ValidationError,
+  InternalValidationError,
 } from './errors';
+import { STATUS_CODE } from '../interface/http_status_code';
```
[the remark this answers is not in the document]
hui-an-yang: We haven't talked and gave definition how error code will be but my understanding now will include internal and external like 400 and 500 for https.

packages/taquito-core/src/error/common.ts
```diff
@@ -183,3 +184,22 @@ export class ProhibitedActionError extends PermissionDeniedError {
     super(message);
```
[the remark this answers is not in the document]
hui-an-yang: good catch will delete line184

packages/taquito-core/src/error/common.ts
```diff
@@ -183,3 +184,22 @@ export class ProhibitedActionError extends PermissionDeniedError {
...
+ */
+export class HttpResponseError extends Error {
+  public name = 'HttpResponse';
+
+  constructor(
+    public message: string,
+    public status: STATUS_CODE,
+    public statusText: string,
+    public body: string,
+    public url: string
```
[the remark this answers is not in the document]
hui-an-yang: will do thank you!!

packages/taquito-core/src/taquito-core.ts
```diff
@@ -5,3 +5,4 @@
 
 export * from './error/errors';
 export * from './error/common';
+export * from './interface/http_status_code';
```
zainen, line 8: curious why its called interface?
I was thinking maybe `constants` since its part of something that should never change

packages/taquito-http-utils/src/taquito-http-utils.ts
```diff
@@ -15,6 +16,7 @@ const adapterPromise = isNode
 
 export * from './status_code';
 export { VERSION } from './version';
+export { HttpResponseError } from '@taquito/core';
```
zainen, line 19: two core exports?

zainen (commented):
i think my last comments that stand out to me right now

packages/taquito-core/src/error/errors.ts
```diff
@@ -1,20 +1,10 @@
...
-  public name = 'ExternalError';
-  constructor(public message: string) {
-    super(message);
-  }
-}
-
-/**
- *  @category Error
- *  @description Error that indicates an invalid user data input
+ *  @description Error that indicates the data get from other servies has incorrect format
```
dsawali, line 3: typo *services

Also I think instead of 'incorrect format' it could be more generalized as just 'invalid' since these errors might happen not only because of format errors.
i.e. 'Error that indicates invalid data being fetched from other services'

packages/taquito-core/src/error/errors.ts
```diff
@@ -1,20 +1,10 @@
...
  */
-export class ValidationError extends Error {
-  public name = 'ValidationError';
+export class InternalValidationError extends Error {
+  public name = 'InternalValidationError';
+  public category = this.name;
```
dsawali, line 7: can you explain why we're assigning category the same as its name? Seems redundant

packages/taquito-core/src/interface/http_status_code.ts
```diff
@@ -0,0 +1,378 @@
...
+ * Hypertext Transfer Protocol (HTTP) response status codes.
+ * @see {@link https://en.wikipedia.org/wiki/List_of_HTTP_status_codes}
+ */
```
dsawali, line 4: wouldn't this make sense to keep at http utils instead?

packages/taquito-core/src/error/common.ts
```diff
@@ -183,3 +184,22 @@ export class ProhibitedActionError extends PermissionDeniedError {
     super(message);
   }
 }
+
+// Unsure the need to extend from high category ReponseError where there will only this error as child
+/**
+ *  @category Error
+ *  @description This error will be thrown when the endpoint returns an HTTP error to the client
+ */
+export class HttpResponseError extends Error {
```
zainen, line 198: could this extend a parent?

packages/taquito-http-utils/src/taquito-http-utils.ts
```diff
@@ -33,29 +35,11 @@ export interface HttpRequestOptions {
...
-    super(message);
-  }
-}
-
 /**
  *  @category Error
  *  @description Error that indicates a general failure in making the HTTP request
  */
-export class HttpRequestFailed extends Error {
+export class HttpRequestFailed extends ParameterValidationError {
```
dsawali, line 42: I don't think this is a ParameterValidationError, it should fall under a category like 'HttpError'

packages/taquito/test/contract/big-map.spec.ts
```diff
@@ -7,169 +7,176 @@ import { BigMapAbstraction } from '../../src/contract/big-map';
  * BigMapAbstraction test
  */
 describe('BigMapAbstraction test', () => {
-    let rpcContractProvider: {
-        getBigMapKeysByID: jest.Mock<any, any>;
-        getStorage: jest.Mock<any, any>;
-        getBigMapKey: jest.Mock<any, any>;
-        getBigMapKeyByID: jest.Mock<any, any>;
+  let rpcContractProvider: {
+    getBigMapKeysByID: jest.Mock<any, any>;
```
dsawali, line 11: are these just lint changes?

packages/taquito-core/src/interface/http_status_code.ts
dsawali: i don't think this should live in core, the only times we will check for http codes will be in side the http util package

dsawali (changes requested):

packages/taquito-core/src/error/errors.ts
zainen: we thought of doing this so that it is visible in the error structure that the parent is InternalValidationError
otherwise since all the parent does is use Super it would just act like `SomeError extends Error`
rather than `SomeError extends ParentError extends Error`

dsawali: couldn't you just grab super(name) from the child to and output it in the error message?

zainen: the idea behind this was that we wouldnt need to change all of the error messages of the children that we would be extending to make visibility of the parent error

packages/taquito-http-utils/src/taquito-http-utils.ts
hui-an-yang: Will create HttpError category and extends from it

packages/taquito-core/src/interface/http_status_code.ts
hui-an-yang: http_status_code is used multiple packages and it's used in HttpResponseError under @taquito/core.  Tried to make @taquito/core a standalone not importing from http utils is the reason that I moved it but understand if it's not ideal too.

packages/taquito-core/src/error/errors.ts
hui-an-yang: tried super(name) and got name as Error not the parent but from grandparent class 🙈

packages/taquito/test/contract/big-map.spec.ts
hui-an-yang: yes I only manually changed line1 from `'@taquito/http-utils';` to `'@taquito/http-core';`

packages/taquito-core/src/error/common.ts
hui-an-yang: Will create httpError and extends from it

packages/taquito-core/src/interface/http_status_code.ts
zainen: i agree it makes more sense in http-utils but see currently how this move is necessary to remove its dependency in multiple packages 
if our util packages do not depend on any of our other base level packages then it should stay in http-utils but until then it would make sense to keep it out of http-utils until those dependencies are sorted out

packages/taquito-utils/src/errors.ts
```diff
@@ -58,7 +58,7 @@ export class InvalidProtocolHashError extends ValidationError {
  *  @category Error
  *  @description General error that indicates a failure when trying to convert data from one type to another
  */
-export class ValueConversionError extends ValidationError {
+export class ValueConversionError extends InternalValidationError {
```
zainen, line 61: should this be ParameterValidationError? 
its used only in the hex2buf helper function 
it may be used internally however the function alone should be by parameter

hui-an-yang: good catch, will update now

packages/taquito-core/src/taquito-core.ts
hui-an-yang: it's updated as constants

zainen (approved):
nice work! looks good to me
double check with @dsawali

packages/taquito-core/src/error/errors.ts
```diff
@@ -1,20 +1,21 @@
...
 /**
  *  @category Error
- *  @description Error that indicates an invalid user data input
+ *  @description Error that indicates invalid data being fetched from other services
  */
-export class ValidationError extends Error {
-  public name = 'ValidationError';
+export class InternalValidationError extends TaquitoError {
+  public name = 'InternalValidationError';
+  public category = this.name;
```
dsawali, line 18: should remove these as we discussed

packages/taquito-core/src/error/errors.ts
```diff
@@ -35,8 +37,9 @@ export class ParameterValidationError extends Error {
  *  @category Error
  *  @description Error returned by RPC
  */
-export class RpcError extends Error {
+export class RpcError extends TaquitoError {
   public name = 'RpcError';
+  public category = this.name;
```
dsawali, line 42: same here, and a few other instances.

packages/taquito-core/src/error/errors.ts
```diff
@@ -68,19 +72,33 @@ export class UnsupportedAction extends Error {
...
   constructor(public message: string) {
     super(message);
   }
 }
 
 /**
  *  @category Error
- *  @description Error that indicates
+ *  @description Error that permission deneied for a certain action
```
dsawali, line 85: typo *denied

packages/taquito-core/test/errors.spec.ts
```diff
@@ -1,12 +1,196 @@
...
+      throw new InvalidOperationKindError('test');
+    } catch (e) {
+      expect(e).toBeInstanceOf(TaquitoError);
+      expect(e).toBeInstanceOf(ParameterValidationError);
+      expect(e).toBeInstanceOf(InvalidOperationKindError);
+      expect(e.message).toContain('test');
+    }
+  });
+
+  it('should throw an DeprecationError', () => {
```
dsawali, line 160: typo *a

dsawali (changes requested):

packages/taquito-utils/src/errors.ts
```diff
@@ -58,7 +58,7 @@ export class InvalidProtocolHashError extends ValidationError {
  *  @category Error
  *  @description General error that indicates a failure when trying to convert data from one type to another
  */
-export class ValueConversionError extends ValidationError {
+export class ValueConversionError extends ParameterValidationError {
```
dsawali, line 61: I don't think this should be a ParameterValidationError. Where is this usually thrown?

dsawali: I think this is usually thrown in utils package. So something like TaquitoUtilsError makes more sense here

dsawali (changes requested):

packages/taquito-michel-codec/src/base58.ts
```diff
@@ -91,8 +91,8 @@ export class Base58DecodingError extends Error {
  *  @category Error
  *  @description
  */
-export class InvalidHashLengthError extends Error {
-  public name = 'InvalidHashLengthError';
+export class InvalidMessageLength extends Error {
+  public name = 'InvalidMessageLength';
   constructor(public message: string) {
     super(message);
   }
```
zainen, line 98: just realised should this be moved to an error file? as well as this may be a core error since there are multiple places where we may reject parts of hashes are are invalid

zainen: as well as are we going to stop our convention of having `....Error` ? or should it be `InvalidMessageLengtherror`

PR #2342, opened 2023-02-13
update contract address mondaynet

zainen:
contracts failed but passed after updating
could you test your local to double check if its just my local and i missed something?


hui-an-yang (approved):
lgtm

PR #2365, opened 2023-02-28
update ledger example to ghostnet

michaelkernaghan:
after a review of the Taquito examples the ledger example needs to be updated to work with Ghostnet.
the rest are already done and working.


zainen (approved):

hui-an-yang (approved):
lgtm:)

PR #2400, opened 2023-03-20
Update to Beacon 4.0.0

ac10n:
Upgraded to the latest version of the Beacon SDK.
We because of build errors in the website, we had to:
- Remove the website from the npm workspaces
- Load dependencies asynchronously

## Release Note Draft Snippet

Upgraded to the latest beta version of the Beacon SDK.

### DIFF
--- a/apps/taquito-test-dapp/src/App.svelte
+++ b/apps/taquito-test-dapp/src/App.svelte
@@ -8,6 +8,8 @@
   import Layout from "./Layout.svelte";
   import TestContainer from "./lib/TestContainer.svelte";
 
+  let layout: Layout;
+
   // https://ide.ligolang.org/p/RL93C86hftTTCNGU0ykLMw | https://ide.ligolang.org/p/61ENTorA4PSfQg-jhpu0jA
   // https://better-call.dev/florencenet/KT1PzUGbdKaN332Smfd1ExpdKQ7BSzzJRqJ4/operations
   // https://better-call.dev/granadanet/KT1T836HqhBu9waqmknStVDCXu2WogZtzsNz/operations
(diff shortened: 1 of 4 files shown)


website/src/theme/CodeBlock/index.js
```diff
@@ -88,54 +57,138 @@ export default ({
...
+        validatePublicKey,
+        validateSignature,
+        validateBlock,
+        validateProtocol,
+        validateOperation,
+        b58cencode,
+        prefix,
+        Prefix
+      } = await import('@taquito/utils');
+      const { BeaconWallet } = await import('@taquito/beacon-wallet');
```
dsawali, line 78: for dynamic imports, wouldn't we only need to do this for the `beacon-wallet` package? does it need to dynamically import all packages?

danielelisi:
Can you note in the PR why removing `website` from NPM workspaces is needed?

danielelisi (approved):
Changes to NPM workspaces LGTM, the website is built and deployed correctly.

dsawali:
when this gets approved, can we merge it into 16.0.1 branch instead of master?

website/src/theme/CodeBlock/index.js
AlirezaHaghshenas: I was thinking that any of the packages can later change a dependency and break the website. So maybe loading all of them is not a terrible idea after all.

AlirezaHaghshenas: @dsawali 
There is something we can discuss. Here, all `async` library loads are done one after the other. An alternative would be to run all the `import` functions, get the `Promise` objects and `awaitAll` them.

This makes the code a bit harder to follow, but would possibly get a performance benefit. In production code it could be useful, but for the website, I can't be sure.

What do you suggest?

dsawali: If we can leverage that for performance, that probably would be nice. But I'm not sure if that would reflect in our website. Docusaurus handles most of the page rendering. I am not opposed to the change, however.

michaelkernaghan:
I tried the netlify preview of the Taquito test dapp, but it's not connecting to any wallets.

PR #2424, opened 2023-04-03
add doc for local forger

michaelkernaghan:
Adds a document describing how to use the local forger.

### DIFF
--- a/docs/local_forger.md
+++ b/docs/local_forger.md
@@ -6,7 +6,7 @@ author: Michael Kernaghan
 
 # How to Use the Taquito Local Forging Package
 
-If you are looking for a way to interact with the Tezos blockchain, Taquito is a great option. One of the tools offered by Taquito is the local forging package, which allows you to simulate the forging and signing of transactions locally without interacting with a node. Here is a step-by-step guide on how to use the Taquito local forging package:
+One of the tools offered by Taquito is the local forging package, which allows you to forge transactions locally without interacting with a node. Here is a step-by-step guide on how to use the Taquito local forging package:
 
 ## Step 1: Import the Local Forger
 
@@ -19,7 +19,7 @@ import { LocalForger } from '@taquito/local-forging';
 
 ## Step 2: Create a Transaction
 
-Next, you must create a transaction you want to simulate. You can do this using the Taquito library as you normally would. Here is an example of creating a transaction to transfer 1 XTZ from one address to another:
+Next, you must create a transaction you want to forge. You can do this using the Taquito library as you normally would. Here is an example of creating a transaction to transfer 1 XTZ from one address to another:
 
 ```
 import { TezosToolkit } from '@taquito/taquito';
@@ -35,7 +35,7 @@ const transferOperation = await tezos.contract.transfer({ to, amount }).send({ f
 
 ## Step 3: Forge the Transaction
 
-Once you have created your transaction, you can use the LocalForger class to simulate the forging and signing process. Here is an example of how to do this:
+Once you have created your transaction, you can use the LocalForger class. Here is an example of how to do this:
 
 ```
 const forger = new LocalForger();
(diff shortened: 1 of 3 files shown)


docs/local_forger.md
```diff
@@ -0,0 +1,64 @@
+---
+title: Local Forging Package
+id: local_forger
+author: Michael Kernaghan
+---
+
+# How to Use the Taquito Local Forging Package
+
+If you are looking for a way to interact with the Tezos blockchain, Taquito is a great option. One of the tools offered by Taquito is the local forging package, which allows you to simulate the forging and signing of transactions locally without interacting with a node. Here is a step-by-step guide on how to use the Taquito local forging package:
```
AlirezaHaghshenas, line 9: > If you are looking for a way to interact with the Tezos blockchain, Taquito is a great option.
I feel this is a bit out of place in here. Anyone who reads the docs for local forging is expected to know what Taquito is.

> which allows you to simulate the forging and signing of transactions
I think "simulate" and "and signing" is better be removed here, and an informal definition of forging can be helpful.

docs/local_forger.md
```diff
@@ -0,0 +1,64 @@
...
+You will need to import the LocalForger class from Taquito to use the local forging package. You can do this by adding the following line to your code:
+
+```
+import { LocalForger } from '@taquito/local-forging';
+
+```
+
+## Step 2: Create a Transaction
+
+Next, you must create a transaction you want to simulate. You can do this using the Taquito library as you normally would. Here is an example of creating a transaction to transfer 1 XTZ from one address to another:
```
AlirezaHaghshenas, line 22: Also, "simulate" here can be replaced with "forge".

docs/local_forger.md
```diff
@@ -0,0 +1,64 @@
...
+const to = 'tz1def...uvw';
+const amount = 1;
+
+const transferOperation = await tezos.contract.transfer({ to, amount }).send({ from });
+
+```
+
+## Step 3: Forge the Transaction
+
+Once you have created your transaction, you can use the LocalForger class to simulate the forging and signing process. Here is an example of how to do this:
```
AlirezaHaghshenas, line 38: Also here

docs/local_forger.md
```diff
@@ -0,0 +1,64 @@
...
+## Step 5: Broadcast the Transaction
+
+Finally, you can broadcast the signed transaction to the blockchain. Here is an example of how to do this using the Taquito library:
+
+```
+const operation = await tezos.rpc.sendOperation({ operation: signatureBytes });
+
+```
+
+That's it! You have now successfully used the Taquito local forging package to simulate the forging and signing of a transaction locally. This can be useful for testing and development, as it allows you to experiment with the transaction process without interacting with the blockchain.
```
AlirezaHaghshenas, line 64: Is there a reason why local forging should not be used for production?

docs/local_forger.md
michaelkernaghan: i removed the first sentence, but the extra words from the second one

docs/local_forger.md
michaelkernaghan: that would be interacting with the blockchain, but the point of local forger is to be able to simulate without using a real blockchain.

docs/local_forger.md
```diff
@@ -0,0 +1,64 @@
...
+const forgedBytes = await forger.forge(transferOperation);
+
+```
+
+## Step 4: Sign the Transaction
+
+After the transaction has been forged, you can sign it using your private key. Here is an example of how to sign the transaction:
+
+```
+const { bytes: signatureBytes } = await tezos.signer.sign(forgedBytes, from);
```
hui-an-yang, line 51: Not about local forging but if using signer here, might want to show how to configure singer before using signer?
`Tezos.setSignerProvider(new InMemorySigner.fromSecretKey('edsk...'))`

hui-an-yang (approved):
minor comment, but overall lgtm 🙌

PR #2442, opened 2023-04-19
2344 support contract call nested option of case some none

### LINKED ISSUE #2344: Unable to pass "Some None" to contract call
With the current implementation for contract call parameters, it is impossible to pass the value `Some None` to an entrypoint that requires a value of type `option (option timestamp)`, for example.
Although this is an edge case, Taquito should provide a more intuitive way to pass options and nested options as contract call parameters, as the current way becomes inefficient or even broken with nested options.

hui-an-yang:

closes #2344 

- [x] You have added or updated corresponding documentation

## Release Note Draft Snippet

In order to support case Some(None) when making contract calls in nested options like type (option (option nat)), we are introducing new semantic  { Some: null } to `EncodeObject()` in @taquito/michelson-encoder. 
So user can make contract call with `.methodsObjects` in contractAPI and walletAPI. 
The old semantic still works but will look to deprecate them in the future. 
Michelson type: option nat as Michelson Value: Some(1) can be called with 1 or [1] can now also be called with { Some: 1 } with methodObject.
And the corresponding `Execute()` in  @taquito/michelson-encoder will return `{ Some: '1' }` instead of previous `1`


claudebarde (approved):

dsawali (approved):

hui-an-yang:
Have tried deprecated option token .encodeObject() with old semantic realizing that Schema class only has one encode function and it's actually using encodeObject under the hood. If deprecate old semantic will cause regression of new Schema().encode() too. (ref https://github.com/ecadlabs/taquito/blob/master/packages/taquito-michelson-encoder/src/schema/storage.ts#L236)

PR #2445, opened 2023-04-19
1344 expose injector in tezos tool kit

hui-an-yang:

## Release Note Draft Snippet


packages/taquito/src/taquito.ts
```diff
@@ -336,6 +343,23 @@ export class TezosToolkit {
...
+  setInjectorProvider(injectorProvider?: SetProviderOptions['injectorProvider']) {
+    if (!this._options.injectorProvider && typeof injectorProvider === 'undefined') {
+      const p = new RpcInjector(this._context);
+      this._context.injector = p;
+      this._options.injectorProvider = p;
+    } else if (typeof injectorProvider !== 'undefined') {
+      this._context.injector = injectorProvider;
+      this._options.injectorProvider = injectorProvider;
+    }
+  }
```
AlirezaHaghshenas, line 361: What happens if `this._options.injectorProvider` is truthy? Neither of the `if` statements will be executed, and the code does nothing without giving feedback to the calling code. So the user thinks they have set the injectorProvider, but they have not.

packages/taquito/src/taquito.ts
```diff
@@ -336,6 +343,23 @@ export class TezosToolkit {
...
+   *
+   */
+  setInjectorProvider(injectorProvider?: SetProviderOptions['injectorProvider']) {
+    if (!this._options.injectorProvider && typeof injectorProvider === 'undefined') {
+      const p = new RpcInjector(this._context);
+      this._context.injector = p;
+      this._options.injectorProvider = p;
+    } else if (typeof injectorProvider !== 'undefined') {
+      this._context.injector = injectorProvider;
+      this._options.injectorProvider = injectorProvider;
```
AlirezaHaghshenas, line 359: I like how setting it to `undefined` has the effect of setting it to the default value, but if it is already set, then setting it to `undefined` will just do nothing. Also, it does not notify the developer that the code has no effect. This can lead to behaviour that is unexpected for the developer. I think if we want to keep this behaviour, it's good to throw an error in the case that we are ignoring the change.

hui-an-yang: Thinking this is a ux design that it's dumb proof that there is already preset value and user accidentally reset to undefined it will use the preset default that TezosToolkit is always functioning. 
When a TezosToolkit can't do any operations without injector when it is initialized it's set to default automatically
What do you guys think @AlirezaHaghshenas @dsawali

AlirezaHaghshenas: Agree. I still think that a `setSomething` function should in all states either actually set that field, or throw and error. Otherwise there could be circumstances where the user expects something be have been changed, while it still keeps its old value.

dsawali: It's an interesting question for sure. I think I am more in favour of setting a default value. Since an injector isn't something that's optional for users to use, it makes sense to have a default behaviour when it's not set. I think making it idiot proof is more beneficial in this case. I don't have the numbers, but I assume most users won't think about using an external injector.

packages/taquito/src/taquito.ts
```diff
@@ -336,6 +343,23 @@ export class TezosToolkit {
...
 
+  /**
+   * @description Sets injector provider on the Tezos Taquito instance
+   *
+   * @param options Injector to use to interact with the Tezos network by default RpcInjector
+   *
+   */
+  setInjectorProvider(injectorProvider?: SetProviderOptions['injectorProvider']) {
+    if (!this._options.injectorProvider && typeof injectorProvider === 'undefined') {
+      const p = new RpcInjector(this._context);
```
dsawali, line 354: small nitpick, but looking at the other set_Provider code, wouldn't the variable `p` be `i` instead since 'injector' starts with an `i`

hui-an-yang: good point, will update

PR #2483, opened 2023-05-16
Replace usage of operation-emitter in favour of the prepare-provider

### LINKED ISSUE #2431: Replace the usage of OperationEmitter class in RpcEstimateProvider
The OperationEmitter class is being extended by a number of Providers and does many things outside of it's intended scope. Refactor the usage of Operation Emitter in `RpcEstimateProvider` and leverage PrepareProvider where necessary.

Possible approaches:
- Replace OperationEmitter class with a Provider parent class for these child providers to give basic access to Context
- Leverage the usage of PrepareProvider in relevant parts of the child providers

dsawali:
closes #2431 

This PR will replace the usage of `OperationEmitter` in favour of the newly created `PrepareProvider`
- replaced `OperationEmitter` with `PrepareProvider` in `RpcContractProvider`
- replaced `OperationEmitter` with `PrepareProvider` in `RpcEstimateProvider`
- replaced `OperationEmitter` with `PrepareProvider` in `RpcBatchProvider`
- replaced `OperationEmitter` with `PrepareProvider` in `RpcTzProvider`
- Updated `PrepareProvider` to not couple with the `EstimateProvider`
- Updated PrepareProvider to use default estimates (fee, gaslimit, storagelimit) when they are not passed in the parameters
- Created a new abstract class `Provider` for all other providers to extend from

## Release Note Draft Snippet


integration-tests/injecting-multiple-manager-operations-in-a-block.spec.ts
```diff
@@ -4,26 +4,22 @@ CONFIGS().forEach(({ lib, rpc, setup }) => {
...
-        const op2Promise = Tezos.contract.transfer({ to: 'tz1ZfrERcALBwmAqwonRXYVQBDT9BjNjBHJu', amount: 2 });
-
-        const [op1, op2] = await Promise.all([op1Promise, op2Promise])
+        const op1 = await Tezos.contract.transfer({ to: 'tz1ZfrERcALBwmAqwonRXYVQBDT9BjNjBHJu', amount: 1 });
+        const op2 = await Tezos.contract.transfer({ to: 'tz1ZfrERcALBwmAqwonRXYVQBDT9BjNjBHJu', amount: 2 });
 
         await op1.confirmation();
         await op2.confirmation();
-
       } catch (error: any) {
```
AlirezaHaghshenas, line 19: This code is from earlier, but it does not verify that the code throws. It tests that **IF** it throws, the error is right. But the test can pass if the code just succeeds. I think we need to replace it with:

```
it(‘blahblah’, () => {
  expect(async () => {
    //my code
  }).toThrow(‘error message’);})
```

dsawali: I agree, I'm actually looking to fix this assertion. The current implementation potentially allow this type of practice to fail with different errors. Depending on the timing, it could fail with `key already revealed`, a counter issue, or a mempool conflicting issue. Which sounds unstable, but is actually expected. It hammers the point even more that we'll need to confirm an op before doing another.

integration-tests/prepare-operation.spec.ts
```diff
@@ -124,16 +124,21 @@ CONFIGS().forEach(({ lib, setup, protocol, createAddress }) => {
       const receiver = await createAddress();
 
       const pkh = await receiver.signer.publicKeyHash();
-      const preparedTransfer = await Tezos.prepare.transaction({ amount: 1, to: pkh });
+      const estimates = await Tezos.estimate.transfer({ to: pkh, amount: 1 });
+      const preparedTransfer = await Tezos.prepare.transaction({
```
AlirezaHaghshenas, line 128: Because the test does not bother with values of estimates, do we need to make this change? Does all code targeting Taquito need to make the same change?

dsawali: due to the nature of this test, when we prepare & want to send at the same time, we will need to estimate.

Since the test is to test the `preapply` method, we'll need to estimate or the preparation will fail due to incorrect fees.

In a normal circumstance, users would just do something like `Tezos.contract.transfer()`

packages/taquito/src/prepare/prepare-provider.ts
```diff
@@ -54,41 +52,48 @@ import {
...
+  storageLimit?: number;
+  gasLimit?: number;
+}
+
+const mergeLimits = (
+  userDefinedLimit: Limits,
+  defaultLimits: Required<Limits>
+): Required<Limits> => {
+  return {
+    fee: typeof userDefinedLimit.fee === 'undefined' ? defaultLimits.fee : userDefinedLimit.fee,
```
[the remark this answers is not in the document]
dsawali: yeah I could make the change, this was copied over from the old code in operation-emitter

packages/taquito/src/prepare/prepare-provider.ts
```diff
@@ -101,6 +106,49 @@ export class PrepareProvider implements PreparationProvider {
     return this.context.readProvider.getCounter(pkh, 'head') ?? '0';
   }
 
+  private adjustGasForBatchOperation(
+    gasLimitBlock: BigNumber,
+    gaslimitOp: BigNumber,
+    numberOfOps: number
+  ) {
+    return BigNumber.min(gaslimitOp, gasLimitBlock.div(numberOfOps + 1));
```
AlirezaHaghshenas, line 114: I'm still worried that this kind of division can cause some problems. But this existed in previous code, so no action needed at this time.

packages/taquito/src/provider.ts
```diff
@@ -0,0 +1,202 @@
...
+    if (calculatedStorage && calculatedStorage % 1 !== 0) {
+      throw new InvalidEstimateValueError(
+        `Storage Limit value must not be a decimal: ${calculatedStorage}`
+      );
+    }
+
+    if (fee === undefined || gasLimit === undefined || storageLimit === undefined) {
+      const estimation = await estimator({ fee, gasLimit, storageLimit, ...(rest as any) });
+
+      if (calculatedFee === undefined) {
```
AlirezaHaghshenas, line 74: again, `??` could be useful here

packages/taquito/src/provider.ts
```diff
@@ -0,0 +1,202 @@
...
+
+    return {
+      fee: calculatedFee,
+      gasLimit: calculatedGas,
+      storageLimit: calculatedStorage,
+    };
+  }
+
+  async getRPCOp(param: ParamsWithKind) {
+    switch (param.kind) {
```
AlirezaHaghshenas, line 95: This switch statement does not cover `TX_ROLLUP_ORIGINATION`, `TX_ROLLUP_SUBMIT_BATCH`, and `UPDATE_CONSENSUS_KEY`.

AlirezaHaghshenas (approved):
I still lack enough context about the flow to give an approval about the changes in the flow. But the code changes seem fine to me. I left some comments, by the way.

packages/taquito/src/provider.ts
dsawali: tx operations are deprecated, whereas I don't think UPDATE_CONSENSUS_KEY is batchable in this case

integration-tests/injecting-multiple-manager-operations-in-a-block.spec.ts
dsawali: so it seems that toThrow probably only looks at the first function that's being executed, it doesn't seem like it captures the throw when it's the subsequent methods. I will keep the assertion block as is for now, but we'll investigate sometime in the future.

packages/taquito-rpc/src/types.ts
```diff
@@ -603,6 +603,18 @@ export type OperationContents =
   | OperationContentsSmartRollupRecoverBond
   | OperationContentsSmartRollupTimeout;
 
+export type OperationContentsWithFee =
```
hui-an-yang, line 606: Didn't see this used is this only defined and exported for end user?

dsawali: you're right, it should be removed instead. good catch

packages/taquito/src/batch/rpc-batch-provider.ts
```diff
@@ -208,17 +191,6 @@ export class OperationBatch extends OperationEmitter {
...
-   * @description Add a tx rollup batch operation to the batch
-   *
-   * @param params Tx rollup batch operation parameter
-   */
-  withTxRollupSubmitBatch(params: TxRollupBatchParams) {
-    this.operations.push({ kind: OpKind.TX_ROLLUP_SUBMIT_BATCH, ...params });
-    return this;
-  }
-
   async getRPCOp(param: ParamsWithKind) {
```
hui-an-yang, line 194: reveal and updateConcensusKey are both manager operation which [can be batch](https://tezos.gitlab.io/nairobi/blocks_ops.html?highlight=update_consensus_key#manager-operation-batches) and activation is not (understand it's old code, just would be nice to correct it)

dsawali: yeah activation is not batchable, that was def copied over. Gonna update that

packages/taquito/src/contract/rpc-contract-provider.ts
```diff
@@ -1,28 +1,38 @@
 import { HttpResponseError, STATUS_CODE } from '@taquito/http-utils';
 import { BigMapKeyType, MichelsonMap, MichelsonMapKey, Schema } from '@taquito/michelson-encoder';
-import { OpKind, ScriptResponse } from '@taquito/rpc';
+import {
+  OpKind,
+  OperationContentsBallot,
```
hui-an-yang, line 5: not sure if it's intentional but love this alphabetic order 🙌

packages/taquito/src/batch/rpc-batch-provider.ts
dsawali: it doesn't seem like updateConsensuskey is supported in batch. Iirc I had a conversation with Roxane to not batch it since it makes a bit less sense for users/bakers to put that into a batch operation. So maybe we can address that some other time?

packages/taquito/src/contract/rpc-contract-provider.ts
dsawali: I credit github copilot for this 😄

packages/taquito/src/provider.ts
hui-an-yang: same missing reveal and updateConsensusKey (understand if reveal is not needed)

hui-an-yang: understand to not support batch updateConsensusKey in taquito from previous thread

packages/taquito/test/contract/rpc-contract-provider.spec.ts
```diff
@@ -2259,78 +1494,6 @@ describe('RpcContractProvider test', () => {
...
-      expect(result.commitments_and_ciphertexts).toEqual([]);
-      expect(result.nullifiers).toEqual([]);
-
-      expect(mockRpcClient.getSaplingDiffById.mock.calls[0][0]).toEqual('133');
-      expect(mockRpcClient.getSaplingDiffById.mock.calls[0][1]).toEqual({ block: '654321' });
-      done();
-    });
-  });
-
-  describe('Storage', () => {
```
hui-an-yang, line 2304: So is storage, why it's taken out now?

hui-an-yang (approved):
lgtm 🎉 great work!!

PR #2535, opened 2023-06-19
updated rxjs version from 6 to 7, ts to 4.2.4, and some refactor

### LINKED ISSUE #2261: Consider upgrading to rxjs version 7
**Description**
RxJS V7 is out for a couple of years now and 8 is coming soon-ish - would you consider upgrading to it? 

It's blocking the upgrade of other libraries.

Thank you.

**Acceptance criteria:**
* Play with an upgrade in a branch
* Explore if we can avoid the TS upgrade. The idea is to not upgrade TS, but upgrade rxjs to the highest version we have. Upgrading with TS at the same time is more risky, so if we can upgrade one and then another, it would be better and easier to downgrade if required.
  * discuss with the team if there is any TS dependencies

dsawali:
closes #2261 

- updated RxJS version from v6.x to v7.8.1
- updated TS version for the `@taquito/taquito` package (and the parent package) from 4.1.5 to 4.2.4 
- refactored some objects to have typechecks
- added a type to encompass some typechecks better
- refactored some RxJS 6 methods that were deprecated
- added ts-ignore to circumvent TS incompatibility with michel-codec (will be removed in the future in a refactor)
- updated some unit tests to comply with additional typechecks

## Release Note Draft Snippet


packages/taquito/package.json
```diff
@@ -31,7 +31,7 @@
     "node": ">=16"
   },
   "scripts": {
-    "test": "jest --coverage",
+    "test": "jest --coverage --testPathIgnorePatterns=operation-factory.spec.ts",
```
hui-an-yang, line 34: would like to understand the reason to ignore the test

packages/taquito/src/operations/operation-errors.ts
```diff
@@ -125,8 +126,13 @@ export const flattenErrors = (
     for (let j = 0; j < results[i].contents.length; j++) {
```
hui-an-yang, line 126: nitpick, kinda feel like refactor the entire flattenErrors function, don't like the nested if and for loops 😅

packages/taquito/src/operations/operations.ts
```diff
@@ -130,7 +135,7 @@ export class Operation {
     return (
       this.results.map((result) => {
         if (hasMetadataWithResult(result)) {
-          return result.metadata.operation_result.status;
+          return (result.metadata.operation_result as OperationResult).status;
```
hui-an-yang, line 138: preference wise, i'd declare type defined variable and return variable.status then casting.

packages/taquito/src/wallet/batch-operation.ts
```diff
@@ -19,28 +21,36 @@ export class BatchWalletOperation extends WalletOperation {
...
   }
 
-  public async revealOperation() {
+  // TODO: update errors being thrown into a custom class
+  public async revealOperation(): Promise<OperationContentsAndResultReveal> {
     const operationResult = await this.operationResults();
-    return operationResult.find((x) => x.kind === OpKind.REVEAL) as
-      | OperationContentsAndResultReveal
-      | undefined;
+    if (!operationResult) {
```
hui-an-yang, line 27: great work to throw error instead letting it be undefined 🙌

packages/taquito-sapling/package.json
```diff
@@ -102,7 +102,7 @@
     "rimraf": "^3.0.2",
     "rollup": "^2.61.1",
     "rollup-plugin-sourcemaps": "^0.6.3",
-    "rollup-plugin-typescript2": "^0.31.1",
+    "rollup-plugin-typescript2": "^0.32.1",
```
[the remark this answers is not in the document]
dsawali: yes, TS 4.2.4 specifically

packages/taquito/package.json
dsawali: this test utilizes a lot of methods from `rx-sandbox` testing package that we can no longer use with RxJS 7. We will need to address this in the near future. But in the interest of time and bandwidth, I chose to skip it for now.

packages/taquito/src/operations/batch-operation.ts
```diff
@@ -58,7 +59,8 @@ export class BatchOperation
         .filter((result) => BATCH_KINDS.indexOf(result.kind) !== -1)
         .map((result) => {
           if (hasMetadataWithResult(result)) {
-            return result.metadata.operation_result.status;
+            const opResult = result.metadata.operation_result as BatchOperationResult;
```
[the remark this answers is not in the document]
dsawali: casting is necessary still since opResult is type unknown. TS will also throw an error if you only type it as such. Same comment as the other one

packages/taquito/src/operations/operations.ts
dsawali: same comment as the one above

packages/taquito/src/operations/operation-errors.ts
dsawali: I agree, I also think that it can probably be written better. But since it's non-blocking and it's logically still correct, I would rather leave it be. What do you think?

packages/taquito/src/wallet/batch-operation.ts
```diff
@@ -19,28 +21,36 @@ export class BatchWalletOperation extends WalletOperation {
...
+      throw new ObservableError('Unable to fetch operation results');
+    } else {
+      return operationResult.find(
+        (x) => x.kind === OpKind.REVEAL
+      ) as OperationContentsAndResultReveal;
```
hui-an-yang, line 32: might want to keep type as below that it's still possible for a wallet batch operations to have no reveal right?

`    | OperationContentsAndResultReveal
      | undefined;
`

packages/taquito/test/wallet/operation.spec.ts
```diff
@@ -1,12 +1,13 @@
+import { TestScheduler } from 'rxjs/testing';
 import { BlockResponse } from '@taquito/rpc';
-import { rxSandbox } from 'rx-sandbox';
 import { Context } from '../../src/context';
 import { WalletOperation } from '../../src/wallet';
 import { blockResponse } from './data';
 
-describe('WalletOperation', () => {
-  const toJSON = (x: any) => JSON.parse(JSON.stringify(x));
+describe('increasePaidStorageOperation', () => {
```
hui-an-yang, line 7: from the file name doesn't seem to be a `increasePaidStorageOperation` specific test suite

packages/taquito/src/operations/batch-operation.ts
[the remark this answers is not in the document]
hui-an-yang: the operation_result type seems lost in `hasMetadataWithResult` function, will we consider defined it better in the function so we can avoid casting

packages/taquito/src/operations/operation-errors.ts
hui-an-yang: I understand, let's leave it for now

hui-an-yang (commented):
Great work to migration the code base to rxjs 7 and ts 4.2.4. 
Besides my impression that type casting compromise type safety was advised try best to avoid it but i'll respect your decision too. 
Just some minor comments to address

packages/taquito/src/operations/batch-operation.ts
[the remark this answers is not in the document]
dsawali: TS unfortunately doesn't derive the subset of types from `hasMetadataWithResult`. It will always return unknown regardless, which is the reason why we needed to cast it

packages/taquito/test/wallet/operation.spec.ts
dsawali: I think that might have been a copy/paste error. Will fix

packages/taquito/src/wallet/batch-operation.ts
dsawali: yeah that makes sense, updated.

packages/taquito/src/operations/batch-operation.ts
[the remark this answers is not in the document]
hui-an-yang: I understand :)

example/example-streamer-custom-retry-logic.ts
hui-an-yang, line 1: thanks for updating example!!

hui-an-yang (approved):
lgtm🎉

PR #2569, opened 2023-07-13
feat: implement the failing_noop operation

AlirezaHaghshenas:

## Release Note Draft Snippet


packages/taquito/src/estimate/rpc-estimate-provider.ts
```diff
@@ -461,4 +462,26 @@ export class RPCEstimateProvider extends Provider implements EstimationProvider
     }
     return Estimate.createEstimateInstanceFromProperties(estimateProperties);
   }
+
```
dsawali, line 465: I'm not sure if we need estimate for failing_noop?

dsawali (changes requested):

dsawali:
I might be mistaken, but from the ticket it seems like we want to support it in the local forger instead and not the contract API. This PR seems like it's been made to 'inject' a failing_noop instead. @AlirezaHaghshenas

apps/taquito-test-dapp/src/tests.ts
```diff
@@ -316,6 +316,30 @@ const signPayloadAndSend = async (
...
+    const signedPayload = await tezos.wallet.signFailingNoop({
+      arbitrary: bytes,
+      basedOnBlock: 'head'
+    });
+
+    return {
+      success: true,
+      opHash: "",
+      output: signedPayload.prefixSig,
+      sigDetails: { input, bytes: signedPayload.bytes, formattedInput: input }
```
hui-an-yang, line 335: might be a good idea to replace formattedInput to `{ signedContent: {
      branch: `blockhash...`,
      contents: [{kind: OpKind.FAILING_NOOP, arbitrary: `123...`}]
    }`

packages/taquito-beacon-wallet/src/taquito-beacon-wallet.ts
```diff
@@ -182,4 +210,42 @@ export class BeaconWallet implements WalletProvider {
   async clearActiveAccount() {
     await this.client.setActiveAccount();
   }
+
+  async sign(signingRequest: { payload: string }) {
+    const watermarkedBytes = '03' + signingRequest.payload;
```
hui-an-yang, line 215: Might worth giving user the option to sign 05 (SigningType.MICHELINE) too

packages/taquito-beacon-wallet/src/taquito-beacon-wallet.ts
```diff
@@ -182,4 +210,42 @@ export class BeaconWallet implements WalletProvider {
   async clearActiveAccount() {
     await this.client.setActiveAccount();
   }
+
+  async sign(signingRequest: { payload: string }) {
```
hui-an-yang, line 214: Understand  the intention it's good consistency for signer and wallet alike's sign function to return the same object format. Also feel like for people having a front end app with connect wallet (use of wallet api) to sign something they might only make use prefix signature which is returned from `this.client.requestSignPayload(..)`. Can be down to preference, just would like to voice out

packages/taquito-utils/src/verify-signature.ts
```diff
@@ -40,14 +41,19 @@ type SigPrefix = Prefix.EDSIG | Prefix.SPSIG | Prefix.P2SIG | Prefix.SIG;
 export function verifySignature(
```
hui-an-yang, line 41: great work adding watermark to verifySignature 🙌

packages/taquito/src/contract/rpc-contract-provider.ts
```diff
@@ -677,6 +680,27 @@ export class RpcContractProvider extends Provider implements ContractProvider, S
...
+   */
+  async failingNoop(params: FailingNoopParams) {
+    const publicKeyHash = await this.signer.publicKeyHash();
+
+    const prepared = await this.prepare.failingNoop({ ...params });
+    const content = prepared.opOb.contents.find(
+      (op) => op.kind === OpKind.FAILING_NOOP
+    ) as OperationContentsFailingNoop;
+    const opBytes = await this.forge(prepared);
+    const { hash, context, forgedBytes, opResponse } = await this.signAndInject(opBytes);
```
hui-an-yang, line 700: Still don't think we actually want contract api to inject it (understand from integration test is worth injecting), thinking contract api return `{signContents: {...}, signature: ''}` will be good enough

packages/taquito/src/operations/failing-noop-operation.ts
hui-an-yang, line 1: don't think failing_noop op needs it's own class since it's definitely going to fail there will be no hash and results

packages/taquito/src/operations/types.ts
```diff
@@ -82,7 +84,7 @@ export type RPCOpWithSource =
 
 export const isOpWithFee = <T extends { kind: OpKind }>(
   op: T
-): op is withKind<T, Exclude<OpKind, OpKind.ACTIVATION>> => {
+): op is withKind<T, Exclude<Exclude<OpKind, OpKind.ACTIVATION>, OpKind.FAILING_NOOP>> => {
```
hui-an-yang, line 87: unsure if it's needed to exclude failing_noop

packages/taquito/src/prepare/prepare-provider.ts
```diff
@@ -1068,6 +1075,40 @@ export class PrepareProvider extends Provider implements PreparationProvider {
...
+    const hash = await this.getBlockHash(rest.basedOnBlock);
+    const protocol = await this.getProtocolHash();
+
+    this.#counters = {};
+    const headCounter = parseInt(await this.getHeadCounter(pkh), 10);
+
+    const contents = this.constructOpContents(ops, headCounter, pkh);
+
+    return {
+      opOb: {
```
hui-an-yang, line 1103: don't think it should prepared it as `PreparedOperation>` like it's going to be injected

packages/taquito/src/wallet/legacy.ts
```diff
@@ -32,8 +33,20 @@ export class LegacyWalletProvider implements WalletProvider {
...
+  async mapFailingNoopParamsToWalletParams(params: () => Promise<WalletFailingNoopParams>) {
+    return attachKind(await params(), OpKind.FAILING_NOOP);
+  }
+
   async sendOperations(params: WalletParamsWithKind[]) {
     const op = await this.context.batch.batch(params as any).send();
     return op.hash;
   }
+
+  sign(signingRequest: { payload: string }) {
```
hui-an-yang, line 45: might be good to have the option to sign 05 here too

packages/taquito/src/operations/operation-emitter.ts
```diff
@@ -206,6 +206,10 @@ export abstract class OperationEmitter {
               ...op,
               period: currentVotingPeriod?.voting_period.index,
             };
+          case OpKind.FAILING_NOOP:
```
dsawali, line 209: FYI operation-emitter is no longer used and was removed in my most recent PR merge.

packages/taquito/src/wallet/wallet.ts
```diff
@@ -257,6 +259,25 @@ export class Wallet {
...
+   *
+   * @param params operation parameter
+   */
+  async signFailingNoop(params: WalletFailingNoopParams) {
+    const failingOperation = await this.context.prepare.failingNoop(params);
+    const forgeable = this.context.prepare.toForge(failingOperation);
+    const protocolHash =
+      (this.context.proto as unknown as ProtocolsHash) ??
+      (await this.context.readProvider.getNextProtocol('head'));
+    const forger = new LocalForger(protocolHash);
```
hui-an-yang, line 276: might not want to fixated to localforger when forge should use the one in the context can be local, remote or composite forger. the one in the context will be what user set up in TezosToolkit

hui-an-yang: this.context.forger.forge(forgeable)

packages/taquito/src/contract/rpc-contract-provider.ts
dsawali: I kind of agree with An. Even if technically a failing_noop is designed to fail, from a user exp perspective, always having to capture an error just to grab a signature seems like bad user flow.

packages/taquito-beacon-wallet/src/taquito-beacon-wallet.ts
AlirezaHaghshenas: What do you suggest? To not do all the calculations, and only return the `prefixSignature`? Then if we want to send it to the node, we need to compute `sbytes` anyways.

apps/taquito-test-dapp/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "taquito-test-dapp-vite",
```
dsawali, line 2: can we confirm whether changing the name doesn't have any implications on builds or deployments?

integration-tests/failing-noop.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+    });
+
+    it('Verify that the failing_noop signs a text on the genesis block', async (done) => {
+      const signed = await Tezos.wallet.signFailingNoop({
+        arbitrary: "48656C6C6F20576F726C64", // Hello World
+        basedOnBlock: 0,
+      });
+      expect(signed.bytes).toBe('df2788eed43ab680c8a2b79969ce4de93b9768cd2786a85ebdfba90ca7612638110000000b48656c6c6f20576f726c64');
+      expect(signed.sbytes).toBe('df2788eed43ab680c8a2b79969ce4de93b9768cd2786a85ebdfba90ca7612638110000000b48656c6c6f20576f726c648ecac9a2652af8de58f675f336f0105227b408af3ba7c6361aeeb8c0be50b8693e931ba83af2086da963d5380e633de6c68f75a23685d4cd07deb7e4b7b5e03e');
+      // const pk = await Tezos.wallet.getPublicKey();
```
dsawali, line 22: probably can remove this comment

packages/taquito/src/operations/types.ts
AlirezaHaghshenas: I think it does. We will have a build time error without that

packages/taquito/src/prepare/prepare-provider.ts
AlirezaHaghshenas: Isn't preparing a prerequisite of signing?

packages/taquito/src/operations/types.ts
```diff
@@ -550,6 +552,22 @@ export interface RPCSmartRollupOriginateOperation {
...
+  kind: OpKind.FAILING_NOOP;
+  arbitrary: string;
+}
+
+/**
+ * @description Parameters for the `failingNoop` method
+ */
+export interface FailingNoopParams {
+  arbitrary: string;
+  basedOnBlock: BlockIdentifier;
```
dsawali, line 568: not really a change request, but I thought we want to point to genesis block instead? 

maybe we could make it an optional param that defaults to 0? what do you think?

packages/taquito/src/wallet/legacy.ts
```diff
@@ -32,8 +33,35 @@ export class LegacyWalletProvider implements WalletProvider {
...
+
   async sendOperations(params: WalletParamsWithKind[]) {
     const op = await this.context.batch.batch(params as any).send();
     return op.hash;
   }
+
+  async sign(signingRequest: { payload: string; signingType: 'operation' | 'micheline' | 'raw' }) {
+    let magicByte: Uint8Array | undefined;
+    switch (signingRequest.signingType) {
+      case 'micheline':
```
dsawali, line 48: slight nitpick, but we could maybe make this an enum along with 'operation' and 'raw' as SigningType?

apps/taquito-test-dapp/src/tests.ts
```diff
@@ -678,6 +702,18 @@ export const init = (
     inputType: "string",
     lastResult: { option: "none", val: false }
   },
+  {
+    id: "sign-failingNoop",
+    name: "Sign the provided payload in a failing noop",
+    description: "This test signs the payload provided by the user wrapped in a failing noop",
+    documentation: 'https://tezostaquito.io/docs/signing/#generating-a-signature-with-beacon-sdk',
```
hui-an-yang, line 709: would be great to have a failing_noop file under doc in our website to reference here

integration-tests/failing-noop.spec.ts
hui-an-yang, line 1: It’s a bit particular, am hoping you don’t mind breaking the integration tests into different files. 
For example operations supporting in contract api that we’ll have a single file naming as `integration-tests/contract-increase-paid-storage-operation.spec.ts` and for wallet api as `integration-tests/wallet-increase-paid-storage-operation.spec.ts` where increase-paid-storage can be replaced by failing-noop. 
And for the forging that we have a dedicated test file `integration-tests/local-forging.spec.ts` where all forging test case will be collected in file `integration-tests/data/allTestsCases.ts`
Will be best that we can keep the convention of file structure that it'll be helpful in the future when we re-organize integration tests files into folders, hope to be understand.

packages/taquito/src/prepare/prepare-provider.ts
hui-an-yang: failing_noop operation is only signing an object like this, will be good enough to prepared as is that will be ready to be signed. 
There is no counter, protocolhash needed needed, doesn't have to use `this.constructOpContents`
```
{
        branch: "B...",
        contents: [
          {
            kind: OpKind.FAILING_NOOP,
            arbitrary: "1234..."
          },
        ],
}
```

packages/taquito/src/wallet/wallet.ts
```diff
@@ -257,6 +258,37 @@ export class Wallet {
...
+   *
+   * @description failing_noop operation that is guaranteed to fail.
+   *
+   * @returns Signature for a failing_noop
+   *
+   * @param params operation parameter
+   */
+  async signFailingNoop(params: WalletFailingNoopParams) {
+    const failingOperation = await this.context.prepare.failingNoop(params);
+    const forgeable = this.context.prepare.toForge(failingOperation);
```
hui-an-yang, line 271: like above when prepare already returning an object ready to forge that there is no need of .toForge(). Can be discussed

apps/taquito-test-dapp/package.json
hui-an-yang: deploying taquito-test-dapp currently failing, need to investigate if it's related to this name change?

packages/taquito-beacon-wallet/src/taquito-beacon-wallet.ts
```diff
@@ -182,4 +197,35 @@ export class BeaconWallet implements WalletProvider {
   async clearActiveAccount() {
     await this.client.setActiveAccount();
   }
+
+  async sign(signingRequest: { payload: string; signingType: 'operation' | 'micheline' | 'raw' }) {
```
hui-an-yang, line 201: don't think we known enough how client.requestSignPayload() works exactly, I am not to confident that we abstract this function in BeaconWallet class.  (reference documentation how to sign Michelin where the bytes need to be padded added length and and more than prefix 05) https://tezostaquito.io/docs/signing#generating-a-signature-with-beacon-sdk

integration-tests/failing-noop.spec.ts
```diff
@@ -74,44 +73,5 @@ CONFIGS().forEach(({ lib, rpc, setup, protocol }) => {
       done();
     });
 
-    it('Verify that the local forger can forge the failing_noop operation on the genesis block', async (done) => {
```
hui-an-yang: we still need to test forger, can you add a test case in `integration-tests/data/allTestsCases.ts`

packages/taquito/src/contract/rpc-contract-provider.ts
hui-an-yang, line 1: overall, good work for the change 🙌

packages/taquito/src/wallet/wallet.ts
hui-an-yang, line 1: overall, good work for the change 🙌. Same as contract API, small suggestion will be great to define type to variable instead of type casting.

docs/failing_noop.md
hui-an-yang, line 1: appreciate adding the doc!

apps/taquito-test-dapp/src/tests.ts
```diff
@@ -8,8 +8,7 @@ import {
 import type { ContractProvider } from "@taquito/taquito";
 import type { BeaconWallet } from "@taquito/beacon-wallet";
 import { char2Bytes, verifySignature } from "@taquito/utils";
-import type { RequestSignPayloadInput } from "@airgap/beacon-sdk";
-import { SigningType } from "./types";
+import { SigningType, type RequestSignPayloadInput } from "@airgap/beacon-sdk";
```
hui-an-yang, line 11: is RequestSignPayloadInput use anywhere in our code base?

integration-tests/contract-failing-noop.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+            arbitrary: '48656C6C6F20576F726C64'
+          }]
+        }
+      });
+      const pk = await Tezos.wallet.getPublicKey();
+      expect(verifySignature(signed.bytes, pk!, signed.signature, new Uint8Array([3]))).toBe(true);
+      done();
+    });
+
+    it('Verify that the contract.failingNoop signs a text', async (done) => {
```
hui-an-yang, line 38: might want to update the description to be signing on head block

integration-tests/wallet-failing-noop.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+            arbitrary: '48656C6C6F20576F726C64'
+          }]
+        }
+      });
+      const pk = await Tezos.wallet.getPublicKey();
+      expect(verifySignature(signed.bytes, pk!, signed.signature, new Uint8Array([3]))).toBe(true);
+      done();
+    });
+
+    it('Verify that the wallet.signFailingNoop signs a text', async (done) => {
```
hui-an-yang, line 38: same can update the description to fit the testing scenario

packages/taquito-beacon-wallet/src/taquito-beacon-wallet.ts
```diff
@@ -182,4 +185,37 @@ export class BeaconWallet implements WalletProvider {
...
+
+  async sign(signingRequest: { payload: string; watermark?: Uint8Array }) {
+    let bb = hex2buf(signingRequest.payload);
+    if (typeof signingRequest.watermark !== 'undefined') {
+      bb = mergebuf(signingRequest.watermark, bb);
+    }
+    const watermarkedBytes = buf2hex(toBuffer(bb));
+
+    let signingType: SigningType;
+    switch (signingRequest.watermark) {
```
hui-an-yang, line 197: signingType is empty don't think the switch will work

packages/taquito/src/wallet/interface.ts
```diff
@@ -34,10 +45,22 @@ export interface WalletProvider {
...
+
+  /**
+   * @description Request the wallet to sign a payload
+   */
+  sign(signingRequest: { payload: string; watermark?: Uint8Array }): Promise<string>;
+
+  /**
+   * @description Get the public key from the wallet
+   */
+  getPublicKey(): Promise<string | undefined>;
```
hui-an-yang, line 65: might not have the context, will this ever get undefined cuz get pkh doesn't that made me wonder should we change the pkh too or keep pkh and make this return string for sure

website/sidebars.js
hui-an-yang, line 1: thanks for sorting the order out!!

dsawali (approved):

integration-tests/contract-failing-noop.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+          signature: 'edsigteDrJhTBCX8cX8swZPEeT7N96wjUxWfFPJSEFL5s5pUT2C3C2FDNF169zYYKktrZHwpHcWan6uyKj89tK43fqY2hbbKoxT',
+          signedContent: {
+            branch: await Tezos.rpc.getBlockHash({block: 'genesis'}),
+            contents: [{
+              kind: OpKind.FAILING_NOOP,
+              arbitrary: '48656C6C6F20576F726C64'
+            }]
+          }
+        });
+      } else {
```
hui-an-yang, line 34: to handle assertion based on `isSandbox` is limited to one other testnet. Find it easier to manage if we can separate into test suites named by testnet like this `const nairobinet = protocol == Protocols.PtNairobi ? it : it.skip;`   and add a comment behind branch that genesis block hash differ by by chain. When new testnet introduced that we can save time debugging why this test isn't happy

PR #2609, opened 2023-08-18
fix: fixed idx calculation in pairs with mix of annotated and unanntated children

### LINKED ISSUE #2540: 3 level Michelson structure missing elements when transformed to JS in for PairToken
**Description**
The transformation from JSON encoded Micheline value into JS representation skips flattens 3 level schema into 2 levels and skips an element.

**Steps To Reproduce**
This bug manifests when we have three-levels of pair nesting:
- Pair1
   - Pair2
      - Pair3
         - Int
         - Int
         - %A
      - Int
   - bool
   
Pair1, Pair2, Pair3 return PairToken.Execute Object1, Object2, Object3 results respectively. 
Pair2 knows that Pair3 is annotated with “A3'' and Pair2.Execute returns Object2, a JS object with a nested JS object accessed via the A3 key. However, Pair2 erroneously establishes an idx value to its rightValue as if both Object3 keys are visible at the Object2 level (i.e., an idx=2). When Pairing1 retrieves a Pairing2 result it correctly establishes an idx value for its rightValue based on the number of keys contained in Object 2 (once again, an idx=2). As a result, the rightValue.Execute result for Pair1 may contribute a value for the “2” key which overwrites a “2” key value contributed by Pair2.

**Expected behavior**
Micheline pairs correspond to a kind of record structure. As such they are mapped to JavaScript (JS) objects. In the case where an unannotated pair is referenced by a parent pair, the nesting is abstracted away in the mapping of a Micheline Pair to a JS object. In the case where an annotated pair is referenced by a parent pair, the nesting is preserved in the mapping of a Micheline Pair to a JS object; a new JS object is nested in the parent object with the Micheline pair annotation serving as the parent object key. 

Pseudo code for PairToken.Execute( val, idx)

```
keyCount = 1;
typ1.idx = idx;

If typ1 is a PairToken with no annotations then    	% abstract away unannotated pair nesting
 	leftValue = typ1.Execute(val1);                      
 	keyCount= Object.keys(leftValue).length;	% count leftValue keys
else
 	leftValue = {[typ1.annot()]: typ1.Execute(val1)};    % wrap leftValue with its annotation
endif;

typ2.idx = idx + keyCount;
If typ2 is a PairToken with no annotations then    	% abstract away unannotated pair nesting
 	rightValue = typ2.Execute(val2);   
else
 	rightValue = {[typ2.annot()]: typ2.Execute(val2)}; % wrap rightValue with its annotation
endif;
 
return { …leftValue, …rightValue};             	% merge left, right values into a single object
```

PairToken idx management ensures that mapped-to objects have their own default property namespace. This corresponds naturally with the notion of object property namespaces in JS. 

**Additional context**
A coding experiment was conducted to illustrate this. 

For a JSON encoded Micheline schema,
```
{prim: "pair", args: [{prim: "pair", args: [{prim: "pair", 
    args: [{prim: "int"}, {prim: "int"}],annots: ["%A3"]}, 
    {prim: "int"}]}, {prim: "bool"}]}

```
We constructed the JSON encoded Micheline value,
```
const MichelsonValue = { prim: "Pair", args:[{prim: "Pair",
    args: [{prim: "Pair",args: [{int: "11"},{int: "22"}]},
    {int: "33"}]}, {prim: "True"}]};
```

And obtained a result for 
`const JSValue = storageSchema.Execute(MichelsonValue); `

namely:
```
{
  "2": true,
  "A3": {
    "0": "11",
    "1": "22"
  }
}
```

Here we can see that the Boolean true value from the rightValue.Execute for Pair1 has contributed a value for the “2” key which has overwritten the “2” key value (“33”) contributed by Pair2. 

The correct result is:
```
{
“1”: “33”,  
"2": true,
  "A3": {
    "0": "11",
    "1": "22"
  }
}
```

Request access to the google document when working on this issue from Inna

AlirezaHaghshenas:
closes #2540 

## Release Note Draft Snippet

### DIFF
--- a/packages/taquito-michelson-encoder/src/tokens/pair.ts
+++ b/packages/taquito-michelson-encoder/src/tokens/pair.ts
@@ -169,11 +169,11 @@ export class PairToken extends ComparableToken {
     const leftToken = this.createToken(args[0], this.idx);
     let keyCount = 1;
     let leftValue;
-    if (leftToken instanceof PairToken) {
-      keyCount = Object.keys(leftToken.ExtractSchema()).length;
-    }
     if (leftToken instanceof PairToken && !leftToken.hasAnnotations()) {
       leftValue = getLeftValue(leftToken);
+      if (leftToken instanceof PairToken) {
+        keyCount = Object.keys(leftToken.ExtractSchema()).length;
+      }
     } else {
       leftValue = { [leftToken.annot()]: getLeftValue(leftToken) };
     }
--- a/packages/taquito-michelson-encoder/test/tokens/constant.spec.ts
+++ b/packages/taquito-michelson-encoder/test/tokens/constant.spec.ts
@@ -79,7 +79,7 @@ describe('Global constant token', () => {
       } catch (e: any) {
         expect(e).toBeInstanceOf(GlobalConstantDecodingError);
         expect(e.message).toEqual(
-          `[2] Unable to decode a value represented by a global constants. Please provide an expanded script to the Michelson-Encoder or semantics for the decoding. The following global constant hash was encountered: expru5X5fvCer8tbRkSAtwyVCs9FUCq46JQG7QCAkhZSumjbZBUGzb.`
+          `[1] Unable to decode a value represented by a global constants. Please provide an expanded script to the Michelson-Encoder or semantics for the decoding. The following global constant hash was encountered: expru5X5fvCer8tbRkSAtwyVCs9FUCq46JQG7QCAkhZSumjbZBUGzb.`
         );
       }
     });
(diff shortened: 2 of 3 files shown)


packages/taquito-michelson-encoder/src/tokens/pair.ts
```diff
@@ -169,11 +169,11 @@ export class PairToken extends ComparableToken {
     const leftToken = this.createToken(args[0], this.idx);
     let keyCount = 1;
     let leftValue;
-    if (leftToken instanceof PairToken) {
-      keyCount = Object.keys(leftToken.ExtractSchema()).length;
-    }
     if (leftToken instanceof PairToken && !leftToken.hasAnnotations()) {
       leftValue = getLeftValue(leftToken);
+      if (leftToken instanceof PairToken) {
```
dsawali, line 174: can you give a little explanation on why it's moved into the if block? for my own reference

AlirezaHaghshenas: From Peter's document:

> The Pseudo code corresponding to the actual implementation differs. In managing the idx keyCount, it counts keys regardless of whether a nested pair has annotation or not.

dsawali (approved):

PR #2611, opened 2023-08-22
Fixes to OrToken

### LINKED ISSUE #2543: Two levels of nesting for OrToken
**Description**
OrToken.Encode and OrToken.Execute  mapping rules both rely on OrToken.ExtractSchema for managing idx. The implication is that not only does OrToken.ExtractSchema offer poor guidance for a developer, but its erroneous result leads to OrToken.Execute and OrToken.Encode execution errors. 

**Steps To Reproduce**
Consider the following:
```
Or1
   Or2
      Or3
      int
      string
      %A
   Bool
   bytes
```

Or1.ExtractSchema() interprets this as:
`{A: {“0”: int, “1”: string}, “1”: bool, “2”: bytes}`

However, Or1.Execute(val) maps conformant JSON-encoded Micheline values to a flat object with a single key.
```
Left (Left (Left 5)) maps to {“0”: 5}
Left (Left (Right “foo”)) maps to {“1”: “foo”}
Left (Right True) maps to {“2”: true}
Right 0xABCF45 maps to {“2”: “ABCF45”}
```

Here we see that on the JS side the property key “2” is erroneously overloaded. This begs the question, “how does Or1.Encode properly map the “2” key to a single Left-Right Or1 position?”. 

Or1.Encode ({“2”: “ABCF45”}) maps to a JSON-encoded version of Right 0xABCF45. However, Or1.Encode( {“2”: true}) throws an invalid-bytes error. Or1.Encode interprets the “2” key as equivalent to the Right OR1 branch rather than as the Left (Right) OR1 branch. As a result, there is no way to encode a JS true value even though this is allowed by Or1. 

**Expected behavior**
The correct way to implement the one-of constraint on the JS side, is to ignore nested or annotation. An easy fix is to change the OrToken.ExtractSchema mapping rule so that nested ORs are abstracted away regardless of whether they are annotated or not. This would align ExtractSchema with similar logic found in both the OrToken.Execute(val) and the OrToken.Encode(val) mapping rules. 

**Additional context**
AS-IS OrToken document has suggestions for a fix

### LINKED ISSUE #2544: Encode accepts non-conformant JS Values for OrToken
**Description**
In the spirit of enforcing type checking on the JS side that is equivalent to that on the Micheline side, OrToken.Encode(val) should only accept an object with a single key, one that corresponds to a Micheline Left-Right position, and one whose type is consistent with the Micheline type found in that Left-Right position of the root OrToken.

The current OrToken.Encode(val) accepts objects with multiple keys. It traverses the root OrToken looking for non-or Token with a matching key. Processing terminates with the first one found. This allows a developer to erroneously conclude that they have successfully encoded a JS object with multiple keys, as an equivalent JSON encoded Micheline value. In the spirit of tight type checking any attempt to OrToken.Encode a val with more than one key should fail. 

**Steps To Reproduce**
Consider the following:
```
Or1
   Or2
      int
      string
      %A
   Bool
```

Or1.Execute(val) maps conformant JSON-encoded Micheline values to a flat object with a single key.
```
Left (Left 5) maps to {“0”: 5}
Left (Right “foo”) maps to {“1”: “foo”}
Right True maps to {“2”: true}
```

However, Or1.Encode( {“0”: 5, “1”: “foo”, “2”: true}) returns a JSON-encoded value for: 
Left (Left 5). The other key-value pairs of are simply ignored. A correct result would be for the Encode to raise an invalid-type error. 

**Expected behavior**
Add logic to OrToken.Encode(val) to ensure that any attempt to encode a val with more than one key fails.

**Additional context**
Refer to AS-IS OrToken document

AlirezaHaghshenas:
closes #2543 
closes #2544 

## Release Note Draft Snippet


dsawali:
@AlirezaHaghshenas could you provide some docs with the changes? Highlighting some of the behaviour changes if any

packages/taquito-michelson-encoder/src/errors.ts
```diff
@@ -3,7 +3,7 @@ export { MapValidationError } from './tokens/map';
 export { SetValidationError } from './tokens/set';
 export { BigMapValidationError } from './tokens/bigmap';
 export { TokenArgumentValidationError, TokenComparisonError } from './tokens/pair';
-export { OrTokenDecodingError } from './tokens/or';
+export { OrTokenDecodingError, OrValidationError as OrValidationError } from './tokens/or';
```
hui-an-yang, line 6: curious the reason of casting as itself here?

packages/taquito-michelson-encoder/src/tokens/or.ts
```diff
@@ -13,6 +20,17 @@ export class OrTokenDecodingError extends TaquitoError {
   }
 }
 
+/**
+ *  @category Error
+ *  @description Error that indicates a failure happening when parsing encoding/executing an OrToken
```
hui-an-yang, line 25: suggestion only, from this description seems where throw OrTokenDecodingError can be replaced instead. (complying with structure of other tokens files)

packages/taquito-michelson-encoder/test/tokens/or.spec.ts
```diff
@@ -3,348 +3,1254 @@ import { createToken } from '../../src/tokens/createToken';
...
+  let tokenComplex: OrToken;
+  let tokenComplexNoAnnots: OrToken;
+  let tokenNestedOr: OrToken;
+  let token2LevelOrMixedAnnots: OrToken;
+  let token3LevelOrMixedAnnots: OrToken;
+  let tokenNestedOrWithoutAnnot: OrToken;
+  let tokenOrWithOption: OrToken;
+  beforeEach(() => {
+    token = createToken(
+      {
```
hui-an-yang, line 17: suggestion only, i found the format of json Michelson in one line will be easier to comprehend what this test is really about.

packages/taquito-michelson-encoder/test/tokens/or.spec.ts
```diff
@@ -3,348 +3,1254 @@ import { createToken } from '../../src/tokens/createToken';
...
+  });
 
-    describe('EncodeObject', () => {
-        it('Should encode properly', () => {
-            expect(token.EncodeObject({ intTest: 10 })).toEqual({ prim: 'Left', args: [{ int: '10' }] });
-            expect(token.EncodeObject({ stringTest: '10' })).toEqual({ prim: 'Right', args: [{ string: '10' }] });
+  describe('EncodeObject', () => {
+    it('Should throw unless input is an object with single key', () => {
+      expect(() => tokenNoAnnots.EncodeObject(undefined)).toThrow(
+        'EncodeObject expectes an object with a single key but got:'
```
hui-an-yang, line 297: 6 typo expects

packages/taquito-michelson-encoder/test/tokens/or.spec.ts
```diff
@@ -3,348 +3,1254 @@ import { createToken } from '../../src/tokens/createToken';
...
+      const schema2LevelMixedannots = token2LevelOrMixedAnnots.generateSchema();
+      expect(schema2LevelMixedannots).toEqual({
+        __michelsonType: 'or',
+        schema: {
+          '0': { __michelsonType: 'int', schema: 'int' },
+          '2': { __michelsonType: 'bool', schema: 'bool' },
+          A: { __michelsonType: 'nat', schema: 'nat' },
+        },
+      });
+      const leftLeft = {
```
hui-an-yang, line 326: might be a good idea to put leftLeft, leftRight and right separately right above it's own expect assertion like token3LevelOrMixedAnnots examples line374

packages/taquito-michelson-encoder/test/tokens/or.spec.ts
```diff
@@ -3,348 +3,1254 @@ import { createToken } from '../../src/tokens/createToken';
...
+        3: 'bool',
+      });
+      schema3LevelMixedannots = token3LevelOrMixedAnnots.ExtractSignature();
+      expect(schema3LevelMixedannots).toEqual([
+        ['0', 'bytes'],
+        ['1', 'int'],
+        ['2', 'nat'],
+        ['3', 'bool'],
+      ]);
+      const leftLeftLeft = { prim: 'Left', args: [{ bytes: '5674' }] };
```
hui-an-yang, line 374: thought this is jus left

packages/taquito-michelson-encoder/test/tokens/or.spec.ts
```diff
@@ -3,348 +3,1254 @@ import { createToken } from '../../src/tokens/createToken';
...
+      });
+      const rightLeftRight = {
+        prim: 'Right',
+        args: [{ prim: 'Left', args: [{ prim: 'Right', args: [{ int: '10' }] }] }],
+      };
+      expect(token3LevelOrMixedAnnots.EncodeObject({ 2: 10 })).toEqual(rightLeftRight);
+      expect(token3LevelOrMixedAnnots.Execute(rightLeftRight)).toEqual({
+        2: BigNumber(10),
+      });
+      const righRigh = {
```
hui-an-yang, line 395: nitpick typo righRight

packages/taquito-michelson-encoder/test/tokens/or.spec.ts
```diff
@@ -3,348 +3,1254 @@ import { createToken } from '../../src/tokens/createToken';
...
+            annots: [],
+          },
+        ],
+        annots: [],
+      },
+      0
+    ) as OrToken;
+  });
 
-    describe('EncodeObject', () => {
```
hui-an-yang, line 23: can we structure EncodeObject test suite like Encode for readability?
image

packages/taquito-michelson-encoder/src/errors.ts
AlirezaHaghshenas: I think this has happened during an automatic rename. I removed it. Thanks

hui-an-yang (approved):
great work 🌟

Issue #1690, opened 2022-06-10
Sometimes operations are ignored and confirmation() fails with "confirmation polling timed out"


852Kerfunkle:
Hello,

Basically, sometimes operations don't seem to be accepted by rpc nodes and they time out with "confirmation polling timed out" after some minutes.

**Steps To Reproduce**
The easiest way to reproduce this seems to be to send an operation with insufficient gas (manually reduce the gas limit in temple to 20% of the estimate):

```js
const batch_op = await some_wallet_operation_batch.send();
const op_res = await batch_op.confirmation(); // throws "confirmation polling timed out"
const op_status = await batch_op.status();
// check status etc
```

**Expected behavior**
I'd expect to get some kind of error when the operation is rejected by the rpc node. (I think that's what's happening here - not sure).

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: Chromium
 - Version: 12.1

claudebarde:
Most of the time, when this error happens, it means that the node took longer than expected to send the confirmation.
You can change the timeout value this way:
```
const Tezos = new TezosToolkit(RPC_URL);
Tezos.setProvider({
    config: { streamerPollingIntervalMilliseconds: VALUE_IN_MILLISECONDS }
});
```
in order to give more time to Taquito to wait for a response.

852Kerfunkle:
@claudebarde I assume you meant `confirmationPollingTimeoutSecond`?

Thing is, the operation is never even added to the mempool (I think) - so I could set the timeout to hours and it would still happen.

If it seems convoluted to force a failure by setting the gas way too low, it's just a way I managed to to find to reproduce this behaviour. It does seem to happen in the wild - but it's difficult to reproduce. It's conceivable increasing the timeout might help there, I suppose.

Innkst:
@852Kerfunkle do you still experience this issue or configuring timeout helped? Thanks

852Kerfunkle:
@Innkst I just assumed the timeout would make it fail faster, but I feel that's just a workaround for the underlying issue here.

I didn't look at the details, but I would assume (maybe incorrectly) the node would give some kind of error when it plainly rejects an operation - it certainly does sometimes, just not in this case. I don't know if this is an RPC node issue or a taquito issue, but it sure is an issue that I occasionally experience under "normal" circumstances.

852Kerfunkle:
Oh, by the way - if my memory serves me right, it's also reproducible in a flextesa sandbox.

[...]

PR #2634, opened 2023-09-01
Reduce jest maxWorkers to 4

danielelisi:
Ref #1690  

Reducing Jest maxWorkers to 4 decreases the chance to hit the ConnectionTimeout errors when running integration tests


hui-an-yang (commented):
✅

PR #2657, opened 2023-09-14
implement staking in oxford

### LINKED ISSUE #2608: Support stake, unstake, and finalize_unstake pseudo operations in Taquito
Protocol Oxford will bring forth a feature flag, that when enabled, will allow users to perform stake, unstake, and finalize_unstake pseudo operations on implicit accounts.

Support the 3 pseudo operations in Taquito Contract API.

### Implementation Details
Stake, unstake, and finalize_unstake are all considered 'pseudo operations' because they are essentially an extension of the existing transaction operation.

Refer to this document for further details on what that means: https://hackmd.io/@dsawali/H1_vK4t33

Implement a method wrapper in contract API similar to `contractCall` or `setDelegate` and `registerDelegate`. The work will encompass these domains:
- Prepare provider
- RPC contract provider
- Estimate provider
- Local forger

The method wrappers should be callable via the contract API as such:
```typescript
await Tezos.contract.stake(...);

await Tezos.contract.unstake(...);

await Tezos.contract.finalizeUnstake(...);
```
note: same application for estimate, and prepare provider

The same goes for wallet API (but need verification on this).

### Testing
Write appropriate unit and integration tests. This will be part of the ticket's acceptance criteria.

ac10n:
closes #2608 

## Release Note Draft Snippet


packages/taquito/src/operations/types.ts
```diff
@@ -568,6 +568,16 @@ export interface FailingNoopParams {
   basedOnBlock: BlockIdentifier;
 }
 
+export interface StakingParams {
```
hui-an-yang, line 571: can consider adding `parameter?: TransactionOperationParameter` will make this interface more versatile

packages/taquito/src/contract/rpc-contract-provider.ts
```diff
@@ -764,6 +765,75 @@ export class RpcContractProvider extends Provider implements ContractProvider, S
 
     return batch;
   }
+
+  /**
+   *
+   * @description Stake founds. Will sign and inject an operation using the current context
```
hui-an-yang, line 771: octez-client description `Stake the given amount for the source. The source must be a delegator to be allowed to stake.`

packages/taquito/src/contract/rpc-contract-provider.ts
```diff
@@ -764,6 +765,75 @@ export class RpcContractProvider extends Provider implements ContractProvider, S
...
+   *
+   * @returns An operation handle with the result from the rpc node
+   *
+   * @param stake operation parameter
+   */
+  async stake(params: StakingParams) {
+    const estimate = await this.estimate(params, this.estimator.stake.bind(this.estimator));
```
hui-an-yang, line 778: might want to validate the amount value being passed in as params, if it's less than zero should throw InvalidAmountError

packages/taquito/src/operations/types.ts
hui-an-yang, line 571: You asked yesterday why there is `source?: `in transferParam that it's for people just want to use taquito to prepare the op without any signer or wallet configured, i think it's also needed for `StakingParams` to have optional source

packages/taquito/src/contract/rpc-contract-provider.ts
```diff
@@ -764,6 +765,75 @@ export class RpcContractProvider extends Provider implements ContractProvider, S
...
+    const source = await this.signer.publicKeyHash();
```
hui-an-yang, line 779: with previous comment that `StakingParmas` having optional source this line can be replaced by `const publicKeyHash = await this.signer.publicKeyHash();
const source = params.source ?? publicKeyHash;`

packages/taquito/src/contract/rpc-contract-provider.ts
```diff
@@ -764,6 +765,75 @@ export class RpcContractProvider extends Provider implements ContractProvider, S
...
+
+    const prepared = await this.prepare.stake({ ...params, ...estimate });
+    const content = prepared.opOb.contents.find(
+      (op) => op.kind === OpKind.TRANSACTION
+    ) as OperationContentsTransaction;
+
+    const opBytes = await this.forge(prepared);
```
hui-an-yang, line 786: It wasn't mentioned in this ticket, think it's worth adding cases of stake, unstake and finalized_unstake in `integration-tests/data/allTestsCases.ts` to test our forger.forge() function in `integration-tests/local-forging.spec.ts`

packages/taquito/src/estimate/rpc-estimate-provider.ts
```diff
@@ -457,4 +458,64 @@ export class RPCEstimateProvider extends Provider implements EstimationProvider
...
+  /**
+   *
+   * @description Estimate gasLimit, storageLimit and fees for stake operation
+   *
+   * @returns An estimation of gasLimit, storageLimit and fees for the stake operation
+   *
+   * @param Estimate
+   */
+  async stake(params: StakingParams) {
+    const protocolConstants = await this.context.readProvider.getProtocolConstants('head');
```
hui-an-yang, line 471: Same as `rpc-contract-provider.ts` we will want to validate amount. Estimate Provider can be use solely we need to validate what's being passed separately

packages/taquito/src/estimate/rpc-estimate-provider.ts
```diff
@@ -457,4 +458,64 @@ export class RPCEstimateProvider extends Provider implements EstimationProvider
...
+
+  /**
+   *
+   * @description Estimate gasLimit, storageLimit and fees for stake operation
+   *
+   * @returns An estimation of gasLimit, storageLimit and fees for the stake operation
+   *
+   * @param Estimate
+   */
+  async stake(params: StakingParams) {
```
hui-an-yang, line 470: Same as rpc-contract-provider.ts we will want to validate amount. Estimate Provider can be use solely we need to validate what's being passed separately

packages/taquito/src/contract/prepare.ts
```diff
@@ -339,3 +341,32 @@ export const createSmartRollupOriginateOperation = async ({
...
+export const createStakingOperation = async (
+  {
+    amount,
+    fee = DEFAULT_FEE.TRANSFER,
+    gasLimit = DEFAULT_GAS_LIMIT.TRANSFER,
+    storageLimit = DEFAULT_STORAGE_LIMIT.TRANSFER,
+    mutez = false,
+  }: StakingParams,
+  pkh: string,
+  entrypoint: StakingEntrypoint
```
hui-an-yang, line 354: if `stakingParmas` has source and parameter that we don't need to pass them in outside of the first object

packages/taquito/src/operations/types.ts
AlirezaHaghshenas: The reason I prefer not to add that is: it will be a less elegant abstraction. If we have that, then in wallet API's `stake` method, we have to either:

1. ignore it
2. make it required
3. make it optional

I guess you mean option 3. So what happens if the user sets it, but sets it to the wrong value? We should throw and error. This is less elegant than an API that does not allow the user to make mistakes in the first place.

packages/taquito/src/prepare/prepare-provider.ts
```diff
@@ -1098,4 +1101,78 @@ export class PrepareProvider extends Provider implements PreparationProvider {
...
   }
+
+  /**
+   *
+   * @description Method to prepare a stake operation
+   * @param operation RPCOperation object
+   * @param source string or undefined source pkh
+   * @returns a PreparedOperation object
+   */
+  stake(operation: StakingParams, source?: string): Promise<PreparedOperation> {
```
hui-an-yang, line 1112: previous comments that would be great that source to be optional in StakingParams. 
it's fine if you don't like to follow the rest of ops defined parameter {fee, storageLImit, gasLImit, ...rest}, think the `operation: StakingParams` of operation can be better named.

packages/taquito/src/contract/prepare.ts
```diff
@@ -339,3 +341,32 @@ export const createSmartRollupOriginateOperation = async ({
...
+    destination: pkh,
+    kind: OpKind.TRANSACTION,
+    fee,
+    gas_limit: gasLimit,
+    storage_limit: storageLimit,
+    amount: mutez ? amount.toString() : format('tz', 'mutez', amount).toString(),
+    parameters: {
+      entrypoint,
+      value: {
+        prim: 'Unit',
```
hui-an-yang, line 367: Would like to not hard code a generalized `createStakingOperation` value: {prim: 'Unit'} here, should happened in their individual function. As hardcoded to unit only workable with current ops supported, we need to leave room for more pseudo ops like set_delegate_parameters to pass value in

packages/taquito/src/prepare/prepare-provider.ts
```diff
@@ -1098,4 +1101,78 @@ export class PrepareProvider extends Provider implements PreparationProvider {
...
+   * @returns a PreparedOperation object
+   */
+  async finalizeUnstake(operation: StakingParams, source?: string): Promise<PreparedOperation> {
+    return this.staking(operation, 'finalize_unstake', source);
+  }
+
+  private async staking(
+    { fee, storageLimit, gasLimit, ...rest }: StakingParams,
+    entrypoint: StakingEntrypoint,
+    source?: string
```
hui-an-yang, line 1141: Like previous comments should `source?` be in `StakingParams` and `entrypoint` be under `parameter: TransactionOperationParameter`

packages/taquito/src/prepare/prepare-provider.ts
```diff
@@ -1098,4 +1101,78 @@ export class PrepareProvider extends Provider implements PreparationProvider {
...
+    const protocolConstants = await this.context.readProvider.getProtocolConstants('head');
+    const DEFAULT_PARAMS = await this.getAccountLimits(pkh, protocolConstants);
+    const mergedEstimates = mergeLimits({ fee, storageLimit, gasLimit }, DEFAULT_PARAMS);
+
+    const op = await createStakingOperation(
+      {
+        ...rest,
+        ...mergedEstimates,
+      },
+      pkh,
```
hui-an-yang, line 1154: passed in as `...rest`

packages/taquito/src/prepare/prepare-provider.ts
```diff
@@ -1098,4 +1101,78 @@ export class PrepareProvider extends Provider implements PreparationProvider {
...
+      entrypoint
```
hui-an-yang, line 1155: pkh and entrypoint passed in as `...rest`

packages/taquito/src/contract/interface.ts
```diff
@@ -283,4 +284,34 @@ export interface ContractProvider extends StorageProvider {
    * @param params FailingNoop operation parameter
    */
   failingNoop(params: FailingNoopParams): Promise<FailingNoopOperation>;
+
+  /**
+   *
+   * @description Stake founds. Will sign and inject an operation using the current context
```
hui-an-yang, line 290: octez-client desc: `Stake the given amount for the source. The source must be a delegator to be allowed to stake.`

packages/taquito/src/contract/interface.ts
```diff
@@ -283,4 +284,34 @@ export interface ContractProvider extends StorageProvider {
...
+   *
+   * @returns An operation handle with the result from the rpc node
+   *
+   * @param stake operation parameter
+   */
+  stake(params: StakingParams): Promise<TransactionOperation>;
+
+  /**
+   *
+   * @description Unstake founds. Will sign and inject an operation using the current context
```
hui-an-yang, line 300: octez-client desc - `Unstake the given amount. If "everything" is given as amount, unstakes everything from the staking balance. Unstaked tez remains frozen for a set amount of cycles (the slashing period) after the operation. Once this period is over, the operation "finalize unstake" must be called for the funds to appear in the liquid balance.`

packages/taquito/src/contract/interface.ts
```diff
@@ -283,4 +284,34 @@ export interface ContractProvider extends StorageProvider {
...
+   *
+   * @returns An operation handle with the result from the rpc node
+   *
+   * @param unstake operation parameter
+   */
+  unstake(params: StakingParams): Promise<TransactionOperation>;
+
+  /**
+   *
+   * @description Finalize unstake founds. Will sign and inject an operation using the current context
```
hui-an-yang, line 310: octez-client desc:  `Transfer all the finalizable unstaked funds of the source to their liquid balance.`

packages/taquito/src/contract/interface.ts
```diff
@@ -283,4 +284,34 @@ export interface ContractProvider extends StorageProvider {
...
+   *
+   * @returns An operation handle with the result from the rpc node
+   *
+   * @param finalizeUnstake operation parameter
+   */
+  unstake(params: StakingParams): Promise<TransactionOperation>;
```
hui-an-yang, line 316: finalizeUnstake?

packages/taquito/src/operations/types.ts
hui-an-yang: wallet api can take `parameter?` too, that's how we make contractCall. If anything we should validate stake op if being passed params should be unit or it's by default unit

packages/taquito/src/contract/prepare.ts
AlirezaHaghshenas: That is more code, and if that is ever needed, then we can move it to the individual functions

packages/taquito-core/src/errors.ts
```diff
@@ -77,10 +77,10 @@ export class InvalidBlockHashError extends ParameterValidationError {
  * @description Error that indicates an invalid amount of tez being passed as a parameter
  */
 export class InvalidAmountError extends ParameterValidationError {
-  constructor(public readonly amount: string) {
+  constructor(public readonly amount: string, message?: string) {
     super();
     this.name = 'InvalidAmountError';
-    this.message = `Invalid amount "${amount}"`;
+    this.message = message ?? `Invalid amount "${amount}"`;
```
hui-an-yang, line 83: Error improvement tickets just unified the message syntax of invalidParamError, can we follow 
"this.message += errorDetail ? ` ${errorDetail}.` : '.'; " please? 

We tried to show message as `Invalid amount "0", expecting a positive number`

packages/taquito/src/operations/types.ts
```diff
@@ -568,6 +569,30 @@ export interface FailingNoopParams {
...
+
+/**
+ *
+ * @throws {@link InvalidAmountError}
+ */
+export const validateStakingParams = (params: StakingParams, entrypoint: StakingEntrypoint) => {
+  if (params.amount <= 0) {
+    throw new InvalidAmountError(
+      params.amount.toString(),
+      `The 'amount' field in ${entrypoint} should be positive, but ${params.amount} was provided.`
```
hui-an-yang, line 589: with previous comments in error class, let's pass `expecting a positive number` . If you insist to be verbose can add after the number

packages/taquito/src/contract/interface.ts
```diff
@@ -283,4 +284,34 @@ export interface ContractProvider extends StorageProvider {
    * @param params FailingNoop operation parameter
    */
   failingNoop(params: FailingNoopParams): Promise<FailingNoopOperation>;
+
+  /**
+   *
+   * @description Unstake the given amount. If "everything" is given as amount, unstakes everything from the staking balance. Unstaked tez remains frozen for a set amount of cycles (the slashing period) after the operation. Once this period is over, the operation "finalize unstake" must be called for the funds to appear in the liquid balance. Will sign and inject an operation using the current context
```
hui-an-yang, line 290: desc is for unstake didn't match op stake

packages/taquito/src/contract/rpc-contract-provider.ts
hui-an-yang: bump

packages/taquito/src/contract/rpc-contract-provider.ts
```diff
@@ -764,6 +766,87 @@ export class RpcContractProvider extends Provider implements ContractProvider, S
 
     return batch;
   }
+
+  /**
+   *
+   * @description Stake founds. Will sign and inject an operation using the current context
```
hui-an-yang, line 772: desc needs to be update as `packages/taquito/src/contract/interface.ts` too

packages/taquito/src/operations/types.ts
hui-an-yang: it's only validating amount now, abstraction and the name validatingStakingm might reduce readability

packages/taquito/src/contract/prepare.ts
hui-an-yang: We can make it scalable now, would like to understand what's the reason to postpone?
Op set_delegate_parameters do need to pass param value in, even we don't look to support it in current release. It's a fact tezos have intention to pass param value with pseudo operations.

packages/taquito/src/operations/types.ts
```diff
@@ -568,6 +569,30 @@ export interface FailingNoopParams {
   basedOnBlock: BlockIdentifier;
 }
 
+export interface StakingParams {
+  sourceAndDestination?: string;
```
hui-an-yang, line 573: personal preference, I'll respect tezos naming `source` like octez-client

packages/taquito/src/operations/types.ts
```diff
@@ -568,6 +569,30 @@ export interface FailingNoopParams {
   basedOnBlock: BlockIdentifier;
 }
 
+export interface StakingParams {
```
hui-an-yang, line 572: considering naming `pseudoOpParams`, `pseudoMethodParams` or `pseudoParams`?

packages/taquito/src/operations/types.ts
hui-an-yang: validation util function can reference how other functions have import `validateAddress` from `@taquito/util`

packages/taquito/src/prepare/prepare-provider.ts
```diff
@@ -1098,4 +1101,75 @@ export class PrepareProvider extends Provider implements PreparationProvider {
...
+   *
+   * @description Method to prepare a finalize_unstake operation
+   * @param params RPCOperation object
+   * @returns a PreparedOperation object
+   */
+  async finalizeUnstake(params: StakingParams): Promise<PreparedOperation> {
+    return this.staking(params, 'finalize_unstake');
+  }
+
+  private async staking(
```
hui-an-yang, line 1137: think this function can be public and called directly from, contract-provider and estimate-provider

packages/taquito/src/prepare/prepare-provider.ts
```diff
@@ -1098,4 +1101,75 @@ export class PrepareProvider extends Provider implements PreparationProvider {
...
+    { fee, storageLimit, gasLimit, sourceAndDestination, ...rest }: StakingParams,
+    entrypoint: StakingEntrypoint
+  ): Promise<PreparedOperation> {
+    const { pkh } = await this.getKeys();
+
+    const protocolConstants = await this.context.readProvider.getProtocolConstants('head');
+    const DEFAULT_PARAMS = await this.getAccountLimits(pkh, protocolConstants);
+    const mergedEstimates = mergeLimits({ fee, storageLimit, gasLimit }, DEFAULT_PARAMS);
```
hui-an-yang, line 1145: before this line think we can call a util function toTransferParams() like contract call, the function need to be created too. It will look similar like 

```

    /**
     *
     * @description Create transfer params to be used with TezosToolkit.contract.transfer methods
     *
     * @param Options generic transfer operation parameters
     */
    toTransferParams({
        fee,
        gasLimit,
        storageLimit,
        source,
        amount = 0,
        mutez = false,
    }: Partial<SendParams> = {}): TransferParams {
        const fullTransferParams: ExplicitTransferParams = {
            to: this.address,
            amount,
            fee,
            mutez,
            source,
            gasLimit,
            storageLimit,
            parameter: {
                entrypoint: this.isMultipleEntrypoint ? this.name : DEFAULT_SMART_CONTRACT_METHOD_NAME,
                value: this.isAnonymous
                    ? this.parameterSchema.EncodeObject({ [this.name]: this.args })
                    : this.parameterSchema.EncodeObject(this.args),
            },
        };
        return fullTransferParams;
    }
}
```

packages/taquito/src/contract/prepare.ts
```diff
@@ -339,3 +342,35 @@ export const createSmartRollupOriginateOperation = async ({
...
+    sourceAndDestination,
+    amount,
+    fee = DEFAULT_FEE.TRANSFER,
+    gasLimit = DEFAULT_GAS_LIMIT.TRANSFER,
+    storageLimit = DEFAULT_STORAGE_LIMIT.TRANSFER,
+    mutez = false,
+  }: StakingParams,
+  entrypoint: StakingEntrypoint
+) => {
+  if (!sourceAndDestination) {
```
hui-an-yang, line 357: This validation is not necessary, if user didn't specify source, the logic will fall into signer.publicKeyHash() default is noopSigner which will throw an error

packages/taquito/src/contract/prepare.ts
```diff
@@ -339,3 +342,35 @@ export const createSmartRollupOriginateOperation = async ({
...
+    throw new ParameterValidationError('The parameter sourceAndDestination is missing');
+  }
+  const operation: RPCTransferOperation = {
+    source: sourceAndDestination,
```
hui-an-yang, line 361: createStakingOperation doesn't need source just like all other create...Operation. source will be added in constructOpContents

packages/taquito/src/prepare/prepare-provider.ts
```diff
@@ -1098,4 +1101,75 @@ export class PrepareProvider extends Provider implements PreparationProvider {
...
+
+    const op = await createStakingOperation(
```
hui-an-yang, line 1147: might want to also reference line1025 of how to prepare contractCall to see if it fits pseudo operation's need better than transaction

packages/taquito/src/contract/rpc-contract-provider.ts
AlirezaHaghshenas: Yes. I did this in the next commit. Can you check please?

packages/taquito/src/operations/types.ts
AlirezaHaghshenas: I still believe that exposing `parameter` to a user of `stake` function, which is optional, but if provided should only be `{entrypoint:'stake', value:{prim: 'Unit'}}` is not a good DX.

packages/taquito/src/contract/prepare.ts
AlirezaHaghshenas: Done for `source`.

packages/taquito/src/contract/prepare.ts
AlirezaHaghshenas: Maybe it's a matter of taste, but the common principle is [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it). Short for "You aren't gonna need it", which states that "a programmer should not add functionality until deemed necessary".

packages/taquito/src/contract/prepare.ts
AlirezaHaghshenas: I still prefer to make sure this is validated here as well, not relying on logic in other parts of code that's not represented in types. (`source` is optional in `StakingParams`).
An alternative might be to define a different type with `source` required.

packages/taquito/src/contract/prepare.ts
AlirezaHaghshenas: Let's discuss this in a huddle

hui-an-yang:
oxford in testnet and flextesa ci run are both not working, we'll need at least one of them working with completed unit tests to have another review

integration-tests/contract-staking.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+        });
+        flextesaOxford('Should be able to stake', async (done) => {
+          const op = await Tezos.contract.stake({
+            amount: 100,
+          });
+          await op.confirmation()
+          expect(op.hash).toBeDefined();
+          
+          const constants = await Tezos.rpc.getConstants();
+          await sleep(((constants.preserved_cycles + 2) * constants.blocks_per_cycle * (constants.minimal_block_delay!.toNumber())) * 1000);
```
hui-an-yang, line 20: from my experiment seems to be able to unstake immediately after stake might not need to sleep (the wait between stake and unstake is to get attestationRight/bakingRight)

packages/taquito-core/package.json
```diff
@@ -59,5 +59,11 @@
     "collectCoverageFrom": [
       "src/**/*.{js,ts}"
     ]
+  },
+  "dependencies": {
+    "json-stringify-safe": "^5.0.1"
```
hui-an-yang, line 64: is this dependency necessary?

packages/taquito-core/src/taquito-core.ts
```diff
@@ -3,4 +3,7 @@
  * @module @taquito/core
  */
 
+import stringify from 'json-stringify-safe';
+
 export * from './errors';
+export { stringify };
```
hui-an-yang, line 9: see `stringify` import and export didn't see it used anywhere

packages/taquito-local-forging/test/codec.spec.ts
```diff
@@ -223,15 +223,15 @@ describe('Tests for Entrypoint functions and for encode and decoder error messag
     expect(() =>
       entrypointDecoder(
         Uint8ArrayConsumer.fromHexString(
-          '085c8244b8de7d57795962c1bfc855d0813f8c61eddf3795f804ccdea3e4c82ae95c8244b8de7d57795962c1bfc855d0813f8c61eddf3795f804ccdea3e4c82ae9'
+          '0a5c8244b8de7d57795962c1bfc855d0813f8c61eddf3795f804ccdea3e4c82ae95c8244b8de7d57795962c1bfc855d0813f8c61eddf3795f804ccdea3e4c82ae9'
```
hui-an-yang, line 226: would like to understand the change of starting from 08 to 0a, please

AlirezaHaghshenas: This is related to this code in `packages/taquito-local-forging/src/codec.ts`:

```
export const entrypointDecoder = (value: Uint8ArrayConsumer) => {
  const preamble = pad(value.consume(1)[0], 2);

  if (preamble in entrypointMapping) {
    return entrypointMapping[preamble];
  } else {
    const entry = extractRequiredLen(value, 1);

    const entrypoint = Buffer.from(entry).toString('utf8');

    if (entrypoint.length > ENTRYPOINT_MAX_LENGTH) {
      throw new OversizedEntryPointError(entrypoint);
    }
    return entrypoint;
  }
};
```

This test would pass `08` to `preample` which would go to `else` and read the entrypoint name. Because we had added new values to `packages/taquito-local-forging/src/constants.ts`:`entrypointMapping`, this would no longer work. The test would not throw. I passed `0a`, to re-enable the intended failure. We could set a much larger number, like `ff`, but with `0a`, this test is much more likely to fail in case of new constants, which is a good thing.

packages/taquito-core/src/taquito-core.ts
AlirezaHaghshenas: I created this issue: #2672 for this, and undid the change here.

packages/taquito-core/package.json
AlirezaHaghshenas: See the response to https://github.com/ecadlabs/taquito/pull/2657#discussion_r1338715294

integration-tests/contract-staking.spec.ts
AlirezaHaghshenas: Without the sleeps, I would get some errors in some runs. I can investigate that further.

packages/taquito/src/prepare/prepare-provider.ts
AlirezaHaghshenas: May be a matter of taste, but I prefer the more encapsulated approach

packages/taquito/src/contract/errors.ts
```diff
@@ -32,6 +32,18 @@ export class InvalidDelegationSource extends ParameterValidationError {
...
 
+/**
+ *  @category Error
+ *  @description Error that indicates an invalid staking source contract address being passed or used
+ */
+export class InvalidStakingSource extends ParameterValidationError {
+  constructor(public readonly source: string) {
+    super();
+    this.name = `InvalidStakingSource`;
+    this.message = `The source for staking operations can not be a contract address ${source}.`;
```
hui-an-yang, line 43: contract in tezos sense, can also be implicit might be better to use `originated address` or `smart contract address`?

packages/taquito/test/contract/rpc-contract-provider.spec.ts
```diff
@@ -19,12 +19,15 @@ import {
...
   InvalidDelegationSource,
   InvalidInitParameter,
+  InvalidStakingSource,
 } from '../../src/contract/errors';
 import { preapplyResultFrom } from './helper';
 import { OpKind, ParamsWithKind, TransferTicketParams } from '../../src/operations/types';
 import { NoopParser } from '../../src/taquito';
 import { OperationBatch } from '../../src/batch/rpc-batch-provider';
 import { PvmKind } from '@taquito/rpc';
+import { InvalidAddressError } from '@taquito/utils';
```
hui-an-yang, line 29: Error exporting from @taquito/core best to stick to core. We export it from utils only not introduce breaking change

packages/taquito/src/contract/rpc-contract-provider.ts
```diff
@@ -759,6 +763,87 @@ export class RpcContractProvider extends Provider implements ContractProvider, S
 
     return batch;
   }
+
+  /**
+   *
+   * @description Stake the given amount for the source. The source must be a delegator to be allowed to stake
```
hui-an-yang, line 769: might want to mention `before adaptive_issuance_launch_cycle the source must be a delegator to be allowed to stake, after the head block's cycle reached adaptive_issuance_launch_cycle any implicit can stake`

packages/taquito/src/operations/types.ts
```diff
@@ -563,6 +566,53 @@ export interface FailingNoopParams {
...
+  source: string;
+  amount: number;
+  fee?: number;
+  gasLimit?: number;
+  storageLimit?: number;
+  mutez?: boolean;
+  parameter: TransactionOperationParameter;
+}
+
+type PartialBy<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>;
```
hui-an-yang, line 579: can we add a comment explaining what PartialBy does?

packages/taquito/src/operations/types.ts
```diff
@@ -563,6 +566,53 @@ export interface FailingNoopParams {
...
+  if (entrypoint !== 'finalize_unstake') {
+    if (!params.amount || params.amount <= 0) {
+      throw new InvalidAmountError(
+        params.amount?.toString() ?? '',
+        `The 'amount' field in ${entrypoint} should be a positive number.`
+      );
+    }
+  }
+  if (params.source !== undefined) {
+    const sourceValidation = validateAddress(params.source);
```
hui-an-yang, line 604: validateKeyHash will validate only implicit line608 can be removed

packages/taquito/src/operations/types.ts
```diff
@@ -563,6 +566,53 @@ export interface FailingNoopParams {
...
+ *
+ * @throws {@link InvalidAmountError}
+ * @throws {@link InvalidAddressError}
+ * @throws {@link InvalidStakingSource}
+ */
+export const validateStakingParams = (
+  params: Omit<PartialBy<StakingParams, 'amount'>, 'parameter'>,
+  entrypoint: StakingEntrypoint
+) => {
+  if (entrypoint !== 'finalize_unstake') {
```
hui-an-yang, line 595: iffy about his design, found it reduced readability. will we considered to explicit validate in each staking function at `packages/taquito/src/contract/rpc-contract-provider.ts`

docs/staking.md
```diff
@@ -0,0 +1,42 @@
+---
+title: Staking in Oxford and later protocols
+id: staking
+author: Alireza Haghshenas
+---
+
+Since the first Tezos protocol, Tez owners could participate in baking, endorsement (now renamed to attestation), and voting by `delegating` their tez to bakers.
+Starting with Oxford, there is now an additional way of participating which is called `staking`. You can read more about it (here)[https://tezos.gitlab.io/protocols/018_oxford.html#adaptive-issuance-ongoing] and (here)[https://spotlight.tezos.com/announcing-oxford-tezos-15th-protocol-upgrade-proposal/] (under "Adaptive Issuance and Staking").
+
+In Taquito, we support three new operations: `stake`, `unstake`, and `finalize_unstake`.
```
dsawali, line 10: probably better to specify that they are pseudo operations to not cause confusion

integration-tests/contract-staking.spec.ts
dsawali: best to figure out why the errors happen without the sleep. Relying on sleeps here might mean users would need to introduce sleeps as well(?)

packages/taquito/src/contract/errors.ts
dsawali: I agree with An here, we should explicitly specify which contract it is.

originated contracts: `KT1...`
implicit contracts: `tz1...`, `tz2...`, etc

packages/taquito/src/operations/types.ts
dsawali: I think separating the validator methods is a good idea.

.github/workflows/main.yml
```diff
@@ -74,11 +74,11 @@ jobs:
           - protocol: Nairobi
             testnet: nairobinet
             testnet_uppercase: NAIROBINET
-            flextesa_docker_image: oxheadalpha/flextesa:20230607
+            flextesa_docker_image: oxheadalpha/flextesa:latest
           - protocol: Oxford
             testnet: oxfordnet
             testnet_uppercase: OXFORDNET
-            flextesa_docker_image: oxheadalpha/flextesa:20230915
+            flextesa_docker_image: oxheadalpha/flextesa:latest
```
dsawali, line 81: I'm still on the fence about changing these to point to `latest`

dsawali: and also, nairobinet and oxfordnet cannot both point to `latest`. The Nairobinet docker image is not going to use a nairobibox if we point to latest.

packages/taquito/src/contract/rpc-contract-provider.ts
```diff
@@ -759,6 +763,87 @@ export class RpcContractProvider extends Provider implements ContractProvider, S
...
+    ) as OperationContentsTransaction;
+
+    const opBytes = await this.forge(prepared);
+    const { hash, context, forgedBytes, opResponse } = await this.signAndInject(opBytes);
+    return new TransactionOperation(hash, content, source, forgedBytes, opResponse, context);
+  }
+
+  /**
+   *
+   * @description Unstake the given amount. If "everything" is given as amount, unstakes everything from the staking balance. Unstaked tez remains frozen for a set amount of cycles (the slashing period) after the operation. Once this period is over, the operation "finalize unstake" must be called for the funds to appear in the liquid balance.
```
hui-an-yang, line 795: I understand it's the client's description, might need to take out `If "everything" is given as amount, unstakes everything from the staking balance.`. 
UnstakeParams.amount is type number, for now we don't want to take string 'everything' and query the staked balance and used it as  TransactionOperationParameter.value

packages/taquito/src/contract/interface.ts
hui-an-yang, line 1: description need to be updated as `packages/taquito/src/contract/rpc-contract-provider.ts`'s comment too

packages/taquito/src/operations/types.ts
```diff
@@ -563,6 +566,53 @@ export interface FailingNoopParams {
...
+  fee?: number;
+  gasLimit?: number;
+  storageLimit?: number;
+  mutez?: boolean;
+  parameter: TransactionOperationParameter;
+}
+
+type PartialBy<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>;
+export type StakingParams = PartialBy<InternalStakingParams, 'source'>;
+export type StakeParams = Omit<StakingParams, 'parameter'>;
```
hui-an-yang, line 581: preferably having each params defined by themselves explicitly for readability and better user experience

packages/taquito/src/operations/types.ts
```diff
@@ -563,6 +566,53 @@ export interface FailingNoopParams {
   basedOnBlock: BlockIdentifier;
 }
 
+export interface InternalStakingParams {
```
hui-an-yang, line 569: what's the reason of creating ` InternalStakingParams` with source required?

packages/taquito/src/prepare/prepare-provider.ts
```diff
@@ -1092,4 +1098,96 @@ export class PrepareProvider extends Provider implements PreparationProvider {
...
+    ...rest
+  }: StakingParams): Promise<PreparedOperation> {
+    const { pkh } = await this.getKeys();
+
+    const protocolConstants = await this.context.readProvider.getProtocolConstants('head');
+    const DEFAULT_PARAMS = await this.getAccountLimits(pkh, protocolConstants);
+    const mergedEstimates = mergeLimits({ fee, storageLimit, gasLimit }, DEFAULT_PARAMS);
+
+    const op = await createStakingOperation({
+      source: source ?? pkh,
```
hui-an-yang, line 1168: it's okey to not have source at this point yet

packages/taquito/src/prepare/prepare-provider.ts
```diff
@@ -1092,4 +1098,96 @@ export class PrepareProvider extends Provider implements PreparationProvider {
...
+    const operation = await this.addRevealOperationIfNeeded(op, pkh);
+    const ops = this.convertIntoArray(operation);
+
+    const hash = await this.getBlockHash();
+    const protocol = await this.getProtocolHash();
+
+    this.#counters = {};
+    const headCounter = parseInt(await this.getHeadCounter(pkh), 10);
+
+    const contents = this.constructOpContents(ops, headCounter, pkh, source ?? pkh);
```
hui-an-yang, line 1182: feels unnecessary to have ` ?? pkh`

packages/taquito/src/operations/types.ts
AlirezaHaghshenas: If someone knows `Partial` helper type in typescript, the name should be self explanatory. I added a comment anyway.

.github/workflows/main.yml
AlirezaHaghshenas: I think even the latest, when called with `NairobiBox` param will run NairobiNet

packages/taquito/src/operations/types.ts
AlirezaHaghshenas: I think that causes worse maintainability. Also it is WET (Write Everything Twice) as opposed to DRY (Don't Repeat Yourself)

packages/taquito/src/operations/types.ts
AlirezaHaghshenas: It is used in `createStakingOperation` to make sure the calling code has added source to the `params` provided by the Dapp developer

packages/taquito/src/operations/types.ts
AlirezaHaghshenas: @dsawali suggestions?

packages/taquito/src/contract/prepare.ts
AlirezaHaghshenas: c.c. @dsawali

packages/taquito/src/prepare/prepare-provider.ts
AlirezaHaghshenas: @dsawali ?

packages/taquito/src/contract/prepare.ts
AlirezaHaghshenas: @dsawali ?

packages/taquito/src/operations/types.ts
AlirezaHaghshenas: I did that, but then realized that I prefer to have different errors in case the address is in wrong format, vs when the address has an unsupported prefix.
Undoing the change I made in response to this comment now.

docs/staking.md
```diff
@@ -0,0 +1,42 @@
+---
+title: Staking in Oxford and later protocols
```
hui-an-yang, line 2: is `in Oxford and later protocols` necessary?

docs/staking.md
```diff
@@ -0,0 +1,42 @@
...
+
+Since the first Tezos protocol, Tez owners could participate in baking, endorsement (now renamed to attestation), and voting by `delegating` their tez to bakers.
+Starting with Oxford, there is now an additional way of participating which is called `staking`. You can read more about it (here)[https://tezos.gitlab.io/protocols/018_oxford.html#adaptive-issuance-ongoing] and (here)[https://spotlight.tezos.com/announcing-oxford-tezos-15th-protocol-upgrade-proposal/] (under "Adaptive Issuance and Staking").
+
+In Taquito, we support three new operations: `stake`, `unstake`, and `finalize_unstake` implemented as pseudo operations (entrypoints on implicit accounts).
+
+## stake
+By `stake`ing your tez to a baker that accepts staking, your tez will be frozen. Also, unlike delegation, your tez is subject to slashing in case the baker misbehaves. In return, you can receive more rewards for baking and attestation (formerly endorsement).
+
+Assuming a backer with address `baker0` is set to accept staking (and after the 3 cycles on testnet or 5 cycles on mainnet are passed), you can stake funds like this:
```
hui-an-yang, line 15: typo backer

docs/staking.md
```diff
@@ -0,0 +1,42 @@
...
+id: staking
+author: Alireza Haghshenas
+---
+
+Since the first Tezos protocol, Tez owners could participate in baking, endorsement (now renamed to attestation), and voting by `delegating` their tez to bakers.
+Starting with Oxford, there is now an additional way of participating which is called `staking`. You can read more about it (here)[https://tezos.gitlab.io/protocols/018_oxford.html#adaptive-issuance-ongoing] and (here)[https://spotlight.tezos.com/announcing-oxford-tezos-15th-protocol-upgrade-proposal/] (under "Adaptive Issuance and Staking").
+
+In Taquito, we support three new operations: `stake`, `unstake`, and `finalize_unstake` implemented as pseudo operations (entrypoints on implicit accounts).
+
+## stake
```
hui-an-yang, line 12: might want to mention that previously delegate set_deposit_limit is replaced by stake now

hui-an-yang: might want to explain the difference before vote off is only delegate can stake and vote on is delegator (implicit addres) can stake too

docs/staking.md
```diff
@@ -0,0 +1,42 @@
...
+
+```javascript
+const Tezos = new TezosToolkit('https://ghostnet.ecadinfra.com');
+const op = await Tezos.contract.stake({
+    amount: 1000,
+});
+await op.confirmation();
+```
+
+Now if you delegate to baker0, the amount you have staked will staked instead of the normal delegation. The 1000 tez you staked will be frozen and can be slashed if the baker misbehaves. You can also unstake your funds by calling `unstake`:
```
hui-an-yang, line 25: > Now if you delegate to baker0, the amount you have staked will staked instead of the normal delegation.
Don't quite understand this line, can you rephrase in simpler words?

packages/taquito/test/estimate/rpc-estimate-provider.spec.ts
```diff
@@ -1588,4 +1588,101 @@ describe('RPCEstimateProvider test wallet', () => {
...
+          {
+            kind: 'transaction',
+            fee: 10018,
+            metadata: {
+              operation_result: { status: 'applied', consumed_milligas: '10014000' },
+            },
+            parameters: {
+              entrypoint: 'unstake',
+              value: {
+                prim: 'Unit',
```
hui-an-yang, line 1638: thought the unstake will have `value: { int: '10000000000'}`

hui-an-yang: actually doesn't even need parameters at all for all three op, will be nice to have to if having might want to be correct

packages/taquito/test/contract/rpc-contract-provider.spec.ts
```diff
@@ -1627,4 +1636,233 @@ describe('RpcContractProvider test', () => {
...
+      mockEstimate.stake.mockResolvedValue(estimate);
+      expect(async () => {
+        await rpcContractProvider.stake({
+          amount: -100,
+        });
+      }).rejects.toThrowError(InvalidAmountError);
+      done();
+    });
+
+    it('should throw InvalidAmountError when passed zero for amount', async (done) => {
```
hui-an-yang, line 1689: think this case will contradict with `packages/taquito/src/prepare/prepare-provider.ts:235`, if undefined will be default 0 (seem fine to return rpc error `"proto.018-Proxford.operations.invalid_nonzero_transaction_amount"
` to user )

packages/taquito/test/contract/rpc-contract-provider.spec.ts
```diff
@@ -1627,4 +1636,233 @@ describe('RpcContractProvider test', () => {
...
+      const estimate = new Estimate(1240001, 95, 146, 248);
+      mockEstimate.unstake.mockResolvedValue(estimate);
+      expect(async () => {
+        await rpcContractProvider.unstake({
+          amount: 0,
```
hui-an-yang, line 1779: same argument as stake

packages/taquito/src/prepare/prepare-provider.ts
hui-an-yang: they put source and pkh separately so contracutOpContents can have source to fall on. And inside contracutOpContents will now have circular logic

packages/taquito/src/prepare/prepare-provider.ts
```diff
@@ -1092,4 +1098,96 @@ export class PrepareProvider extends Provider implements PreparationProvider {
...
+   * @returns a PreparedOperation object
+   */
+  async finalizeUnstake(params: FinalizeUnstakeParams): Promise<PreparedOperation> {
+    return this.staking({
+      parameter: {
+        entrypoint: 'finalize_unstake',
+        value: { prim: 'Unit' },
+      },
+      ...params,
+      amount: 0,
```
hui-an-yang, line 1150: FinalizeUnstakeParams doesn't allow specify amount will go to `packages/taquito/src/prepare/prepare-provider.ts:235` to be '0'

packages/taquito/test/estimate/rpc-estimate-provider.spec.ts
dsawali: > thought the unstake will have value: { int: '10000000000'}

did we ever get to confirm with the protocol team whether this value is arbitrary, fixed, or if it has any significance?

packages/taquito/src/operations/types.ts
hui-an-yang: in general all transaction op can't have kt1 as source, `packages/taquito/src/prepare/prepare-provider.ts:239` have addressed that not necessary for extra InvalidStakingSource

packages/taquito/src/operations/types.ts
```diff
@@ -563,6 +566,69 @@ export interface FailingNoopParams {
...
+ *
+ * @throws {@link InvalidAmountError}
+ * @throws {@link InvalidAddressError}
+ * @throws {@link InvalidStakingSource}
+ */
+export const validateStakeOrUnstakeParams = (
+  params: StakeParams | UnstakeParams,
+  entrypoint: StakingEntrypoint
+) => {
+  if (!params.amount || params.amount <= 0) {
```
hui-an-yang, line 596: with previous comment that <0 is sufficient

packages/taquito/src/operations/types.ts
```diff
@@ -563,6 +566,69 @@ export interface FailingNoopParams {
...
+      `The 'amount' field in ${entrypoint} should be a positive number.`
+    );
+  }
+  if (params.source !== undefined) {
+    const sourceValidation = validateAddress(params.source);
+    if (sourceValidation !== ValidationResult.VALID) {
+      throw new InvalidAddressError(params.source, invalidDetail(sourceValidation));
+    }
+    if (/kt/i.test(params.source)) {
+      throw new InvalidStakingSource(params.source);
```
hui-an-yang, line 608: same as the other comment, source can't be kt1

packages/taquito/src/contract/rpc-contract-provider.ts
hui-an-yang: bump

packages/taquito/src/contract/interface.ts
```diff
@@ -283,4 +286,34 @@ export interface ContractProvider extends StorageProvider {
    * @param params FailingNoop operation parameter
    */
   failingNoop(params: FailingNoopParams): Promise<FailingNoopOperation>;
+
+  /**
+   *
+   * @description Stake the given amount for the source. The source must be a delegator to be allowed to stake. Will sign and inject an operation using the current context
```
hui-an-yang, line 292: desc can be update too. (diff between vote on/off )

packages/taquito/src/contract/errors.ts
hui-an-yang: bump

docs/staking.md
```diff
@@ -0,0 +1,42 @@
...
+```javascript
+const Tezos = new TezosToolkit('https://ghostnet.ecadinfra.com');
+const op = await Tezos.contract.unstake({
+    amount: 1000,
+});
+await op.confirmation();
+```
+
+Which moves the 1000tez from `frozen` to `unstaked` balance. You can pass a larger amount to `unstake` than the amount you have staked, and that will unstake all your funds.
+After 5 cycles on testnet or 7 cycles on mainnet, the stake moves to `finalizable` balance, and you can use `finalizeUnstake` to move the funds to your `free` balance:
```
hui-an-yang, line 36: can we unify our wording to be `liquid balance`

hui-an-yang: don't mind it to all be updated `liquid balance` (from nomadic docs) or `spendable` (from developer doc)

integration-tests/contract-staking.spec.ts
hui-an-yang: is this finalized?

integration-tests/contract-staking.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+            await op.confirmation()
+          }).rejects.toThrowError(/Invalid unstake request amount/);
+          done();
+        });
+        flextesaOxford('Should fail with invalid_nonzero_transaction_amount', async (done) => {
+          expect(async () => {
+            const address = await Tezos.signer.publicKeyHash();
+            const op = await Tezos.contract.transfer({
+              to: address,
+              amount: 10000,
```
hui-an-yang, line 56: if you use contract transfer to construct an unstake op, amount should be 0

integration-tests/contract-staking.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+              parameter: {
+                entrypoint: "unstake",
+                value: {
+                  int: "10000",
+                },
+              },
+            });
+            await op.confirmation()
+          }).rejects.toThrowError(/\(permanent\) proto.\d{3}-\w+\.operations\.invalid_nonzero_transaction_amount/);
```
hui-an-yang, line 65: "invalid_unstake_request_amount"?

packages/taquito/test/estimate/rpc-estimate-provider.spec.ts
AlirezaHaghshenas: From the types in Taquito is seems to be required when `entrypoint` is set

packages/taquito/src/operations/types.ts
AlirezaHaghshenas: No problem, I can remove this check.

integration-tests/contract-staking.spec.ts
AlirezaHaghshenas: Do you suggest that we catch this error and throw the one you suggest?

integration-tests/contract-staking.spec.ts
AlirezaHaghshenas: This is the intention. We are testing that we can see this error.

PR #2689, opened 2023-10-04
docs: add documentation about flattening of pairs and unions

### LINKED ISSUE #2328: Unable to pass complex parameter to contract calls that include unions
Parameters for contract calls that include a union are pretty complicated to format properly with Taquito.

In the case of an entrypoint that accepts a simple union value of type `or nat string`, Taquito expects this parameter:
```
// to pass `Left 23`
let op = await contract.methods.entrypoint(0, 23).send();
// to pass `Right "tezos"`
let op = await contract.methods.entrypoint(1, "tezos").send();
```
The formatting is not intuitive and it took me a few tries with different combinations to guess it.

The formatting becomes unclear when a union has to be passed as a member of a complex type. I managed to pass a list of union to a contract as long as the unions only have a `Left ...` value, for example:
```
let op = await contract.methods.entrypoint([
    [0, 23],
    [0, 45],
    [0, 67]
]).send();
```
but it seems impossible to pass a value like `Right "tezos"` within the list, the error message implied that the forger was looking for a value `0` and passing `[1, "tezos"]` didn't work. I haven't tried to pass a union in a pair, but I imagine I would face the same issue.

Unions in a list are not an efficient pattern (they require an `IF_LEFT` instruction that would be called on each element of the list and use additional gas), but high-level languages like Ligo use them to represent `variants` and they may end up in a list.

The current formatting of unions in JS by Taquito doesn't seem optimal, here is what I would recommend:
- An array of 2 elements like `["Left", 23]` or `["Right", "tezos"]` that looks more like its Michelson counterpart and would be easier to integrate as a nested member of a pair, a list, etc.
- An object like `{ "left": 23 }` or `{ "right": "tezos" }` that has the same qualities as the one above, may be easier for the forger to parse and for TypeScript to type.

### LINKED ISSUE #2458: formatting of unions in JS by Taquito isn't optimal
The current formatting of unions in JS by Taquito doesn't seem optimal, here is what I would recommend:

An array of 2 elements like ["Left", 23] or ["Right", "tezos"] that looks more like its Michelson counterpart and would be easier to integrate as a nested member of a pair, a list, etc.
An object like { "left": 23 } or { "right": "tezos" } that has the same qualities as the one above, may be easier for the forger to parse and for TypeScript to type.

see #2328

AlirezaHaghshenas:
Added some clarification about flattening and field naming (or numbering) in nested `pair`s and `union`s.
closes #2458 
closes #2328 

- [x] You have added or updated corresponding documentation


hui-an-yang (commented):
it's an add-on to the current version can you update `website/versioned_docs/version-17.3.0/michelson_encoder.md` too

AlirezaHaghshenas:
> it's an add-on to the current version can you update `website/versioned_docs/version-17.3.0/michelson_encoder.md` too

This is done

hui-an-yang (approved):

PR #2736, opened 2023-11-23
Changed axios and axios fetch adapter usage into node-fetch 

### LINKED ISSUE #2461: Replace Axios with Fetch in Taquito
Taquito currently relies on `Axios` HTTP request implementation to make calls to any external endpoints. While `Axios` was quite easy to work with, it had its limitations when it comes to running in worker environments or some browser environments. 

In previous version of nodes, fetch wasn't available.
The `axios-fetch-adapter` package was introduced to alleviate such limitations. `Axios` provides an `adapter` parameter property when creating HTTP requests to make sure the requests are made using the custom adapter when certain conditions are met. 

While the `axios-fetch-adapter` was a worthwhile solution at the time, it had introduced new issues due to the package not being maintained actively anymore.

The issues it introduced are as follows:
- Version conflicts between the `axios` version imported in `axios-fetch-adapter` and Taquito, and sometimes even for our users when they also use Axios in their package
- incompatibility with newer ESM packages due to it being written in old JS

### Solution Proposal
One proposed solution we discussed is to use `fetch` and remove `Axios` completely from our dependencies. 

This will involve adapting our current HTTP utils package to account for the different behaviours of `Axios` vs `fetch`
Some differences to note:
- `fetch` is very lean, so there aren't any out of the box features that comes with `Axios` (e.g. auto JSON data transformation, response timeouts)
- Some potential browser incompatibility that might be introduced by `fetch`. `Axios` is built using `XMLHttpRequest`, making it very browser friendly, even with very old browsers
- In `Axios`, anything HTTP code 400 and above are considered `errors`. Fetch just sees them all as responses and will not reject any promises, so we will need to handle them in Taquito accordingly since we have been relying on `Axios` response errors in some of our code

The benefits of using `fetch` as opposed to `Axios` are as follows:
- `fetch` is a built in feature in Node, so no package imports are necessary
- Since it's a built in Node feature, there no chance it will conflict with other packages
- It gives us more flexibility to what we want to implement as opposed to something that's abstracted like Axios
- Removes the dependency on `Axios` when something doesn't work properly

### Acceptance Criteria
- Replace the usage of `Axios` in Taquito in favour of `fetch`
- ensure backwards compatibility with our existing integration/unit tests as to not introduce new issues to our users

dsawali:
closes #2461 

- Removed axios implementation in taquito-http-utils
- Added node-fetch to match Axios' implementation in Taquito
- Wrapped fetch in the new `createRequest` method 
- Added better error handling 
- Updated some integration tests to match new implementation
- Updated axios reference in the sapling package to use fetch 

- [x] All public methods or types have TypeDoc coverage with a complete description, and ideally an @example

- [x] You have added a concise description on your changes
## Release Note Draft Snippet


packages/taquito-http-utils/src/taquito-http-utils.ts
```diff
@@ -74,60 +67,63 @@ export class HttpBackend {
   /**
    *
    * @param options contains options to be passed for the HTTP request (url, method and timeout)
-   * @throws {@link HttpRequestFailed} | {@link HttpResponseError}
+   * @throws {@link HttpRequestFailed} | {@link HttpResponseError} | {@link HttpTimeoutError}
    */
   async createRequest<T>(
-    { url, method, timeout = this.timeout, query, headers = {}, json = true }: HttpRequestOptions,
+    { url, method, query, timeout = this.timeout, headers = {}, json = true }: HttpRequestOptions,
```
hui-an-yang, line 73: createRequest is an exported function the order of params changes might create breaking change for user. Worth thinking about what value does the order change brings and does it out weight creating breaking change?

docs/cancel_http_requests.md
```diff
@@ -32,25 +40,36 @@ class CancellableHttpBackend extends HttpBackend {
...
+
+      if (json) {
+        return response.json() as T;
+      } else {
+        return response.text() as unknown as T;
+      }
+    } catch (e: unknown) {
+      ...
+    } finally {
+      clearTimeout(t);
```
hui-an-yang, line 72: unsure i followed this example from line 52-72, is this a fetch way to do thing?

packages/taquito-http-utils/src/taquito-http-utils.ts
```diff
@@ -74,60 +67,63 @@ export class HttpBackend {
...
     }
 
-    if (!json) {
-      resType = ResponseType.TEXT;
-      transformResponse = [<Type>(v: Type) => v];
-    } else {
-      resType = ResponseType.JSON;
-    }
+    // Creates a new AbortController instance to handle timeouts
+    const controller = new AbortController();
```
hui-an-yang, line 85: please correct if i am mistaken just don't know where is AbortController is defined or imported to this file

hui-an-yang: read the node doc realizing it's native in node, will resolve this convo

docs/cancel_http_requests.md
dsawali: yes, fetch does not timeout by default. If you want to timeout your fetch requests, you'll need to send an abort signal with a manual timeout method. Line 72 just cleans the timer out of the scope.

packages/taquito-http-utils/src/taquito-http-utils.ts
dsawali: AbortController is built in from node

packages/taquito-http-utils/src/taquito-http-utils.ts
dsawali: I'm pretty sure this destructure does not hold sequence, hence no breaking change. But I can revert the sequence back for aesthetics

hui-an-yang: that's great realized it's an object the order doesn't matter will resolve

docs/cancel_http_requests.md
dsawali: note that this is a doc to customize our existing `createRequest` method.

hui-an-yang (approved):
lgtm

packages/taquito-http-utils/src/taquito-http-utils.ts
```diff
@@ -3,23 +3,16 @@
...
  */
 
-import fetchAdapter from './fetch-adapter';
-import { STATUS_CODE } from './status_code';
-import axios from 'axios';
-import { HttpRequestFailed, HttpResponseError } from './errors';
+// const fetch = require('node-fetch');
 
-const isNode = typeof process !== 'undefined' && !!process?.versions?.node;
+import fetch from 'node-fetch';
```
AlirezaHaghshenas, line 8: Wondering, in here, is this line (and the similar one in `packages/taquito-sapling/fetch-sapling-params.js`), specific to `node`? What about browser?

If we change this line so that it only imports if `fetch` is not already available?

From [this answer](https://stackoverflow.com/a/72440263):

```
let fetch = globalThis?.fetch;

if (!fetch && process?.versions?.node) {
  fetch = (await import('node-fetch')).default;
}

console.log(fetch);
```

But I guess something more relevant to us can conceptually be:
```
import fetch as node_fetch from 'node-fetch';
let fetch = globalThis?.fetch;
if (!fetch && process?.versions?.node) {
  fetch = node_fetch;
}
```

But I don't have enough context.

dsawali: the reason for that existing before is to use the `axios-fetch-adapter` when the runtime is a browser. 

But since we've now updated to fetch completely, there would be no need to check what runtime Taquito is currently running on.

AlirezaHaghshenas: I mean, `fetch` is natively available on browsers, and `node-fetch` is there to make it available for `node`. So in browser we don't need to import `fetch` from `node-fetch`?

dsawali: oh I see what you mean. That makes sense, I hadn't consider that. I'll update it

PR #2781, opened 2024-01-02
2776 smart rollup origination with whitelist

### LINKED ISSUE #2776: [Oxford2] Update local forger to include 'whitelist' property for smart rollup originate
Oxford2 introduced changes to Smart Rollups where you can now make use of private rollups.

The whitelist property is introduced as an optional property now in `smart_rollup_originate`

Investigate and implement the changes in local forger to reflect the changes.

> The smart_rollup_originate operation now also takes an optional whitelist of public key hashes. This whitelist cannot be used yet (the sc_rollup.private_enable flag has to be set to true). (MR [!9401](https://gitlab.com/tezos/tezos/-/merge_requests/9401))

Relevant docs:
- https://tezos.gitlab.io/protocols/018_oxford.html#smart-rollups

hui-an-yang:
supporting whitelist property and new PVMkind riscv of smartRollupOrigination in localForging 
closes #2776 

- [x] You have added a concise description on your changes
## Release Note Draft Snippet


packages/taquito-local-forging/src/decoder.ts
```diff
@@ -63,6 +64,7 @@ export const decoders: { [key: string]: Decoder } = {
   [CODEC.ZARITH]: zarithDecoder,
   [CODEC.PUBLIC_KEY]: publicKeyDecoder,
   [CODEC.PKH]: pkhDecoder,
+  [CODEC.PKH_ARR]: pkhsDecoder,
```
dsawali, line 67: would you consider updating this to PKH_ARRAY or PKH_LIST instead for clarity?

packages/taquito-local-forging/src/codec.ts
```diff
@@ -57,6 +57,18 @@ export const pkhDecoder = (val: Uint8ArrayConsumer) => {
   }
 };
 
+export const pkhsDecoder = (val: Uint8ArrayConsumer) => {
```
dsawali, line 60: what would the `s` be in this case in `pkhs`

packages/taquito-local-forging/src/decoder.ts
hui-an-yang: At the time, trying to be consistent with existing `PROPOSAL_ARR` but I like your suggestions better

packages/taquito-local-forging/src/codec.ts
hui-an-yang: same as the other comment inherit the naming consistency from `proposalsDecoder`, I think `s` means plural

dsawali: I think for proposals it's because the operation itself is called `Proposals` instead of `proposal`.

I think we can remove the `S` in this case, but that's just personal preference.

hui-an-yang: there are `proposalDecoder` and `proposalsDecoder`. There is already `pkhDecoder` will we consider `pkhArrayDecoder`?

packages/taquito-local-forging/src/decoder.ts
hui-an-yang: if the decoder going with pkhArrayDecoder will update this as PKH_ARRAY.

packages/taquito-local-forging/src/codec.ts
dsawali: yeah let's do pkhArrayDecoder then, seems more informative than just using `s`.

dsawali (approved):
good work

PR #2848, opened 2024-02-29
Updating fixed reveal op fee with buffer

hui-an-yang:
Realizing fixed reveal operation fee reduced to base works for testnets might not work for mainnet when there is higher valume, decided to add a 20% buffer just in case

## Release Note Draft Snippet


dsawali:
20% is quite a significant bump, do we have a reason for that particular number?

hui-an-yang:
The fee used to work consistently is 375 and for the highest tz3 address base fee is 305 multiply by 1.2 is 366. 1.3 times will be paying more than before optimization. What do you think about reduce it to 1.1 or used a fixed buffer like 10 or 20 added to base fee instead? Welcome other suggestions too

dsawali:
I haven't got the data to interpolate what values might be best, but I guess there isn't much harm in keeping it to 20% for now with the option of reducing it in the future if needed.

dsawali (approved):

PR #2866, opened 2024-03-12
2858 ux improvements to search

### LINKED ISSUE #2858: Improvements to search functionality in the website
Currently, I observe these search related problems. These can affect a user trying to find information.
The information I'm trying to find is: `RemoteSigner`.

1. When going to the website, in home page, the search functionality returns nothing
2. In docs, in any version other than the currently latest version, the search functionality returns nothing
3. When in the current version, the most relevant result is not shown in the first page, and then when clicking on the "See all 11 results":
4. We are taken to a page with Dropbox for version, where the item is on "next", so the search returns nothing.

As a user, I could not guess changing the version fixes this.

Also, searching in Google for `RemoteSigner site:tezostaquito.io` does not return the "Class RemoteSigner" in the first page of results

ac10n:
fixes #2858 

## Release Note Draft Snippet

### DIFF
--- a/website/src/theme/SearchPage/index.js
+++ b/website/src/theme/SearchPage/index.js
@@ -1,0 +1,457 @@
+/* eslint-disable jsx-a11y/no-autofocus */
+import React, {useEffect, useReducer, useRef, useState} from 'react';
+import clsx from 'clsx';
+import algoliaSearchHelper from 'algoliasearch-helper';
+import algoliaSearch from 'algoliasearch/lite';
+import ExecutionEnvironment from '@docusaurus/ExecutionEnvironment';
+import Head from '@docusaurus/Head';
+import Link from '@docusaurus/Link';
+import {useAllDocsData} from '@docusaurus/plugin-content-docs/client';
+import {
+  HtmlClassNameProvider,
+  useEvent,
+  usePluralForm,
+  useSearchQueryString,
+} from '@docusaurus/theme-common';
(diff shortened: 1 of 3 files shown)


website/src/theme/SearchPage/index.js
```diff
@@ -0,0 +1,459 @@
+/* eslint-disable jsx-a11y/no-autofocus */
+import React, {useEffect, useReducer, useRef, useState} from 'react';
```
dsawali, line 2: could you remind me why we needed to add an index.js file here? I think you mentioned it during standup but I can't remember what it actually does

ac10n: In the [issue](https://github.com/ecadlabs/taquito/issues/2858) I have [commented](https://github.com/ecadlabs/taquito/issues/2858#issuecomment-1992270479): 

> It seems to be a [known issue](https://github.com/facebook/docusaurus/issues/5084) that the search page does not respect the DocSearch options, so while turning the contextualSearch off changes the behaviour of the search modal, it does not affect the search page.
> 
> I swizzled the search page. In a separate commit, I modified it to respect the option. After merging, and verifying that the search works properly, it might be fixed for the searchpage as well, because a re-index with `contextualSearch:off` will fix that, according to the linked issue.
> 
> In that case, we can do a separate PR to undo the swizzling of the SearchPage, which is a good idea, as it reduces the code we need to maintain.

dsawali (approved):

PR #2894, opened 2024-04-04
refactor(ci): migrate test dapp deploy to cf pages using wrangler

GImbrailo:


hui-an-yang (approved):
Great work! Just a nitpick would be nice to update file name `.github/workflows/deploy_website.yml` to be test-dapp relevant

PR #2950, opened 2024-05-03
Staking wallet api refine

hui-an-yang:
@taquito/taquito updated walletAPI typedoc 
integration-tests updated assertion to be robusst

## Release Note Draft Snippet


packages/taquito/src/wallet/wallet.ts
```diff
@@ -409,10 +372,8 @@ export class Wallet {
   }
 
   /**
-   *
    * @description Transfer all the finalizable unstaked funds of the source to their liquid balance
-   * @returns An operation handle with the result from the rpc node
-   *
+   * @returns A wallet command from which we can send the operation to the wallet
```
dsawali, line 376: I would keep the previous description. This is technically correct, but doesn't tell our users much because a wallet command is just a helper method.

dsawali (changes requested):

packages/taquito/src/wallet/wallet.ts
hui-an-yang: for contractAPI it will return and make sense to have`An operation handle with the result from the rpc node` but walletAPI does return a send command only, it's after the send command is called that will return a WalletOperaion, i think something like `@returns A wallet command "send" to send the operation to the wallet that will return the WalletOperation` might be more appropriate.  Do you still think it's best to keep ` An operation handle with the result from the rpc node` and update all operations supported in walletApi accordinly

dsawali: `@returns a wallet command...` is not correct. The returns will bubble up. A `stake` method returns a `Promise<TransactionWalletOperation>` object. The description should not have to mention the 'wallet command' part of the execution. That should just stay internal. 

The previous description is correct, but not explicit. If you want to make it a bit more clear. It should say something like `@returns TransactionWalletOperation promise object, which contains the operation hash and the block response observable`.

dsawali (approved):

dsawali (approved):

PR #2960, opened 2024-05-13
2958 eventsub

### LINKED ISSUE #2958: Update remaining Taquito references of getActiveAccount()
Since Beacon v4.2.0, `getActiveAccount()` needs to be replaced by the event subscription as stated in Beacon's migration guide:
https://docs.walletbeacon.io/guides/migration-guide/

In a previous conversation, we have determined together in a video call with Beacon some time ago that Taquito's internal usage of `getActiveAccount()` to not be intrusive since it's only been used for 'validation' checks when trying to send operations to the wallet. For example: https://github.com/ecadlabs/taquito/blob/master/packages/taquito-beacon-wallet/src/taquito-beacon-wallet.ts#L62

However, for the sake of consistency and reliability, we should update Taquito's Beacon wallet implementation to match the recommended usage from Beacon. That is to use the `ACTIVE_ACCOUNT_SET` event using the `eventSubscribe` method instead of `getActiveAccount()`

dsawali:
closes #2958 
Updated taquito beacon wallet to use `subscribeToEvent` instead of `getActiveAccount`

## Release Note Draft Snippet

### DIFF
--- a/packages/taquito-beacon-wallet/src/taquito-beacon-wallet.ts
+++ b/packages/taquito-beacon-wallet/src/taquito-beacon-wallet.ts
@@ -10,6 +10,8 @@ import {
   PermissionScope,
   getDAppClientInstance,
   SigningType,
+  AccountInfo,
+  BeaconEvent,
 } from '@airgap/beacon-dapp';
 import { BeaconWalletNotInitialized, MissingRequiredScopes } from './errors';
 import toBuffer from 'typedarray-to-buffer';
@@ -32,9 +34,15 @@ export { BeaconWalletNotInitialized, MissingRequiredScopes } from './errors';
 
 export class BeaconWallet implements WalletProvider {
   public client: DAppClient;
+  public account: AccountInfo | undefined;
 
   constructor(options: DAppClientOptions) {
     this.client = getDAppClientInstance(options);
+
+    this.client.subscribeToEvent(BeaconEvent.ACTIVE_ACCOUNT_SET, async (data) => {
+      console.log(`${BeaconEvent.ACTIVE_ACCOUNT_SET} triggered: `, data);
+      this.account = data;
+    });
   }
 
   private validateRequiredScopesOrFail(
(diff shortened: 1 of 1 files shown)


packages/taquito-beacon-wallet/src/taquito-beacon-wallet.ts
```diff
@@ -59,19 +66,17 @@ export class BeaconWallet implements WalletProvider {
   }
 
   async getPKH() {
-    const account = await this.client.getActiveAccount();
-    if (!account) {
+    if (!this.account) {
```
ac10n, line 69: One concern:
I the dApp code calls getPkh immediately after connecting, but before the first `ACTIVE_ACCOUNT_SET` event is received, they will get an error. Not sure if this is actually possible, or even if it's possible if there is a better solution.

dsawali: I don't think that will happen since the even will be subscribed to in the constructor. So getPkh will likely never be called unless the client has been constructed, and in consequence, the `account` member variable will always be assigned a value. And if it doesn't it will throw an error, so at least it signals the user (or developer) that the wallet has not been connected, or the wallet info is incomplete. I hope that makes sense!

ac10n: While the event is subscribed in the constructor, it might not be triggered until a split-second later

dsawali: hmm, I figured it would be a rare occurrence. But do you have any suggestions to improve it then?

ac10n (approved):
It might be a good idea to investigate if it's possible to create the error mentioned in comments. If so, we might want to add something to release notes.

hui-an-yang (commented):
Code change in `packages/taquito-beacon-wallet/src/taquito-beacon-wallet.ts` looks good!! Might be pointing to master makes this pr regressed some commit changes already in master rn. Other than that just would be good to have test-dapp to switch account to prove the change to subscribeToEvent will capture switching account successfully in beaconWallet class.

PR #2986, opened 2024-06-11
added keepalive=false in createRequest

dsawali:
closes #2973 

- updated taquito-http-util fetch call to include `keepalive=false`. Node 19 onwards now defaults keepalive to `true`, which were causing the socket hangup errors

Thank you for your contribution to Taquito.

Before submitting this PR, please make sure:

- [ ] Your code builds cleanly without any errors or warnings
- [ ] You have run the linter against the changes
- [ ] You have added unit tests (if relevant/appropriate)
- [ ] You have added integration tests (if relevant/appropriate)
- [ ] All public methods or types have TypeDoc coverage with a complete description, and ideally an @example
- [ ] You have added or updated corresponding documentation
- [ ] If relevant, you have written a first draft summary describing the change for inclusion in Release Notes. 

In this PR, please also make sure: 

- [ ] You have linked this PR to the issue by putting `closes #TICKETNUMBER` in the description box (when applicable)
- [ ] You have added a concise description on your changes
## Release Note Draft Snippet

__If relevant, please write a summary of your change that will be suitable for
inclusion in the Release Notes for the next Taquito release.__

Issue #2973, opened 2024-05-27
Unable to run integration-test against private rpc node with node v20 


hui-an-yang:
When I tried to run integration-tests with node lts/iron -> v20.13.1 will get errors like below, lts/hydrogen -> v18.20.3 were fine.
Weirdly when i switch to public rpc url with the same node v20 were fine, might be a http and https differences.
I have back track to node v19.0.0 is when this behaviour started, will need to look into node release notes to find what need to be updated on taquito side. 

```
    Failed to originate ticket contract {"method":"GET","url":"http://ecad-oxfordnet-full.i.tez.ie:8732/chains/main/chain_id","cause":{"message":"request to http://ecad-oxfordnet-full.i.tez.ie:8732/chains/main/chain_id failed, reason: socket hang up","type":"system","errno":"ECONNRESET","code":"ECONNRESET"},"name":"HttpRequestFailed","message":"GET http://ecad-oxfordnet-full.i.tez.ie:8732/chains/main/chain_id FetchError: request to http://ecad-oxfordnet-full.i.tez.ie:8732/chains/main/chain_id failed, reason: socket hang up"}
```
and 
```
    HttpRequestFailed: GET http://ecad-ghostnet-rolling-2.i.tez.ie:8732/chains/main/chain_id FetchError: request to http://ecad-ghostnet-rolling-2.i.tez.ie:8732/chains/main/chain_id failed, reason: socket hang up
        at HttpBackend.<anonymous> (/Users/huianyang/ECAD_LABS/taquito/packages/taquito-http-utils/dist/taquito-http-utils.umd.js:504:31)
        at Generator.throw (<anonymous>)
        at rejected (/Users/huianyang/ECAD_LABS/taquito/packages/taquito-http-utils/dist/taquito-http-utils.umd.js:28:69)
        at processTicksAndRejections (node:internal/process/task_queues:95:5) {
      method: 'GET',
      url: 'http://ecad-ghostnet-rolling-2.i.tez.ie:8732/chains/main/chain_id',
      cause: FetchError {
        message: 'request to http://ecad-ghostnet-rolling-2.i.tez.ie:8732/chains/main/chain_id failed, reason: socket hang up',
        type: 'system',
        errno: 'ECONNRESET',
        code: 'ECONNRESET'
      }
    }
```

dsawali:
relevant issues:
- https://github.com/nodejs/node/issues/46287
- https://github.com/whatwg/fetch/issues/1631

PR #3169, opened 2025-06-20
Keepalive false for node v19+

cryi:
Fixes the issue mentioned in #2973.

It was supposedly resolved by #2986, but in Node.js v20 and later, `fetch` requires an explicit HTTP/HTTPS agent with `keepAlive: false`—setting `keepalive: false` alone is not sufficient/effective.

### DIFF
--- a/packages/taquito-http-utils/package.json
+++ b/packages/taquito-http-utils/package.json
@@ -12,6 +12,10 @@
     "signature.json",
     "dist"
   ],
+  "browser": {
+    "http": false,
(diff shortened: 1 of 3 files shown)


hui-an-yang:
Hello @cryi 
Thank you so much for the pr and the fix, we look forward to merge this!
I noticed all the ci test failed at build stage with @taquito/http-utils package
Seems a type issue from the error messages, would you be able to fix and push another commit?
image

PR #3228, opened 2025-11-13
3200 clean up all deprecated export functions and constants

maxwellward:
Thank you for your contribution to Taquito.

Before submitting this PR, please make sure:

- [ ] Your code builds cleanly without any errors or warnings
- [ ] You have run the linter against the changes
- [ ] You have added unit tests (if relevant/appropriate)
- [ ] You have added integration tests (if relevant/appropriate)
- [ ] All public methods or types have TypeDoc coverage with a complete description, and ideally an @example
- [ ] You have added or updated corresponding documentation
- [ ] If relevant, you have written a first draft summary describing the change for inclusion in Release Notes. 

In this PR, please also make sure: 

- [ ] You have linked this PR to the issue by putting `closes #TICKETNUMBER` in the description box (when applicable)
- [ ] You have added a concise description on your changes
## Release Note Draft Snippet

__If relevant, please write a summary of your change that will be suitable for
inclusion in the Release Notes for the next Taquito release.__

PR #3230, opened 2025-11-17
3010 fix all build warnings

### LINKED ISSUE #2997: SPIKE: Lerna/Nx can hide important build problems of individual packages.
Consider the following command: `npm run build` which translates (in `package.json`) to: `nx run-many --target=build --exclude=@taquito/website,taquito-test-dapp-vite` outputs the following:
image

But then if we build an individual package separately: `npm run build -w packages/taquito-michelson-encoder` we can see:
image

Please note that the build problems are masked in the `nx` run.

There are two problems with this:

- If it's not caught when developing and merges into master, the released files could be defective
- It can lead to problems that are difficult to debug (because the developer thinks the build is working) when developing

### LINKED ISSUE #3010: Fix all build warnings
This is related #2997 

There are currently some build warnings when individual packages are built, there are lots of build warnings.
It's a good idea to fix all these warnings.

maxwellward:
Depends on #3228 

Closes #3010 
Closes #2997 

Fixes build warnings (global variable names, circular dependencies, and unresolved dependencies) across all packages except Michelson packages.

Also adds a `build-verbose` command that will provide more logs about the build process, including printing warnings (which are suppressed by Nx by default)


packages/taquito/src/contract/contract-methods/contract-method-flat-param.ts
```diff
@@ -70,7 +75,7 @@ export class ContractMethod<T extends ContractProvider | Wallet>
   send(
     params: Partial<SendParams> = {}
   ): Promise<T extends Wallet ? TransactionWalletOperation : TransactionOperation> {
-    if (this.provider instanceof Wallet) {
+    if (isWallet(this.provider)) {
```
hui-an-yang: Curious of this change what value odes this brings?

packages/taquito-utils/rollup.config.ts
```diff
@@ -9,14 +10,36 @@ const libraryName = 'taquito-utils';
...
+        "@taquito/core": "core",
+        "@stablelib/ed25519": "ed25519",
+        elliptic: "elliptic",
+        "@noble/curves/bls12-381": "bls12381"
+      } 
+    },
     { file: pkg.module, format: 'es', sourcemap: true },
   ],
   // Indicate here external modules you don't wanna include in your bundle (i.e.: 'lodash')
   external: [
```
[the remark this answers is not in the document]
maxwellward: Buffer here is handled by the nodePolyfills plugin. If we include it in the external array, users will have to set up buffer polyfills themselves, whereas this way we do all the polyfills within Taquito, simplifying the install and setup process.

We still get most of the benefits of a smaller bundle size by excluding most things, but I (from a user point of view) would weigh the benefit of not having to mess with polyfills as better than a slightly smaller package size :)

Open to feedback though!

packages/taquito/src/wallet/wallet.ts
```diff
@@ -197,6 +197,10 @@ export class WalletOperationBatch {
 }
 
 export class Wallet {
+  static isWallet(obj: any): obj is Wallet {
```
hui-an-yang: Would also like to understand the change adding static isWallet to Wallet class?

packages/taquito/src/contract/contract-methods/contract-method-flat-param.ts
hui-an-yang: just realized this file can be deleted after the method being removed after deprecation

packages/taquito/src/contract/contract-methods/contract-method-object-param.ts
```diff
@@ -1,80 +1,93 @@
...
+   *
+   * @description Send the smart contract operation
+   *
+   * @param Options generic operation parameter
+   */
+  send(
+    params: Partial<SendParams> = {}
+  ): Promise<T extends Wallet ? TransactionWalletOperation : TransactionOperation> {
+    // provider is either ContractProvider | Wallet but pkh method is unique to Wallet
+    if (this.provider && 'pkh' in this.provider) {
```
hui-an-yang: Hey @maxwellward, i realized this logic is not shared  in other files anymore so decided to save the trouble of isolating it into a reusable function and risk circular dependencies hence wallet-guard file is removed. And I came to senses to appreciate your original logic which is very neat!

PR #3259, opened 2025-12-18
Add tallinnnet rpc node to rpc json

maxwellward:
Depends on Tallinnnet rpc node


hui-an-yang (approved):
lgtm, made small adjustment but ready to merge now

PR #3289, opened 2026-01-06
fix: update hardcoded version links to use default version

maxwellward:
Previously, certain links on the Taquito site would direct to 23.0.0 no matter what. These have been updated to use the default version variable.


hui-an-yang (approved):
lgtm

PR #3291, opened 2026-01-07
Minor website bugfixes

maxwellward:
- Minor XSS bugfix in search results (non-exploitable)
- Add null checks to chatkit initialization
- ARIA labels and accessibility improvements to carousels


hui-an-yang (approved):
lgtm

PR #3326, opened 2026-02-27
Override Beacon SDK default matrix relay nodes

jevonearth:
- Override Beacon SDK's default matrix relay nodes with a curated list of Papers and Trilitech operated nodes (Synapse 1.148.0)
- Empty decommissioned Papers regional nodes (NA, Asia, Australia) to prevent connections to dead servers
- Users can still pass their own `matrixNodes` to override Taquito's defaults
- Fix various doc issues: stale wallet list, incorrect network types, WalletConnect beta label, Ghostnet shutdown notice

Based on Beacon SDK v4.8.0 (`@airgap/beacon-dapp ^4.8.0`).


hui-an-yang (approved):
lgtm

PR #3401, opened 2026-04-13
Stabilize shadownet integration tests

jevonearth:
## Summary
- add an RPC visibility helper that clears cached reads while polling for post-operation state
- harden the flaky shadownet integration specs by waiting for contract, balance, delegate, storage, and ticket state to become visible
- reduce signer counter drift by using fresh signers in the batch and tz2/tz4 suites, and raise a tight funding floor for the key collections test

## Verification
- npx prettier --write on touched files
- git diff --check
- npm --prefix integration-tests run test:shadownet -- __tests__/contract/call-estimate.spec.ts __tests__/contract/set-delegate-auto-estimate.spec.ts
- npm --prefix integration-tests run test:shadownet -- __tests__/contract/batch/batch.spec.ts __tests__/contract/tz4-operations-without-reveal-tests.spec.ts

## Notes
- the second shadownet run passed batch.spec.ts and most of tz4-operations-without-reveal-tests.spec.ts
- the remaining tz4 failures were update_consensus_key_on_unregistered_delegate in the update_consensus_key and update_companion_key cases, which looks separate from the original flap

### DIFF
--- a/integration-tests/__tests__/contract/batch/batch.spec.ts
+++ b/integration-tests/__tests__/contract/batch/batch.spec.ts
@@ -7,7 +7,7 @@ CONFIGS().forEach(({ lib, rpc, setup, knownBaker, knownContract, createAddress }
   const Tezos = lib;
 
   describe(`Test contract.batch through contract api using: ${rpc}`, () => {
-    beforeAll(async () => {
+    beforeEach(async () => {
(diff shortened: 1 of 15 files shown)


jevonearth:
I worked these improvements into PR #3425 so this PR is being closed.

PR #3408, opened 2026-04-14
fix(local-forging): use default bignumber import for ESM

jevonearth:
## Summary
- replace the named `BigNumber` import in `@taquito/local-forging` Michelson codec with a default import
- avoid generating an invalid ESM bundle form like `import BigNumber$1, { BigNumber } from "bignumber.js"`
- fix browser/runtime failures in consumers such as the Sapling test in the Vue test dApp

## Verification
- confirmed the currently published `@taquito/local-forging` ESM bundle in the Vue test dApp contains the invalid named import
- verified the one-line source fix locally


jevonearth:
Closing this narrow fix. We need to handle the broader bignumber.js v10 browser/ESM regression class systematically instead of patching one symptom.

